PR #11, opened 2026-05-18
Clear Slack status when live delivery falls back

decofe:
## Summary
- investigated the stuck Slack SDK spinner on thread `slack:T092R71U6QY:C0AJ07U8Z1N:1779136007.907719`
- found the execution completed and final delivery succeeded, but live Slack delivery failed on `item.started` and disabled `slackbot_session_id` before the finalizer could call `/done`
- close the Slackbot agent session immediately when live harness-event delivery fails so assistant status is cleared before falling back to final delivery
- add a regression test for this failure path

## Verification
- `uv run ruff check api tests/test_agent_control_plane.py`
- `uv run python -m compileall -q api tests/test_agent_control_plane.py`
- `uv run pytest tests/test_agent_control_plane.py -k closes_slackbot_session_when_live_delivery_fails` (skipped locally because the DB fixture is unavailable)

### DIFF
--- a/services/api/api/runtime_control.py
+++ b/services/api/api/runtime_control.py
@@ -2530,6 +2530,11 @@ async def _process_execution_impl(pool, row: dict[str, Any]) -> None:
                             execution_id,
                             "harness_event_failed",
                         )
+                        await slackbot_client.session_done(
+                            slackbot_session_id,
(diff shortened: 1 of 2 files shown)


goksu:
Not sure if this is actually right. Closing.

PR #79, opened 2026-05-19
fix: make docs dark mode only

goksu:
## Summary
- force docs theme assets and variables to dark mode
- set docs main content background to the Vocs primary background token
- keep sidebar active item styling intact

## Test
- npm run build


goksu:
Something borked about this change. Closing. Will re-try later.

PR #83, opened 2026-05-19
Set docs favicon to white mark

goksu:
## Summary
- point the Vocs docs favicon at /brand/mark-white.png

## Testing
- npm run build


goksu:
This already was the case. Closing.

PR #85, opened 2026-05-20
fix: handle slack stop as cancellation

o-az:
- support `stop` / `cancel` as a control command
 - clear persisted resume state on stop so a later mention in the
   same thread starts cleanly instead of resuming a dead Codex
   rollout


Zygimantass:

PR #86, opened 2026-05-20
fix: frame slack history handoff

o-az:
the agent has sufficient context on what the current prompt is and what the other messages in the thread are, however it sometimes treats the thread starter as the primary piece of context. This makes Slack history handoff frame prior thread messages as background context


Zygimantass:

PR #88, opened 2026-05-20
feat: handle `slackbot_live_delivery_failed` & `msg_too_long`

o-az:
- split oversized slack final-delivery text into follow-up replies
 - preserve already streamed answer offset when live slack delivery fails, so fallback delivery only posts the missing continuation


arjunblj:
Superseded by #90, which includes this delivery overflow work plus retry/streak/idempotency/permanent-error handling. I credited you in the PR body and commit footer, and left this open so it closes automatically when #90 merges.

PR #89, opened 2026-05-20
feat: add oauth_token password grant and token_endpoint_headers

mslipper:
Adds the RFC 6749 4.3 `password` grant to the `oauth_token` secret type (required fields: `username`, `password`, `client_id`; optional: `client_secret`). Adds a new `token_endpoint_headers` map on any `oauth_token` entry that sends extra headers on the token POST itself, each value resolved from its own secret source: useful when the token endpoint requires an API key alongside the standard form-body client auth. Bumps iron-proxy from 0.37.0 to 0.38.0.


arjunblj:
lgtm

PR #99, opened 2026-05-20
fix(slackbot): drop the 'Thinking...' assistant.threads.setStatus pill

arjunblj:
## Summary
- Stop calling `assistant.threads.setStatus(Thinking...)` on agent-session open; the streaming UI already signals work.
- Initialize `statusCleared: true` so the clear-status paths short-circuit.

## Test plan
- `bun --cwd services/slackbot test src/slack/agent-session.test.ts`


Zygimantass:

PR #102, opened 2026-05-20
fix(slackbot): drop standalone *Thinking* block from final composed message

arjunblj:
## Summary
- Remove the duplicate `*Thinking*` context block from `chat.stopStream` blocks; commentary already streams live via the codex commentary path or as plan-task `details`.
- Keep commentary in the chat fallback text so mobile push / notification previews still carry the reasoning.
- Delete the now-unused `thinkingContextBlock` / `shouldShowThinkingBlock` helpers + their orphan test.

## Test plan
- `bun --cwd services/slackbot test src/slack/agent-session.test.ts src/slack/codex-session.test.ts`


Zygimantass:

PR #103, opened 2026-05-20
fix(api): eagerly close slackbot sessions when release cancels in-flight work

arjunblj:
## Summary
- `release_assignment(..., cancel_inflight=true)` now `RETURNING`s cancelled rows' metadata, collects distinct `slackbot_agent_session_id` values, and calls `slackbot_client.session_done` post-commit so the "Thinking" pill disappears within ~5s instead of hanging until the next message.
- Errors are logged but never block the release response.

## Test plan
- `uv run --project services/api pytest tests/test_runtime_control_release_session_done.py -q`

### DIFF
--- a/services/api/api/runtime_control.py
+++ b/services/api/api/runtime_control.py
@@ -1632,6 +1632,34 @@ async def steer_execution(
     }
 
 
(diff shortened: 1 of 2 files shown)


Zygimantass:

PR #110, opened 2026-05-20
Classify Slack final delivery API errors

gakonst:
## Summary
- classify Slack final-delivery failures using thrown Web API `data.error` and `code`, not only `Error.message`
- add a regression test for thrown `msg_blocks_too_long` errors so they are marked non-retryable immediately

## Evidence
- vlogs showed `msg_blocks_too_long` final delivery failures retried 50 times with `error_class=unknown` before dead-lettering
- current Slack turn trace reached `execute_started` and `turn_first_output` cleanly; the API execution path was not the failing surface

## Tests
- `bun test src/centaur/final-delivery.test.ts`
- `bun test src/index.test.ts src/slack/dedup.test.ts`

### DIFF
--- a/services/api/api/workflow_engine.py
+++ b/services/api/api/workflow_engine.py
@@ -1783,6 +1783,17 @@ async def _insert_workflow_run(
         )
         if existing:
             if existing["request_hash"] != req_hash:
+                keys = ",".join(sorted(str(key) for key in run_input.keys()))
+                log.warning(
+                    "workflow_idempotency_payload_mismatch",
+                    workflow_name=workflow_name,
+                    trigger_key=trigger_key,
+                    thread_key=run_input.get("thread_key"),
+                    input_keys=keys,
+                    existing_request_hash_prefix=str(existing["request_hash"])[:12],
+                    request_hash_prefix=req_hash[:12],
+                    run_id=str(existing["run_id"]),
+                )
                 raise ControlPlaneError(
                     "IDEMPOTENCY_PAYLOAD_MISMATCH",
                     "trigger_key was already used with a different payload",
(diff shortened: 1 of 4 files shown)


gakonst:
Additional debug pass evidence:

- Runtime trace for the original Slack turn showed workflow enqueue/claim, one fast /workflows/runs 409, then successful 200 and execution start.
- 24h logs show repeated generic /workflows/runs 409s without a structured mismatch event, so this PR now adds workflow_idempotency_payload_mismatch with safe fields and hash prefixes.
- Live dev workflow smoke: run wfr_fde3fb3c29b74725 / execution exe_39538bbe27fb44b5 completed with result_text=PONG and delivery.platform=dev.
- Local checks: bun test src/centaur/final-delivery.test.ts; uv run ruff check api/workflow_engine.py tests/test_workflow_idempotency_unit.py; uv run pytest tests/test_workflow_idempotency_unit.py.
- CI is green for API workflow/control-plane tests, Slackbot interface/delivery tests, and publish jobs.

gakonst:
Follow-up root-cause pass:

- The recent /workflows/runs 409s correlate with workflow_history_backfill_skipped events, not duplicate active-turn execution. Example trace 927a9847-602b-4a6f-9c32-8b63e75b5bd0 enqueued/ran the active turn, then backfill hit an existing Slack message_id with different workflow-specific metadata and skipped it.
- This PR now skips history backfill messages that already exist in agent_message_requests before calling append_message, so historical Slack replay no longer creates false idempotency mismatch noise. Current-turn idempotency remains unchanged.
- Added unit coverage for both the structured workflow idempotency mismatch log and the existing-history-message skip.
- Local checks: uv run pytest tests/test_workflow_idempotency_unit.py; uv run ruff check api/workflow_engine.py tests/test_workflow_idempotency_unit.py; bun test src/centaur/final-delivery.test.ts.
- CI is green again.

Issue #109, opened 2026-05-20
Integrate Hermes Agent as a Centaur harness


decofe:
Integrate Hermes Agent as another Centaur harness/runtime option.

Where it fits: Centaur's current architecture keeps Slackbot and other clients thin: they call `spawn -> message -> execute`, the API assigns or reuses a sandbox runtime, and the sandbox runs the selected harness command (`codex`, `claude-code`, `amp`, etc.) while normalized events stream back through `/agent/threads/{thread}/events`. Hermes should fit at that harness layer, not as a separate Slackbot path.

Suggested scope:
- Add a Hermes harness/engine entry alongside the existing sandbox harnesses.
- Add the sandbox command/config needed to launch Hermes Agent inside the agent image.
- Normalize Hermes output into Centaur's canonical execution events so Slackbot, API clients, and final delivery keep working through the existing event stream.
- Document how to select Hermes as a harness.

Acceptance criteria:
- A caller can request Hermes through the existing agent API harness selection flow.
- Hermes runs inside the existing Kubernetes sandbox runtime.
- Hermes output appears through the normal Centaur event stream and final delivery path.
- Basic tests or fixtures cover command selection and event normalization.

Osraka:
I'd like to work on this. I plan to start with the narrow harness-server path: add a Hermes harness kind, a deterministic protocol adapter/fixture, and tests that normalize Hermes terminal and assistant updates into Centaur's canonical execution events before wiring the sandbox image and API configuration. Existing Codex/Claude/Amp paths would remain unchanged. Before I start, could you confirm the expected Hermes invocation/auth mode and whether the integration should use Hermes' ACP stdio server with one persistent session per sandbox, or a different in-image command?

PR #142, opened 2026-05-22
feat(harness): integrate Hermes Agent as a Centaur harness

0xprames:
Add NousResearch's Hermes Agent as a selectable harness/engine, fitting at the existing sandbox-runtime layer so Slackbot and API clients stay thin and keep streaming through the normal /agent event pipeline (issue #109).

Hermes is driven through its Agent Client Protocol (ACP) stdio server (`hermes-acp`), not the oneshot CLI: every existing wrapper (codex/claude/amp) is a persistent, streaming, interruptible bridge, and ACP matches that contract 1:1 — session/new + session/prompt per turn, streaming session/update notifications, session/cancel for interrupt. Its event taxonomy (agent_message_chunk, agent_thought_chunk, tool_call/tool_call_update, plan) maps directly onto Centaur's canonical events. Oneshot would have yielded only a final blob with no streaming/tools/interrupt.

- services/sandbox/hermes-app-wrapper.py: ACP client that spawns hermes-acp, drives the session, auto-approves tool permission requests, and re-emits Hermes session updates as Centaur NDJSON. Imports `acp` lazily so the module stays importable + unit-testable outside the sandbox image.
- Selection: "hermes" added to _ENGINE_HARNESSES (agent.py, normalize.py) and _EXECUTION_HARNESSES (slack_thread_turn.py); build_harness_cmd → ["hermes-app-wrapper"]; --hermes / harness=hermes flags; HERMES_MODEL + HERMES_CONTINUE_SESSION_ID env and HERMES_PROVIDER/HERMES_MODEL passthrough.
- Normalization: _normalize_hermes_event (normalize.py) with TS parity in packages/harness-events/src/normalize.ts; is_turn_done / extract_result / extract_thread_id branches in harness_protocol.py.
- Image/runtime: Dockerfile installs hermes-agent[acp]==0.14.0 and COPYs the wrapper; entrypoint.sh writes ~/.hermes/config.yaml from HERMES_PROVIDER / HERMES_MODEL, defaulting to the Anthropic provider + a Claude model so it reuses the stubbed ANTHROPIC_API_KEY + iron-proxy path out of the box.
- Tests: hermes cases in test_normalize.py + test_harness_protocol.py and a new test_hermes_app_wrapper.py covering command selection + event translation. Docs: harness-credentials table + flag list in both deploy guides.


0xprames:
stubbed out a PR with claude for #109 , can take a further look shortly

needs e2e testing / manual review

gakonst:
Should we make everything work with ACP?

gakonst:
I guess the thing I like rn is that the agent/message endpoint queues up a message without submitting it

0xprames:
the choice for ACP with `hermes` here was based on Claude stating it would maintain the contract maintained by existing harnesses (Codex etc) in the codebase

i wanted to verify this with a deeper dive, but did not get around to it yet.

the other option seems to be `oneshot` cli mode, which apparently does not give parity with the other harnesses Centaur uses

gakonst:
Yeah Def don't want one shot I wanna be able to pipe over stdio over a running process

0xprames:
needs e2e testing, actually deploying the change and playing with the functionality

from a first pass at the code, it seems OKish - removed ACP and instead:

```
constructs AIAgent like hermes_cli.oneshot but wires stream_delta_callback→agent_message_chunk, reasoning_callback→agent_thought_chunk, tool_progress_callback→tool_call, step_callback→tool_call_update (+todo→plan); chat() per turn on
   a worker thread so interrupt() fires from another thread; dups fd 1 to a private NDJSON channel and routes fd 1→stderr so Hermes' stray prints can't corrupt the stream (done in main(), not import, so the module stays test-loadable).
```

ill show this PR some manual love (both testing and review) shortly.

Epistetechnician:
I pushed the Hermes E2E hardening/fixup as a stacked PR against this PR's source branch:

https://github.com/0xprames/centaur/pull/1

Key E2E proof from local kind:
- Hermes spawn -> message -> execute completed with result_text=PONG
- Hermes tool/todo run emitted tool + plan events and completed with result_text=DONE
- Hermes steer/interrupt E2E completed through the normal event stream with result_text=STEERED
  - thread: test-hermes-steer-e2e-20260524134107
  - execution: exe_7d1a508ec5ce4a92

Validation run:
- focused Hermes/control-plane pytest suite: 125 passed
- ruff check: passed
- ruff format --check: passed
- git diff --check: passed
- just build-one api: passed
- just build-one sandbox: passed

0xprames:
ty @Epistetechnician - as discussed offline, lets get some deep manual testing done as well. Can help after i wrap things up

Epistetechnician:
## Hermes harness follow-up: acceptance status + Slack E2E evidence

I pushed a clean follow-up branch against this PR branch here:

https://github.com/0xprames/centaur/pull/2

That PR is based directly on `0xprames:add-hermes-harness` and contains only the follow-up hardening commits:

- `85bc4364 fix(harness): harden Hermes E2E runtime`
- `7d24de91 fix(harness): auto-configure Hermes runtime`
- `9a4daf9f test: keep Hermes proxy fixture provider agnostic`

### Acceptance criteria status

| Criteria | Status | Notes |
|---|---:|---|
| Caller can request Hermes through existing agent API harness selection flow | PASS | `harness="hermes"` is wired through the existing API/router/workflow harness selection path. Slack remains a thin caller using the same flow. |
| Hermes runs inside existing Kubernetes sandbox runtime | PASS | Hermes runs through the normal Centaur sandbox image/runtime, not a separate Slackbot path. |
| Hermes output appears through normal Centaur event stream and final delivery path | PASS | Live API and Slack runs completed through normal execution rows/events and Slack final delivery. |
| Basic tests/fixtures cover command selection and event normalization | PASS | Focused tests cover wrapper behavior, sandbox command selection, Kubernetes env/config, proxy config, and event normalization. |

### Focused validation on the clean prames-based branch

```bash
uv run --project services/api pytest \
  services/api/tests/test_proxy_config.py \
  services/api/tests/test_hermes_app_wrapper.py \
  services/api/tests/test_sandbox_entrypoint.py \
  services/api/tests/test_sandbox_kubernetes_backend.py -q
```

Result: `148 passed`.

Provider/model agnostic check: searched for Minimax-specific strings after cleanup; no `minimax`, `MiniMax`, `M2.7`, or `api.minimax` references remain in the clean branch.

### Manual Slack/Centaur E2E evidence from 2026-05-24

Slack channel used: `C0B5E9NV1K9`.

Representative completed Slack/API runs:

- `@centaur --hermes Reply with exactly HERMES_SLACK_OK...` returned through Slack final delivery.
- Direct API `/agent/spawn -> /agent/message -> /agent/execute` Hermes continuity check completed two turns: first `SAVED`, follow-up `HERMES_NONCE_7429`.
- Slack workspace/tool execution returned `WORKSPACE_TOOL_OK` and durable events showed Hermes `search_files` tool calls.
- Slack plan/tool normalization returned `PLAN_TOOL_OK`; durable events showed Hermes todo/tool activity and Centaur `turn.plan.updated` events.
- Parallel Slack safe rerun completed both independent Hermes sessions: `SAFE_PARALLEL_A_OK` and `SAFE_PARALLEL_B_OK`.
- Sequential Slack tool audit on thread `1779660602.669229` completed multiple Hermes turns and confirmed event-backed tool payloads for `search_files`, `read_file`, `terminal`, `execute_code`, `skills_list`, `todo`, `browser_navigate`, `vision_analyze`, `text_to_speech`, and `session_search`.

Observed Hermes runtime/tool caveats from the Slack audit, not Centaur integration blockers:

- Browser tool call reached Hermes/Centaur events, but Chromium is missing from the sandbox image.
- Vision/TTS calls reached Hermes/Centaur events, but providers were not configured in the test runtime.
- `web_search`/`web_extract` and `image_generate` were listed by Hermes CLI toolsets but not exposed in the model-facing callable list in this runtime.

Slash-command audit: `/help` and `/status` through `@centaur --hermes` completed as normal model turns, not Hermes gateway command dispatch. That is expected for the Centaur harness path; Hermes slash commands belong to Hermes CLI/gateway surfaces unless we intentionally add a Centaur bridge later.

### Scope check

This keeps Hermes at the harness/runtime layer. No separate Slackbot route was added, and Slack/API clients continue to use the existing Centaur control-plane flow.

0xprames:
@gakonst @arjunblj  should be ready for another pass

par4m:
@0xprames   This looks like the right integration pattern, and it’s very relevant to the Claude Agents SDK work I’m thinking through.

    The main thing I want to preserve is the Centaur-native boundary:

    - Slack/API/Web callers stay thin.
    - Runtime selection flows through the existing harness/persona/thread path.
    - Turns still go through agent_execution_requests.
    - The worker still owns claim/lease/retry/cancel.
    - Output still streams through the normal /agent event pipeline.
    - Results/observations still persist through Centaur.
    - There is no separate Slackbot/Web route or parallel agent runtime.

    For Claude Agents SDK, I think the goal should be the same shape as this Hermes harness: plug in at the runtime/session/event layer, not at the client layer.

    Current sandbox harnesses:

    Centaur runtime worker
    → sandbox backend
    → wrapper process
    → streaming NDJSON
    → Centaur event normalization/result pipeline


    Claude Agents SDK could follow the same contract, but potentially as a direct/virtual backend instead of another sandbox wrapper:

   
    Centaur runtime worker
    → ClaudeDirectBackend / virtual session
    → persistent ClaudeSDKClient
    → SDK events mapped to Claude/Centaur-compatible NDJSON
    → Centaur event normalization/result pipeline


The design goal is not to put Claude Agents SDK in Slackbot or the Web UI. Web/Slack should only call Centaur. Centaur should remain the source of truth for durability, audit, permissions, retries, cancellation, workflow resume, and delivery.

    The reason Claude Agents SDK is interesting is that it can provide a persistent, streaming, interruptible agent session with warm MCP/tool connections and less startup overhead than spawning Claude Code each time. But it should still fit behind the same runtime contract used here.

    A few questions this raises:

    1. Should Hermes be treated as the precedent for future agent runtimes: persistent streaming bridge at the Centaur runtime layer, never a client-specific route?

    2. For Claude Agents SDK, should we:
       - first add a Hermes-style sandbox wrapper for fastest parity, then
       - later move to a direct/virtual backend to remove K8s/pod startup,

       or go directly to the virtual backend if it can preserve the same write_stdin / stream_stdout / interrupt semantics?

    3. If we add direct SDK backends, should we eventually rename/generalize SandboxBackend to something like RuntimeBackend or ExecutionBackend, with Kubernetes as one backend implementation?

    4. For direct SDK mode, how strict should the tool boundary be? My instinct is that SDK tools should route through Centaur’s tool gateway / ToolManager, so secrets, policy, audit, and idempotency stay in Centaur. I’d avoid exposing raw filesystem/bash/customer credentials directly to the SDK unless there’s a separate isolation boundary.

 Overall, I think this PR points in the right direction: new agent runtimes should plug into Centaur’s runtime/session/event pipeline, not bypass it. Claude Agents SDK would follow the same shape; the main open question is sandbox wrapper first vs direct backend first.


tldr: use the Hermes pattern for runtime integration, but for Claude Agents SDK explore a direct/warm backend so we don’t have to spin up a full Claude Code/sandbox process for every lightweight agent turn while still preserving Centaur’s durable execution pipeline. Since Claude Agents SDK can do everything claude code can!

Zygimantass:
do you think there's a chance that we can translate hermes-agent traces into codex-app-server format? if so, the PR should be ported over to new api-rs / slackbot, closing til' then. ping me to take a look if you're still interested:)

PR #143, opened 2026-05-22
fix: show Slack metadata once per thread

### LINKED ISSUE #107: Show Slack harness metadata only once per thread
Slack replies currently include the harness/model metadata at the top of every response, which makes back-to-back thread reading noisier than necessary.

Suggested behavior: show the harness/model metadata on the first Centaur reply in a Slack thread, then omit it from later Centaur replies in the same thread.

Acceptance criteria:
- The first Centaur reply in a Slack thread still shows harness/model metadata.
- Follow-up Centaur replies in that same thread omit the metadata line.
- Existing non-Slack delivery surfaces are unchanged unless they intentionally share this formatter.

arihantbansal:
## Summary
- show Slack persona/model metadata only on the first live Slackbot reply in a thread
- suppress that metadata for follow-up turns when prior Slackbot session metadata or assistant history exists
- update stale workflow attachment assertions for preserved attachment names

## Tests
- `uv run --project services/api pytest services/api/tests/test_workflows.py -q`
- `uv run --project services/api ruff check services/api/api/workflow_engine.py services/api/tests/test_workflows.py`
- `git diff --check`

Closes #107

### DIFF
--- a/services/api/api/workflow_engine.py
+++ b/services/api/api/workflow_engine.py
@@ -973,6 +973,46 @@ async def _compute_agent_session_header(
     )
 
 
+def _history_has_assistant_message(history_messages: Any) -> bool:
+    if not isinstance(history_messages, list):
+        return False
+    return any(
+        isinstance(item, dict)
+        and str(item.get("role") or "").strip().lower() == "assistant"
+        for item in history_messages
+    )
+
+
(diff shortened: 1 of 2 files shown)


arjunblj:
Thank you for this. FYI - rebasing on main and testing locally now.

PR #155, opened 2026-05-22
revert: refresh session context on every turn

akshaan:
## Summary
- revert #154 / 23c74446, which prepended generated session context to every Slack turn
- return session-context delivery behavior to the #153 state while keeping core PR attribution text

## Testing
- uv run pytest tests/test_integration.py -k "BuildSessionContext or insert_system_message" (from services/api)
- uv run ruff check api/agent.py tests/test_inflight_replay.py (from services/api)

Not merging without explicit approval.


Zygimantass:

PR #162, opened 2026-05-23
Support pausable agent sandboxes

Zygimantass:
## What changed

This adds an opt-in Agent Sandbox backend path for Centaur sandboxes so we can pause pods without throwing away local agent state.

The important bit: idle sandboxes can now scale to zero through the Agent Sandbox controller, keep their PVC around, and come back on the next turn with the same workspace/codex/claude state.

## How people use this

- If your cluster already installs Agent Sandbox, set `sandbox.controller=agent-sandbox` and `sandbox.stateVolume.enabled=true`.
- If this Centaur release should install it, set `agentSandbox.enabled=true`. This enables the vendored upstream Agent Sandbox Helm chart as an optional dependency.
- The normal pod backend remains the default. It still lives in `kubernetes.py`.
- The Agent Sandbox CRD/PVC lifecycle lives in `kubernetes_agent_sandbox.py`, so we can delete the old pod implementation later without hunting through controller branches.
- The vendored chart is pinned from upstream Agent Sandbox `v0.4.6`; the controller image tag defaults to `v0.4.6`.
- Set `agentSandbox.controller.extensions=true` only if you want the optional upstream extension controllers enabled.
- Tune `sandbox.stateVolume.size` / `sandbox.stateVolume.storageClassName` for the cluster.
- Set `IDLE_TTL_S` and `SUSPENDED_RETENTION_S` through `api.extraEnv` when you want shorter/longer pause and cleanup windows.

Example local testing values:

```bash
helm upgrade centaur contrib/chart -n centaur --reuse-values \
  --set agentSandbox.enabled=true \
  --set sandbox.controller=agent-sandbox \
  --set sandbox.stateVolume.enabled=true \
  --set sandbox.stateVolume.size=1Gi \
  --set api.extraEnv.IDLE_TTL_S=30 \
  --set api.extraEnv.SUSPENDED_RETENTION_S=120
```

## Notes

- The upstream chart currently lives in the Agent Sandbox repo, so this vendors it as a local `file://charts/agent-sandbox` dependency instead of using the old manifest-apply hook.
- PVC cleanup only happens on final suspended-retention cleanup / explicit stop, not on idle pause.
- Resume handles the pod deletion race from `replicas=0` without falling through to a cold spawn that deletes state.
- The sandbox entrypoint now persists workspace, uploads, branches, `.codex`, and `.claude` under `/home/agent/state` when a state volume is mounted.

## Tested

- `helm dependency build contrib/chart`
- `helm template centaur contrib/chart -n centaur -f contrib/chart/values.dev.yaml`
- `helm template centaur contrib/chart -n centaur -f contrib/chart/values.dev.yaml --include-crds --set agentSandbox.enabled=true --set sandbox.controller=agent-sandbox --set sandbox.stateVolume.enabled=true`
- `uv run --project services/api ruff check services/api/api/sandbox/kubernetes.py services/api/api/sandbox/kubernetes_agent_sandbox.py services/api/api/sandbox/registry.py services/api/tests/test_sandbox_kubernetes_backend.py`
- `uv run --project services/api pytest services/api/tests/test_sandbox_kubernetes_backend.py -q`
- `just build-one api`
- `just deploy`
- Built and rolled the API locally on OrbStack.
- OrbStack smoke: `POST /agent/spawn` through the Agent Sandbox backend created a Sandbox CR, Running pod, and Bound state PVC; `/agent/stop` cleaned them back up.
- Earlier OrbStack smoke: create sandbox, write marker under `/home/agent/state`, wait for suspend, resume same thread, verify same runtime id and marker persisted.

### DIFF
--- a/contrib/chart/.gitignore
+++ b/contrib/chart/.gitignore
@@ -1,1 +1,3 @@
-charts/
+charts/*
+!charts/agent-sandbox/
+!charts/agent-sandbox/**
--- a/contrib/chart/Chart.lock
+++ b/contrib/chart/Chart.lock
@@ -2,5 +2,8 @@ dependencies:
 - name: connect
   repository: https://1password.github.io/connect-helm-charts
   version: 2.4.1
-digest: sha256:d77b0cc806ebf6c61fe9dd45f7e29717493589ac611d198d79e4e63f40aa39ac
-generated: "2026-05-12T12:48:48.525819-06:00"
+- name: agent-sandbox
+  repository: file://charts/agent-sandbox
+  version: 0.1.0
+digest: sha256:0e9f0aac94b5d72367ccf7e8259b50685595b11aba7d656b6c09b8f1fae4e15c
+generated: "2026-05-23T17:05:36.099275+02:00"
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.38
+version: 0.1.39
 appVersion: "0.1.0"
 dependencies:
   - name: connect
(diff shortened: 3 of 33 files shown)


Zygimantass:
#171 and #172 should be fixed after we migrate to agent-sandboxes

PR #176, opened 2026-05-24
fix: gate plugin file watchers behind PLUGIN_WATCHER_ENABLED

arjunblj:
The tool/workflow `awatch` loops hold inotify FDs per pod; rolling deploys exhaust the per-process limit and pods die with `WatchfilesRustInternalError: Too many open files`, killing in-flight executions.

Gate behind `PLUGIN_WATCHER_ENABLED` — default on locally, default off in the chart, override on for `just up` via `values.dev.yaml`. Naming matches the existing `EXECUTION_WORKER_ENABLED` / `WORKFLOW_WORKER_ENABLED` / `WARM_POOL_ENABLED` family.

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.38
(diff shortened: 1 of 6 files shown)


Zygimantass:
lgtm

PR #178, opened 2026-05-24
fix: recycle sandbox sessions when the running image is stale

arjunblj:
After an image bump, sandbox pods pinned to a thread keep running the old image and the warm pool keeps handing out warm sandboxes built from a prior image.

New `SandboxBackend.runtime_image_tags` reads the pod spec — no DB column, no migration. `_sandbox_image_drift` returns a `drift_kind` that flows into logs and the new `agent_sandbox_recycled_total` metric. Wired into `get_or_spawn`, `warm_pool.claim_container`, and `warm_pool.replenish`.

When recycle fires mid-retry on an execution with `durable_turn_id` set, the worker now calls `replay_inflight_turn` so the persisted turn is re-injected into the fresh sandbox.

Textual conflict with #162 in `get_or_spawn`; designs are orthogonal — happy to land either order.


Zygimantass:

PR #186, opened 2026-05-24
fix(security): enforce table allowlist on /agent/query

aaronjmars:
## Summary

`POST /agent/query` exposes a read-only SQL surface that the docstring
and `services/sandbox/SYSTEM_PROMPT.md` both describe as **scoped to an
allowlist** of eleven tables (`chat_messages`, `sandbox_sessions`,
`attachments`, `api_keys`, `agent_runtime_assignments`,
`agent_message_requests`, `agent_execution_requests`,
`agent_execution_events`, `agent_final_delivery_outbox`,
`agent_spawn_requests`, `agent_release_requests`). The
`_ALLOWED_TABLES` constant captures that list and the SYSTEM_PROMPT
literally advertises it to in-sandbox agents as **"Available tables:
…"**.

The implementation, however, only ran three checks:

1. Reject sandbox tokens (`sbx1.*`).
2. Require `sql.lower().startswith("select")`.
3. Reject SQL containing the substrings in `_BLOCKED_PATTERNS`
   (`drop `, `delete `, `insert `, …).

`_ALLOWED_TABLES` was never consulted. Any DB-backed key
(`aiv2_*`) — regardless of its scopes — could read any table in the
public schema (and any system catalog the role had grants on),
including `slack_sync_messages`, `slack_sync_users`,
`company_context_documents`, `muesli_meetings`, `workflow_runs`,
`workflow_events`, `usage_stats`, etc.

## Impact

A holder of a narrow-scope DB key (e.g. one bound to a single tool or
workflow) could call `/agent/query` with
`SELECT * FROM slack_sync_messages` (or any other non-advertised
table) and exfiltrate every synced Slack message, every company
context document, every Muesli meeting transcript, every workflow
input/output, and every usage stat the centaur Postgres role has read
access to. Equivalent reads of `pg_catalog.pg_authid` and
`information_schema.tables` were also permitted, leaking schema and
role metadata.

The scope `agent:execute` (and others) does not transitively imply
"may read the entire database", but with the allowlist not enforced,
any authenticated DB key effectively had full read access. The
sandbox-claim block (line 1555-1556) is intentional and is preserved
unchanged.

## Location

`services/api/api/routers/agent.py` — `query_db` /
`_ALLOWED_TABLES` (defined at line 1538, never read).

## Fix

`_validate_query_safety(sql)` runs after the existing
`startswith("select")` and blocked-keyword checks. It:

- Strips SQL string literals and `--` / `/* */` / dollar-quoted
  comments via `_strip_sql_noise` so an identifier that only appears
  inside a literal does not produce a false positive (e.g.
  `WHERE parts::text LIKE '%pg_read_file(%'` keeps working).
- Extracts every `FROM` / `JOIN` / `INTO` / `USING` / `UPDATE` table
  reference (`_TABLE_REF_RE` + `_normalize_table_ref`), supporting
  unquoted, quoted (`"foo"`), schema-qualified (`public.foo`,
  `"public"."foo"`), and three-part (`db.schema.foo`) forms.
- Rejects any schema other than `public` — closes
  `pg_catalog.pg_authid` and `information_schema.tables` reads.
- Rejects any table outside `_ALLOWED_TABLES` — closes the
  `slack_sync_messages` / `company_context_documents` /
  `muesli_meetings` / `workflow_runs` reads.
- Rejects `pg_read_file` / `pg_read_server_files` /
  `pg_read_binary_file` / `pg_ls_dir` / `pg_stat_file` / `lo_import`
  / `lo_export` / `dblink` / `dblink_exec` / `pg_terminate_backend`
  / `pg_cancel_backend` function calls (caught even after the
  string-literal strip).

The token-level scan is best-effort; the existing
`startswith("select")` + blocked-keyword gates + the sandbox-claim
block stay in place as defense-in-depth so a regression in any one
layer does not re-open arbitrary SQL.

Both SYSTEM_PROMPT.md example queries — the `attachments LIMIT 10`
snippet and the `agent_execution_requests LIMIT 20` snippet — are
included in the new test suite and remain accepted by the validator.

## Detected by

Aeon + manual review (threat-model-claims-invite-scrutiny axis: dead
`_ALLOWED_TABLES` constant + matching docstring + matching
SYSTEM_PROMPT advertisement, none of which were enforced)

- Severity: **medium**
- CWE-1220 — Insufficient Granularity of Access Control

## Verification

`services/api/tests/test_query_validator.py` — a pure-function test
module (no Postgres bootstrap required) that exercises:

- `_normalize_table_ref` on 8 identifier shapes (unquoted, quoted,
  schema-qualified, quoted-with-escaped-double-quote).
- `_strip_sql_noise` on 3 noise shapes (single-quoted literal,
  `--` line comment, `/* */` block comment).
- `_validate_query_safety` on **10 accepted** queries (the two
  SYSTEM_PROMPT examples, plus JOINs across allowlisted tables,
  IN-subqueries, schema-qualified-to-public, quoted identifiers, and
  string-literal / block-comment bypass attempts that the strip
  helper must defang).
- `_validate_query_safety` on **13 rejected** queries
  (`pg_authid`, `pg_catalog.pg_class`, `information_schema.tables`,
  `slack_sync_messages`, `company_context_documents`,
  `muesli_meetings`, `workflow_runs`, `usage_stats`, JOIN-to-non-
  allowlisted, quoted-schema bypass, three-part identifier,
  `pg_read_file`, `dblink`, `pg_terminate_backend`, and mixed-case
  whitespace around a blocked function call).

All 34 cases pass — verified against the patched
`services/api/api/routers/agent.py` outside the integration harness
by extracting the validator block and running it through Python
3.11+ directly. The existing integration suite (which needs
Postgres) is unaffected and will exercise the same validator inside
the `/agent/query` HTTP path under CI.

The change is additive — no existing public API behavior changes for
queries that were already within the documented allowlist.

---
Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).

### DIFF
--- a/services/api/api/routers/agent.py
+++ b/services/api/api/routers/agent.py
@@ -1542,7 +1542,120 @@ _ALLOWED_TABLES = {
     "agent_final_delivery_outbox", "agent_spawn_requests",
     "agent_release_requests",
 }
+_ALLOWED_SCHEMAS = {"public"}
 _BLOCKED_PATTERNS = {"drop ", "delete ", "insert ", "update ", "alter ", "create ", "truncate ", "grant ", "revoke "}
+# Server-side functions that can read arbitrary files, cross-DB-link, or
+# perturb other backends. None of these are needed for the introspection
+# queries the SYSTEM_PROMPT advertises; refuse them outright.
+_BLOCKED_FUNCTIONS = (
+    "pg_read_file",
+    "pg_read_server_files",
+    "pg_read_binary_file",
+    "pg_ls_dir",
+    "pg_stat_file",
+    "lo_import",
+    "lo_export",
+    "dblink",
+    "dblink_exec",
+    "pg_terminate_backend",
+    "pg_cancel_backend",
+)
+# Match SQL noise so the table-reference scan does not trip on identifiers
+# that appear inside string literals or comments.
+_SQL_NOISE_RE = re.compile(
+    r"""'(?:[^']|'')*'|--[^\n]*|/\*[\s\S]*?\*/|\$(\w*)\$[\s\S]*?\$\1\$""",
+    re.MULTILINE,
+)
+_TABLE_REF_RE = re.compile(
+    # Match table identifier after FROM / JOIN / INTO / USING / UPDATE.
+    # Accepts unquoted (foo, public.foo) and quoted ("foo", "public"."foo")
+    # forms, optionally schema-qualified.
+    r"""\b(?:from|join|into|using|update)\s+
+        (?P<ident>
+            (?:"(?:[^"]|"")+"|[A-Za-z_][\w$]*)
+            (?:\s*\.\s*(?:"(?:[^"]|"")+"|[A-Za-z_][\w$]*))?
+        )""",
+    re.IGNORECASE | re.VERBOSE,
+)
+
+
+def _strip_sql_noise(sql: str) -> str:
+    """Replace string literals and comments with spaces.
+
+    Preserves overall character count so callsite line/column reporting
+    (if any) stays aligned. Conservative — leaves anything we cannot
+    classify untouched.
+    """
+
+    def _replace(match: re.Match[str]) -> str:
+        return " " * (match.end() - match.start())
+
+    return _SQL_NOISE_RE.sub(_replace, sql)
+
+
+def _normalize_table_ref(ident: str) -> tuple[str, str | None]:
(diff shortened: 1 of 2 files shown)


aaronjmars:
Friendly bump — enforces a table allowlist on `/agent/query` so the agent can't read outside the intended tables. Mergeable against `main`. Happy to address any feedback whenever a maintainer has a moment.

Zygimantass:

PR #187, opened 2026-05-24
feat(chatbot): fetch thread history when @mentioned

jamalavedra:
## Summary

When the Centaur Google Chat app was @mentioned in a thread, it could only see the message that mentioned it — every prior message (including its own previous responses) was invisible. The agent kept replying with "I don't see what you're referring to in this thread."

Two stacked blockers:
1. `collectThreadHistory` was a stub returning `[]` (`services/chatbot/src/chat/normalize.ts:168` on `main`).
2. The OAuth scope was only `chat.bot`, which Google rejects on `spaces.messages.list` regardless.

This PR fixes both, with hardening from an adversarial code review.

## Changes

**Auth scope**
- `client.ts:41` — adds `chat.app.messages.readonly` alongside `chat.bot`. Requires the existing Workspace admin install of our private Marketplace listing (already granted in our org; verified the scope works via live API call before this PR).

**Thread history fetch**
- `client.ts` — `listMessages` now accepts `filter` and `orderBy` params.
- `normalize.ts` — `collectThreadHistory` is now a real function: paginated `spaces.messages.list?filter=thread.name="..."&orderBy=createTime desc`, capped at 50 most-recent messages, mapped to chronologically-ordered `ChatHistoryMessage[]` with role inferred from `sender.type === 'BOT'`.
- `index.ts` — runs history fetch in parallel with the ack post (`Promise.all`), so a slow Chat backend can't blow the 5s "bot not responding" SLA Google enforces.

**Adversarial-review hardening**
- Real thread-root detection: compares the message id against the thread id across the `/messages/` vs `/threads/` namespaces. The previous `threadName === currentMessageName` check would never fire in production.
- `threadName` is validated against the canonical resource-name shape before interpolation — defense in depth against filter injection if the envelope is ever malformed/forged.
- The bot's own `_Centaur is thinking…_` ack messages and empty-content messages are filtered out of history (otherwise they'd appear as phantom assistant turns on every follow-up mention).
- 401/403 errors log as `chat_thread_history_scope_denied` (distinct from generic `_collect_failed`) so future scope misconfigurations are visible in logs instead of silently degrading every event.

**Type updates**
- New `ChatListMessage` type in `types.ts` modeling the rich shape returned by `spaces.messages.list` (separate from the outbound `GoogleChatMessage`).

## Test plan

- [x] `bun test` — 26 tests pass (12 new for collectThreadHistory: role inference, pagination + 50-cap, ack/empty filtering, injection rejection, thread-root detection, 503/403 error paths, bot mention stripping).
- [x] `bun run check:types` — only the preexisting `client.ts:127` error remains (verified on main; unrelated to this PR).
- [x] Live API verification on the production service account: minted a token with both scopes, called `spaces.messages.list?filter=thread.name="..."` against a real Workspace thread, got back the full 7-message thread including the bot's own prior turns.
- [ ] Deploy to `centaur-vps` and verify a follow-up @mention now references prior thread content correctly.

## Why this matters

Direct demo of the bug in production traffic: in the AI room, Joan asked `"What do you think about this response? @aigent"` referring to a 26KB analysis the bot had posted earlier in the thread. The bot replied `"I can review it, but I don't see the response text in this thread"` — because it literally couldn't. With this PR, the bot would have received all prior thread messages in the agent prompt.


jamalavedra:
Wrong repo — recreating on openfort-xyz/centaur.

PR #204, opened 2026-05-26
feat(tools): add BrightData public web tool

altanapps:
## Summary
- Adds a BrightData REST-backed Centaur tool for public web search, discovery, page scraping, and zone bandwidth stats.
- Uses BrightData's direct REST APIs (`api.brightdata.com`) rather than the MCP Streamable-HTTP server.
- Exposes curated methods only: `search`, `discover`, `discover_result`, `scrape_markdown`, `scrape_html`, and `session_stats`; no raw passthrough.

## Credentials
- Adds `BRIGHTDATA_API_TOKEN` as an HTTP secret scoped to `api.brightdata.com` and matched in the `Authorization` header.
- The tool sends `Authorization: Bearer <placeholder>`; iron-proxy substitutes the real token before requests leave the cluster.
- Optional local zone overrides are supported through `BRIGHTDATA_SERP_ZONE` and `BRIGHTDATA_UNLOCKER_ZONE`, defaulting to `serp_api` and `unlocker`.

## Client behavior
- `search` posts to `/request` with BrightData SERP URLs, `brd_json=1`, `format: "json"`, and `method: "GET"`.
- `scrape_markdown` and `scrape_html` use Web Unlocker via `/request`, with Markdown using `data_format: "markdown"`.
- `discover` uses the direct Discover API (`POST /discover`) and polls `GET /discover?task_id=...` by default so agents get results in one tool call.
- `session_stats` reads zone bandwidth data from `/zone/bw` for the configured SERP and Unlocker zones.
- Bounded 429 retry is retained for `/request` calls when `Retry-After` is present and small.
- Token-bearing values are redacted from exceptions and DEBUG-level logs.

## Safety
- Public web data only; no browser automation, Pro mode, or raw BrightData passthrough.
- The method surface is tested so new public methods are not accidentally exposed.

## Testing
- `uv run --project services/api ruff check tools/research/brightdata`
- `uv run --project services/api ruff format --check tools/research/brightdata`
- `uv run --project services/api pytest tools/research/brightdata/tests/test_client.py` (37 passed)
- `uv run --project tools/research/brightdata brightdata --help`
- ToolManager discovery smoke test confirmed `brightdata` loads with the expected method list and `BRIGHTDATA_API_TOKEN` header secret wiring.

## Note
- I could not run the final local in-cluster `/tools/brightdata/...` smoke test from this checkout because the local Kubernetes `centaur` namespace was not present at validation time.

### DIFF
--- a/tools/research/brightdata/.env.example
+++ b/tools/research/brightdata/.env.example
@@ -1,0 +1,1 @@
+BRIGHTDATA_API_TOKEN=your-brightdata-api-token
--- a/tools/research/brightdata/.env.example
+++ b/tools/research/brightdata/.env.example
@@ -1,1 +1,3 @@
 BRIGHTDATA_API_TOKEN=your-brightdata-api-token
+BRIGHTDATA_SERP_ZONE=serp_api
+BRIGHTDATA_UNLOCKER_ZONE=unlocker
--- a/tools/research/brightdata/cli.py
+++ b/tools/research/brightdata/cli.py
@@ -1,0 +1,84 @@
+"""CLI for the BrightData public web tool."""
+
+from __future__ import annotations
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+import json
+
+import typer
+from rich.console import Console
+
+from .client import BrightDataClient
+
+
+app = typer.Typer(name="brightdata", help="BrightData public web search, discovery, and scraping")
+console = Console()
+
+
+def _emit(payload: object, as_json: bool) -> None:
+    if as_json:
+        console.print_json(json.dumps(payload, default=str))
+    else:
+        console.print(payload)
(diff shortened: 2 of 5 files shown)


Zygimantass:
is there any reason we can't use the API and have to use the MCP provided? we'd much rather use the APIs directly

altanapps:
Yeah, we can use the API. Happy to switch. 

Background on the MCP choice: I'd been driving BrightData from a Claude Code harness via their MCP server, so wrapping that same surface was the path of least resistance for a first cut.
 
Going to rewrite the client against the REST API on this branch. I believe they have the same public methods as the MCP.

gakonst:
closing - feel free to reopen if you really want it / against latest!

PR #212, opened 2026-05-26
fix(chart): allow API egress to the Kubernetes apiserver (k3s/kind/kubeadm)

smartcontracts:
## Problem

On clusters where the in-cluster `kubernetes` Service remaps **443 → 6443** (k3s, k3d, kind, kubeadm), the API pod cannot reach the Kubernetes apiserver, and agent turns never start. This affects the documented [quickstart](https://github.com/paradigmxyz/centaur) local-dev path (k3d/k3s).

### Root cause

The API calls the apiserver to provision its own iron-proxy pod (`ensure_api_proxy_pod`) and agent sandboxes. Network-policy controllers such as k3s/kube-router enforce **egress** on the **post-DNAT** destination, so a connection to `10.43.0.1:443` is filtered as the apiserver endpoint port — commonly **`:6443`**. The API `NetworkPolicy` only allowed egress on `443`:

```yaml
- ports:
    - protocol: TCP
      port: 443   # to: all — covers external HTTPS, but NOT apiserver :6443
```

So apiserver traffic was silently dropped. (Contrast: the postgres egress rule works because its service port `5432` isn't remapped.)

### Symptom cascade (hard to diagnose)

1. `ensure_api_proxy_pod()` fails with `Cannot connect to host 10.43.0.1:443` — logged only as a swallowed warning (`api_proxy_pod_bootstrap_failed`).
2. The control-plane egress proxy `centaur-api-proxy` is never created.
3. `HTTPS_PROXY` on the API and slackbot then points at a non-existent host → the slackbot cannot reach Slack and agent turns never run.

## Fix

Add a configurable `networkPolicy.apiServer` egress rule to the API `NetworkPolicy`:

```yaml
networkPolicy:
  apiServer:
    cidr: 0.0.0.0/0
    ports: [443, 6443]
```

- Managed control planes (EKS/GKE/AKS) keep the `kubernetes` endpoint on `443` and are unaffected.
- k3s/k3d/kind/kubeadm now work out of the box.
- Operators can tighten `cidr` to a stable apiserver address, or set `ports: []` to opt out (e.g. when another egress rule already covers it).

## Verification

- `helm lint contrib/chart` passes; `helm template` renders the new egress rule on the API NetworkPolicy.
- On a live **k3d** cluster (apiserver endpoint `:6443`): before the fix the API pod could not reach `10.43.0.1:443` (egress dropped) and `centaur-api-proxy` was absent. After applying the rule and restarting the API, the API reaches the apiserver, `ensure_api_proxy_pod()` succeeds (`api_proxy_deployment_ready`), `centaur-api-proxy` is created, and the slackbot → Slack path works end-to-end.


Zygimantass:
closed since we refactored the API server

PR #235, opened 2026-05-27
fix: continue active Slack bot threads

vidhathat:
## Summary

Allow Centaur to continue an active Slack thread without requiring every follow-up reply to mention the bot again. The handoff remains gated for safety: unmentioned replies are only processed when the event is a thread reply and the collected thread history already includes a Centaur/bot assistant message.

This keeps top-level channel messages mention-gated while making ongoing bot conversations feel natural inside Slack threads.

## Changes

- Adds explicit normalized Slack event flags for thread replies and whether the bot is already present in the thread.
- Uses those flags to hand off either direct mentions or active-thread continuations.
- Backfills thread history for unmentioned replies so the bot-presence gate can be evaluated.
- Preserves the current Slackbot tracing/ignore-reason instrumentation from upstream main.
- Covers the continuation case in Slack normalization tests and updates existing event fixtures.

## Validation

- `pnpm --filter slackbot test`

### DIFF
--- a/services/slackbot/src/centaur/handoff.test.ts
+++ b/services/slackbot/src/centaur/handoff.test.ts
@@ -38,6 +38,8 @@ describe('CentaurHandoff', () => {
         channel_id: 'C123',
         thread_ts: '1778883099.579529',
         is_mention: true,
+        is_thread_reply: false,
+        bot_in_thread: false,
         parts: [{ type: 'text', text: 'hello' }],
         slack: {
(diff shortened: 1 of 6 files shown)


goksu:
Common usage pattern we observe is actually when multiple team members are discussing a topic in thread with Centaur triggers along the way. This looks like it would break that model as every response in thread would trigger Centaur response. Closing due to this. Let me know if that is not accurate!

PR #243, opened 2026-05-27
fix(slackbot): validate Slack Block Kit payloads

goksu:
## Summary
- add a Slack Block Kit validation boundary for message blocks and stream chunks
- route Slack write APIs through typed SlackBlock and SlackStreamChunk helpers
- update @slack/types and remove the undocumented plan_id field from generated plan blocks

## Tests
- bun run check:types
- bun test src/*.test.ts src/**/*.test.ts
- git diff --check


goksu:
Closing this as we are going to rework core renderers soon.

Issue #211, opened 2026-05-26
Bare-pod backend: pause_by_id deletes pod, resume_by_id is a no-op — suspended sessions are unrecoverable


j-s:
## Problem

`SandboxBackend.pause_by_id()` defaults to calling `stop_by_id()`, which deletes the pod entirely. `resume_by_id()` is a no-op (returns `None`). The `KubernetesExecutorBackend` inherits both defaults without overriding either.

Meanwhile, `get_or_spawn()` in `agent.py:846-873` treats `state='suspended'` as resumable: it calls `resume_by_id()` (no-op), checks status (pod is gone), and raises `RuntimeError("failed to resume suspended sandbox: ...")` — surfaced to the end user as "Failed to start the agent runtime."

Only the CRD-based `KubernetesAgentSandboxBackend` actually supports pause/resume (replicas 0→1), but there is no guard preventing the suspend codepath from running on backends that don't support it.

## Impact

Every deployment using the default bare-pod backend accumulates `suspended` sessions in the DB (we had 2,206) whose pods are already deleted. Any user message to one of those threads permanently errors until the DB row is manually cleaned up.

## Suggested fix

Either:
1. Have `get_or_spawn()` fall through to spawn a new sandbox when resume fails (instead of raising), clearing `agent_thread_id` since the session state died with the pod
2. Or have `KubernetesExecutorBackend` override `pause_by_id` to set DB state to `gone` instead of `suspended`, so resume is never attempted

Option 1 is more defensive and handles edge cases in the CRD backend too. We're running it as a hotfix now.

## Relevant code

- `services/api/api/sandbox/base.py:135-141` — base class defaults
- `services/api/api/agent.py:846-873` — resume logic that assumes pod exists
- `services/api/api/sandbox/kubernetes.py` — bare-pod backend, no pause/resume override

zac-idiliq:
Third independent production confirmation (plain `kubernetes` backend, chart v0.1.49, codex harness). We hit this 2026-06-09 with 11 stuck `suspended` sessions — every previously-working Slack thread errored `failed to resume suspended sandbox` on its next mention while new threads worked fine. Diagnosis matched this issue exactly (`base.py:135-141` defaults inherited un-overridden, `agent.py` ~846-873 re-raise).

Two additions to what's written above:

[...]

PR #247, opened 2026-05-27
fix: avoid cold-resuming agent threads across sandboxes

odyslam:
## Summary
- do not pass prior `agent_thread_id` values into any cold-spawned sandbox
- keep Centaur delivery state such as `last_delivered_id` while clearing sandbox-local agent thread state
- add focused regression coverage across Codex, Amp, and Claude Code cold-spawn paths

## Context
We hit a Slack runtime failure after a stale sandbox was cleared and Centaur cold-spawned a fresh sandbox while still passing the previous `agent_thread_id` as `resume_thread_id`. For Codex this surfaced as `no rollout found for thread id ...`, but the underlying lifecycle rule is not Codex-specific: an `agent_thread_id` belongs to the sandbox/runtime that produced it. Once Centaur cold-spawns a replacement sandbox, the durable source of context should be Centaur chat/message state, not harness-private thread state from the prior sandbox.

This is related to #211, but distinct: #211 covers suspended bare-pod resume failing before a new sandbox is spawned. This PR covers the later cold-spawn path where a fresh sandbox is created but receives stale local runtime state from a deleted/stopped sandbox. PR #84 is auth transport only; PR #85 handles explicit Slack stop/cancel state, not implicit cold-spawn after stale runtime cleanup.

## Verification
- `uv run --project services/api ruff check services/api/api/agent.py services/api/tests/test_agent_control_plane.py`
- `uv run --project services/api pytest services/api/tests/test_agent_control_plane.py::test_get_or_spawn_does_not_resume_agent_thread_on_cold_spawn -q`

### DIFF
--- a/services/api/api/agent.py
+++ b/services/api/api/agent.py
@@ -944,18 +944,23 @@ async def get_or_spawn(
     backend = get_backend()
     await _evict_idle_sessions_for_capacity(backend)
     trace_id = old_trace_id or thread_trace_id or str(uuid.uuid4())
+    if old_agent_thread_id:
+        log.info(
+            "cold_spawn_resume_skipped",
+            thread_key=thread_key,
+            engine=resolved_engine,
+            reason="agent_thread_state_belongs_to_previous_sandbox",
+        )
     session = await backend.create(
         thread_key,
         effective_harness,
         resolved_engine,
         persona=resolved_persona,
         repo=repo,
-        resume_thread_id=old_agent_thread_id or None,
+        resume_thread_id=None,
         trace_id=trace_id,
     )
     session.trace_id = trace_id
-    if old_agent_thread_id:
-        session.agent_thread_id = old_agent_thread_id
     _get_runtime(session.sandbox_id)
     log.info(
         "pipe_session_spawned", thread_key=thread_key, sandbox=session.sandbox_id[:12]
@@ -966,7 +971,7 @@ async def get_or_spawn(
         session,
         harness=session.harness,
         engine=session.engine,
-        agent_thread_id=old_agent_thread_id,
+        agent_thread_id="",
         last_delivered_id=old_last_delivered_id,
         inflight_turn_id=old_inflight_turn_id,
         inflight_turn_input=old_inflight_turn_input,
--- a/services/api/tests/test_agent_control_plane.py
+++ b/services/api/tests/test_agent_control_plane.py
@@ -61,6 +61,70 @@ async def _insert_assignment(db_pool, thread_key: str, generation: int = 1) -> N
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("engine", "stale_agent_thread_id"),
+    [
+        ("codex", "019e69c0-05c3-7841-9340-2cafeec83972"),
+        ("amp", "T-123"),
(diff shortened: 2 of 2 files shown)


odyslam:
I rechecked the repro and the key detail is that this is **not** the suspended-sandbox resume path. It is the cold-spawn path after a previous runtime is gone/stopped while Centaur still has the old harness-local `agent_thread_id`.

Repro from Slack:

1. Run unpatched `origin/main`.
2. Set the deployment default harness to Codex.
3. Ensure Slack Events API, workflow runner, and Codex auth are working.
4. Start a new Slack thread without any harness/persona flag:

   `@Phai reply exactly: FIRST-CODEX-THREAD-OK`

5. Wait for the successful reply.
6. Confirm Centaur persisted a Codex-local thread id:

   ```sql
   select sandbox_id, state, agent_thread_id
   from sandbox_sessions
   where thread_key = '<slack:<team_id>:<channel_id>:<root_ts>';
   ```

   `agent_thread_id` should be non-empty.

7. Force the backing runtime to disappear without clearing replay state:

   ```bash
   kubectl -n centaur delete pod <sandbox_id>
   ```

8. Reply in the same Slack thread, still with no harness/persona flag:

   `@Phai continue and reply exactly: SECOND-CODEX-COLD-SPAWN-OK`

On unpatched main, `get_or_spawn` cold-spawns a fresh sandbox but passes the previous sandbox’s `agent_thread_id` as `resume_thread_id`. Codex then tries to resume harness-local state that does not exist in the fresh sandbox and fails with:

```text
no rollout found for thread id <old-agent_thread_id>
```

DB confirmation:

```sql
select execution_id, status, terminal_reason, error_text
from agent_execution_requests
where thread_key = '<thread_key>'
order by created_at desc
limit 5;
```

Expected bad row:

```text
status = failed_permanent
terminal_reason = harness_error
error_text = no rollout found for thread id ...
```

This PR fixes that exact invalid handoff. On a cold spawn, a fresh sandbox should preserve Centaur-owned state such as Slack thread context, delivery cursor, and last result, but it must not receive a previous sandbox’s harness-local `agent_thread_id`.

I also validated the behavior manually:

Unpatched `origin/main`:

```text
backend.create.resume_thread_id='019e69c0-05c3-7841-9340-2cafeec83972'
returned_session.agent_thread_id='019e69c0-05c3-7841-9340-2cafeec83972'
db_insert.agent_thread_id='019e69c0-05c3-7841-9340-2cafeec83972'
db_insert.last_delivered_id='msg-prev'
RESULT=BUG_REPRODUCED_STALE_AGENT_THREAD_ID_SENT_TO_FRESH_SANDBOX
```

With this PR:

```text
backend.create.resume_thread_id=None
returned_session.agent_thread_id=''
db_insert.agent_thread_id=''
db_insert.last_delivered_id='msg-prev'
RESULT=FIX_VALIDATED_FRESH_SANDBOX_DOES_NOT_RECEIVE_STALE_AGENT_THREAD_ID
```

The important invariant is: warm same-runtime continuation may use harness-local ids; cold new-runtime spawn must not. Once the previous sandbox is gone, the durable source of context is Centaur’s own thread/message state, not Codex’s local rollout id from that old sandbox.

PR #251, opened 2026-05-27
fix: exempt Slackbot internal delivery from short timeout

odyslam:
## Summary
- keep the 5s timeout for Slack-facing webhook/command requests
- exempt internal Slackbot delivery endpoints from that short timeout so harness-event streaming can survive slower Slack API calls
- add coverage for the timeout routing rules

## Test plan
- pnpm --filter slackbot test

## Production note
- Hotfixed the Latitude deployment by rebuilding/importing `centaur-slackbot:latest` and restarting only the Slackbot deployment.


gakonst:
I think #249 fixes this

PR #253, opened 2026-05-28
fix(workflows): clone API env into workflow-run pod; drop per-run iron-proxy

mslipper:
Workflows are trusted code, so the per-run iron-proxy and per-run NetworkPolicy bring-up added in #216 was overkill and flaky: workflow executor pods could exit before writing terminal state, leaving Slack turns silently abandoned (the regression that prompted setting `WORKFLOW_RUN_SANDBOX_ENABLED=0` as a hotfix).

This replaces that machinery with the simplest model that still gives workflows their own pod: the workflow-run pod inherits the API container's env (`env`, `envFrom`, volume mounts, volumes) at spawn time, so it shares the API's view of secrets, `HTTPS_PROXY` (pointing at the shared API iron-proxy), and the firewall CA mount. Only a handful of overrides are applied — disabling background loops and preventing recursion. `spawn_workflow_run` no longer touches ConfigMaps, Services, NetworkPolicies, or per-run proxy pods, and cleanup is correspondingly trivial.

`workflow_executor` now populates `app.state.db_pool` before invoking `_run_handler` so handler code that reaches into the FastAPI app state (`api.agent._get_pool`) finds a live pool — that was the next breakage uncovered once the pod actually ran.

NetworkPolicy is updated to permit workflow-run pods to reach postgres, slackbot, the shared API iron-proxy, and the API itself.

Validated end-to-end: `echo`, `slack_sync`, and `slack_thread_turn` all run cleanly in the new pod against a real cluster.

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.44
+version: 0.1.45
 appVersion: "0.1.0"
 dependencies:
   - name: connect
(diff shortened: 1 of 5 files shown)


0xdiid:
was having some issues and this fixed it, ty!

PR #260, opened 2026-05-28
fix(tool-server): run the sandbox sidecar without a Postgres pool

0xdiid:
The per-sandbox tool-server sidecar (#216) only ever authenticates HMAC sandbox tokens (sbx1.*), which need no database, and the chart's NetworkPolicy denies sandbox pods Postgres egress ("only the API process connects to the postgres listeners"). But tool_server_app.lifespan unconditionally opened an asyncpg pool and ran migrations, so on any NetworkPolicy-enforcing cluster (e.g. k3s) the sidecar crashed at startup with `dbmate ... connect: connection refused` and every agent tool call failed with a transport error. It only worked on kind, which does not enforce NetworkPolicy.

Gate the pool + migrations behind TOOL_SERVER_SKIP_DB, which _build_tool_server_container sets on the sidecar. verify_api_key now returns 401 when the pool is absent — DB-backed aiv2_* keys never reach the sidecar, so this only affects a misdirected non-sandbox request. The shared tool-server Deployment leaves the flag unset and keeps its pool.

### DIFF
--- a/services/api/api/deps.py
+++ b/services/api/api/deps.py
@@ -140,8 +140,12 @@ async def verify_api_key(
         )
         raise HTTPException(status_code=401, detail="Invalid or expired sandbox token")
 
-    # DB key lookup — all external callers use DB-backed aiv2_* keys
+    # DB key lookup — all external callers use DB-backed aiv2_* keys. The
+    # tool-server sidecar runs without a DB pool (sandbox tokens only), so a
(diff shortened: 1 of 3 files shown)


0xdiid:
@mslipper this alongside #253 is what I needed to restore my setup. This is admittedly very much so vibe coded, so would love a vibe check to see if this aligns with the new sandboxing

PR #261, opened 2026-05-28
fix: bridge workflow executor db pool

### LINKED ISSUE #245: Slack workflow worker fails with missing app.state.db_pool
## Summary
Slack-triggered `slack_thread_turn` workflow runs can fail immediately in the API workflow worker with:

```text
'State' object has no attribute 'db_pool'
```

When this happens, the Slack webhook is acknowledged by the Slackbot, but no final delivery is enqueued, so the bot appears not to respond in Slack.

## Impact
- Slack webhook POSTs return 200, masking the failure from Slack retries.
- `workflow_runs` for `slack_thread_turn` become terminal `failed`.
- `agent_final_delivery_outbox` has no corresponding Slack delivery rows for the failed runs.
- Users see no Slack response even though the ingress and Slackbot pod are healthy.

## Evidence from production incident
Recent failed rows in Postgres showed:

```sql
select run_id, workflow_name, status, created_at, started_at, completed_at, error_text
from workflow_runs
where workflow_name = 'slack_thread_turn'
order by created_at desc
limit 5;
```

All recent failures had `error_text = 'State' object has no attribute 'db_pool'`.

API logs also showed the final delivery poller successfully calling `/agent/final-deliveries/claim`, while Slackbot logs showed webhook POSTs to `/api/webhooks/slack` returning 200.

## Root cause
Workflow handlers execute from the durable workflow worker path, not from a normal FastAPI request path. Some shared handler code expects `api.app.state.db_pool` to exist, but the workflow worker only had its local DB pool and did not bridge that pool onto `api.app.state`.

## Expected behavior
`slack_thread_turn` workflow runs should have access to the same DB pool state expected by shared runtime code, complete normally, and enqueue final Slack delivery rows.

## Fix direction
Bridge the workflow worker DB pool into `api.app.state.db_pool` before invoking workflow handlers. A local hotfix did this in `services/api/api/workflow_engine.py` and added targeted regression coverage in `services/api/tests/test_workflow_engine_title.py`.

Local hotfix commit: `086ec913 Fix Slack workflow DB pool bridge`.

## Verification
Targeted regression test used for the hotfix:

```bash
uv run --project services/api pytest services/api/tests/test_workflow_engine_title.py -q
```

Result: `18 passed`.

odyslam:
## Summary
- expose the workflow worker pool on `api.app.state.db_pool` before workflow handlers run
- preserve an existing FastAPI lifespan-provided pool when present
- add regression coverage for the bridge behavior

## Context
Fixes #245. The one-shot workflow executor does not run FastAPI lifespan, but shared handler code can call helpers that expect `api.app.state.db_pool`. That crashes Slack workflow turns after Slack ingress returns 200.

## Tests
- `uv run --project services/api pytest services/api/tests/test_workflow_engine_title.py -q`
- `uv run --project services/api ruff check services/api/api/workflow_engine.py services/api/tests/test_workflow_engine_title.py`

### DIFF
--- a/services/api/api/workflow_engine.py
+++ b/services/api/api/workflow_engine.py
@@ -2358,6 +2358,22 @@ async def _load_checkpoints(
     }
 
 
+def _bind_app_db_pool(pool) -> None:
+    """Expose the workflow worker pool to shared API helpers.
+
+    The out-of-process workflow executor does not run FastAPI lifespan, so
+    ``api.app.state.db_pool`` is otherwise missing for workflow handlers that
+    call shared API helpers.
+    """
+    try:
+        from api.app import app as api_app
+    except Exception:
+        log.debug("workflow_app_db_pool_bind_skipped", exc_info=True)
+        return
+    if not hasattr(api_app.state, "db_pool"):
+        api_app.state.db_pool = pool
+
+
 async def _execute_run(pool, run_id: str) -> None:
     """Claim a specific run by ID and execute it."""
     worker_id = _new_worker_id()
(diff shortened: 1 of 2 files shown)


Zygimantass:

PR #267, opened 2026-05-28
fix(slackbot): stop reposting the answer tail after a durable final delivery

0xdiid:
#249 fixed fast-turn answer loss by folding the whole answer into the durable chat.stopStream blocks instead of a finalize-time appendStream that Slack drops. But the control plane still gauges live-delivery coverage by comparing its result_text length against the slackbot's streamed-char tally, and that tally counts only the live-streamed delta source chars. When result_text carries a few trailing characters the live deltas did not, coverage reads as incomplete and the fallback reposts result_text.slice(streamed_chars) — the answer's last 2-3 characters — as a separate thread message, duplicating text the durable block already shows.

Have the renderer report finalAnswerDurablyDelivered when the complete answer is composed into the durable blocks (durable blocks are never dropped by Slack), thread it through the codex harness-event response, and treat it as full coverage in _mark_execution_terminal so the fallback continuation is skipped. Streamed-live answers keep the existing coverage fallback as the cut-off safety net.

### DIFF
--- a/services/api/api/runtime_control.py
+++ b/services/api/api/runtime_control.py
@@ -1940,6 +1940,7 @@ async def _mark_execution_terminal(
     result_text: str,
     error_text: str | None,
     slackbot_streamed_answer_chars_override: int | None = None,
+    slackbot_final_answer_durably_delivered: bool = False,
 ) -> None:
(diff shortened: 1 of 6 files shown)


0xdiid:
Just deferring as discussed in slack

PR #278, opened 2026-05-28
feat: support pvc-backed repo cache

tarrencev:
## Summary
- add API support for mounting sandbox repo caches from a PVC via REPOS_PVC_CLAIM_NAME
- keep existing REPOS_PATH hostPath behavior for local/dev compatibility
- add Helm repoCache.persistence settings that create/use a PVC and switch repo-cache from a DaemonSet to a single Deployment in PVC mode
- add tests for hostPath, PVC, and conflicting repo volume sources

## Why
Managed Kubernetes clusters commonly reject sandbox hostPath mounts. PVC-backed repo caches keep the fast shared-clone workflow without granting sandboxes node filesystem access.

## Validation
- uv run pytest tests/test_sandbox_kubernetes_backend.py -k 'repo_cache or builds_pod_and_prompt_secret'
- uv run ruff check api/sandbox/kubernetes.py tests/test_sandbox_kubernetes_backend.py
- git diff --check
- helm template PVC mode with repoCache.persistence.enabled=true
- helm template legacy mode with repoCache.persistence.enabled=false

### DIFF
--- a/contrib/chart/templates/repo-cache.yaml
+++ b/contrib/chart/templates/repo-cache.yaml
@@ -5,13 +5,34 @@
 {{- $repoCacheImageRepository := default .Values.sandbox.image.repository .Values.repoCache.image.repository -}}
 {{- $repoCacheImageTag := default .Values.sandbox.image.tag .Values.repoCache.image.tag -}}
 {{- $repoCacheImagePullPolicy := default .Values.sandbox.image.pullPolicy .Values.repoCache.image.pullPolicy -}}
+{{- if and .Values.repoCache.persistence.enabled (not .Values.repoCache.persistence.existingClaim) }}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "centaur.repoCachePvcName" . }}
+  labels:
+{{ include "centaur.componentLabels" (dict "root" . "component" "repo-cache") | nindent 4 }}
+spec:
+  accessModes:
+{{ toYaml .Values.repoCache.persistence.accessModes | nindent 4 }}
+  {{- with .Values.repoCache.persistence.storageClassName }}
+  storageClassName: {{ . | quote }}
+  {{- end }}
+  resources:
+    requests:
+      storage: {{ .Values.repoCache.persistence.size }}
+---
+{{- end }}
 apiVersion: apps/v1
-kind: DaemonSet
+kind: {{ ternary "Deployment" "DaemonSet" .Values.repoCache.persistence.enabled }}
 metadata:
   name: {{ include "centaur.componentName" (dict "root" . "component" "repo-cache") }}
   labels:
 {{ include "centaur.componentLabels" (dict "root" . "component" "repo-cache") | nindent 4 }}
 spec:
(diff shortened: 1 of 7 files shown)


contrib/chart/templates/repo-cache.yaml
```diff
@@ -5,13 +5,34 @@
...
 apiVersion: apps/v1
-kind: DaemonSet
+kind: {{ ternary "Deployment" "DaemonSet" .Values.repoCache.persistence.enabled }}
 metadata:
   name: {{ include "centaur.componentName" (dict "root" . "component" "repo-cache") }}
   labels:
 {{ include "centaur.componentLabels" (dict "root" . "component" "repo-cache") | nindent 4 }}
 spec:
+{{- if .Values.repoCache.persistence.enabled }}
+  replicas: 1
```
0xprames, line 34: what we're saying here is: when persistence is enabled, make the repo-cache a single-replica deployment, which will only ever run on one node. 

The PVC will also default to `ReadWriteOnce` IIRC - which means only one node at a time will be able to write into the PV, and all sandbox pods that need to access the repo-cache *must be scheduled on that single node*

forcing the PVC to be `ReadWriteMany` idt is the right fix either - that implies that the underlying volume plugin supports it (usually a *FS: gluster, EFS/NFS etc). k3s default to `local-path-provisioner` iirc

0xprames (changes requested):

0xprames:
i think the `Daemonset` to single-replica `Deployment` with a PVC may be a breaking change 

left a comment explaining why it may be the case

contrib/chart/templates/repo-cache.yaml
tarrencev: Yeah that is true. It does assume sandboxes are scheduled on the same node. I'm not sure there is a simple / widely supported way to handle this with PVCs. We can change it to a `Daemonset` but I don't think that will resolve this issue, since any additional instance still wouldn't be able to mount the PVC.

A better solution might be to have a git mirror service and automatically configure git to pull from it rather than having sandboxes mount it directly.

Something like:

  1. repo-cache service with PVC-backed bare mirrors.
  2. Sandbox image includes global Git URL rewrite to repo-cache.
  
     ```
     git config --system url."http://centaur-repo-cache/repos/github.com/".insteadOf "https://github.com/"
     git config --system url."http://centaur-repo-cache/repos/github.com/".insteadOf "git@github.com:"
     git config --system url."http://centaur-repo-cache/repos/github.com/".insteadOf "ssh://git@github.com/"
     ```
  3. Repo-cache lazily mirrors/fetches repos on clone request.
  4. Sandbox clone works with normal Git commands.
  5. After clone/materialization, Centaur sets remote to push to the remote origin:

     `git remote set-url --push origin https://github.com/<org>/<repo>.git`

  6. GitHub auth remains via existing credential injection/proxy path.

0xprames: yeah def, I think that would make more sense than using a PVC

tarrencev:
Closing in favor of #294

PR #286, opened 2026-05-28
fix(tool-server): route the sandbox sidecar's DB pool through the iron-proxy

mslipper:
Makes the per-sandbox tool-server work under an **enforced** NetworkPolicy (paradigm does not currently enforce it; clusters that do — e.g. default k3s — hit these bugs).

**Sidecar DB pool via iron-proxy.** The sidecar opened an asyncpg pool and ran dbmate migrations against the core DB directly, which the sandbox NetworkPolicy denies, so it crashed at startup and every agent tool call failed. It now reaches the core DB through the per-sandbox iron-proxy: a core-DB listener is added to the proxy and the sidecar `DATABASE_URL` points at it. Real DB creds stay in the proxy pod; the sandbox only holds an `app_user` proxied DSN, injected into the tool-server container only (never the agent container). The sidecar opens its pool with `apply_migrations=False` (the API owns migrations).

**API egress to the k8s API server.** The chart default-deny egress only allowed `:443`, but the `kubernetes.default` ClusterIP is DNAT'd to the real endpoint port (6443 on k3s/kubeadm) and enforcing CNIs match egress post-DNAT, so the API could not reach the control plane and no iron-proxy came up. Adds an egress rule for the API-server port, configurable via `networkPolicy.apiServerPort` (default 6443).

**Dev tooling.** `just up k3s` imports locally-built images into k3s containerd.

Supersedes the skip-DB approach in #260 by @0xdiid, whose report identified the original bug.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -5,6 +5,9 @@ release := env_var_or_default("CENTAUR_RELEASE", "centaur")
 source := env_var_or_default("CENTAUR_IMAGE_SOURCE", "local")
 chart := "contrib/chart"
 dev_values := "contrib/chart/values.dev.yaml"
+# Command used to import images into k3s's containerd. Override for rootless or
+# remote setups, e.g. CENTAUR_K3S_CTR="k3s ctr" or "ssh host sudo k3s ctr".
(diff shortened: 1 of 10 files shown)


0xdiid:
Lgtm, will close mine!

PR #292, opened 2026-05-29
fix(tool-server): split sandbox pool reset into single statements

zasulskii:
## Problem

After #286 the sandbox tool-server sidecar reaches Postgres through the
per-sandbox iron-proxy. iron-proxy 0.42+ blocks multi-statement SQL by
policy. asyncpg's default `Connection.reset()` concatenates several
cleanup statements (`SELECT pg_advisory_unlock_all(); CLOSE ALL; UNLISTEN *; RESET ALL;`)
into one query and sends them on every `pool.release()`.

Result: every `pool.release()` — and therefore every `fetchrow` via
`async with pool.acquire()` — raises

```
asyncpg.exceptions.PostgresSyntaxError: blocked by iron-proxy policy:
multi-statement queries not permitted
```

and the sidecar returns HTTP 500 on every tool call (slack,
company_context, wrike, etc.).

### Traceback

```
File "/app/api/tool_manager.py", line 2338, in call_tool
    parent_context = await _active_execution_parent_context(request, sandbox_claims)
File "/app/api/tool_manager.py", line 1140, in _active_execution_parent_context
    row = await pool.fetchrow(...)
File "asyncpg/pool.py", line 239, in release
    await self._con.reset(timeout=budget)
File "asyncpg/connection.py", line 1571, in reset
    await self.execute(reset_query)
asyncpg.exceptions.PostgresSyntaxError: blocked by iron-proxy policy:
    multi-statement queries not permitted
```

The bug is silent on clusters without iron-proxy on the DB path, which
is why #286's tests don't catch it.

## Fix

Override the connection class for the sandbox pool only
(`apply_migrations=False`) with one that runs the reset query as one
`asyncpg.execute()` per statement. Semantics are preserved; iron-proxy
accepts each statement individually.

The API and shared tool-server keep asyncpg's default connection class
since they don't sit behind a multi-stmt-blocking proxy.

## Reproduction

Any sandbox-agent tool call on a cluster running iron-proxy 0.42.0-rc.2
(the centaur 0.1.46 chart's default) after #286. Spin up a sandbox via
the slack mention path and call any tool — the sidecar's tool-server
returns 500 with the traceback above in its logs.

## Tests

`services/api/tests/test_db_pool_reset.py`:

- Pin that `reset()` executes each cleanup statement individually
  (regression guard against re-introducing a multi-statement reset).
- Pin that empty / blank-line reset queries are no-ops / skipped.
- Pin that `create_pool(apply_migrations=False)` wires the override and
  `apply_migrations=True` does not (the API/shared tool-server path
  stays on stock asyncpg).

## Refs

- Introduced by: #286 (07c03fb)
- iron-proxy version: 0.42.0-rc.2 (the policy enforcing the block)

### DIFF
--- a/services/api/api/db.py
+++ b/services/api/api/db.py
@@ -96,18 +96,56 @@ def _dbmate_url(database_url: str) -> str:
     return f"{database_url}{sep}sslmode=disable"
 
 
+class _SingleStmtResetConnection(asyncpg.Connection):
+    """Connection that runs the pool-release reset as single-statement queries.
+
+    asyncpg's default ``Connection.reset`` joins ``SELECT
+    pg_advisory_unlock_all()``, ``CLOSE ALL``, ``UNLISTEN *`` and
+    ``RESET ALL`` into a single multi-statement query. The per-sandbox
+    iron-proxy added in #286 blocks multi-statement SQL by policy, so every
+    ``pool.release()`` from the tool-server sidecar raises
+    ``asyncpg.exceptions.PostgresSyntaxError: blocked by iron-proxy policy:
+    multi-statement queries not permitted`` and the sidecar returns HTTP 500
+    on every tool call.
+
+    This subclass keeps asyncpg's reset semantics (rollback open txn, clear
+    listeners, then run the reset query) but executes each reset statement
+    individually so iron-proxy accepts them.
+    """
+
+    async def reset(self, *, timeout=None):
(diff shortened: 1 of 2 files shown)


mslipper:
Hello! Thanks for reporting this. I added multi-statement support to iron-proxy so this should no longer be needed. Please let me know if you see any other issues like this.

PR #300, opened 2026-05-29
fix(slackbot): stop splicing/duplicating claude-code answers (#268)

### LINKED ISSUE #268: [bug] slackbot CodexSessionRenderer duplicates content when harness is claude-code: live commentary + final answer concatenated in same Slack message
## Summary

Slackbot's `CodexSessionRenderer` concatenates the live-streamed commentary and the final answer into the same Slack message body when the harness is `claude-code`. Every reply that takes more than a couple of seconds visibly contains the agent's output **twice** — first in raw markdown as it streamed, then again in slack-mrkdwn after final post-processing.

## Repro

1. Deploy at `main-sha-3588318` (api / agent / iron-proxy / slackbot, all same tag).
2. Set the harness to `claude-code` — either chart-default `api.defaultHarness: claude-code` or env `CENTAUR_DEFAULT_HARNESS=claude-code`.
3. Mention the bot with any question that takes more than ~1 sentence to answer with citations, e.g. `@archie is base azul out yet?`

## Observed

The single Slack message body ends up with the same content twice:

```
> archie: base · claude-opus-4-7
Yes — Base
Azul activated on mainnet on May 13, 2026 ([Bitget](https://news.google.com/rss/articles/CBMi...))...

Yes — Base Azul activated on mainnet on May 13, 2026 (<https://news.google.com/rss/articles/CBMi...|Bitget>)...
```

- **First half** uses markdown link syntax `[Bitget](URL)` — that's the agent's raw output as it streamed.
- **Second half** uses Slack mrkdwn `<URL|Bitget>` — the same content post-processed for Slack.

Both are in the same Slack message (single `chat.update` body), concatenated.

Concrete production sample: `wfr_3b4ea68658684270`, agent session `01KSPFP8WKD5HEX6NG6YW4S4G2`. Output `metadata.slackbot_live_delivery=true`, `metadata.slackbot_agent_session_id=01KSPFP8WKD5HEX6NG6YW4S4G2`. The `agent_final_delivery_outbox` table has no row for that thread, so this is not a live-vs-final-outbox doubling — it's purely inside the slackbot renderer.

Also from our slackbot pod logs for that session, two `slack_codex_terminal_event_received` events fire in a row, with the second returning `502` (`http_request status=502`). The slackbot's own dedup *almost* catches it (`slack_codex_terminal_event_ignored_after_done` fires for subsequent events) but the doubled rendering happens before that suppression kicks in.

## Diagnosis

`services/slackbot/src/slack/codex-session.ts` `CodexSessionRenderer` keeps two text accumulators:

- `state.commentaryText` — appended via `publishPendingAssistantText` as streaming `agentMessage` events arrive
- `state.answerText` — the final-answer block

At line 250:

```ts
const { streamedTextChars } = await this.renderer.done(agentSessionId, {
  streamFinalUpdates: true,
  ...
})
```

`streamFinalUpdates: true` flushes the final answer through the same streaming path *in addition to* what already accumulated in `commentaryText`. The renderer was designed against codex's emit pattern (separate `agentMessage` events that distinguish commentary vs final answer). Claude Code's SDK doesn't make that distinction — every assistant chunk arrives as a streaming `assistant` message, so it all lands in `commentaryText`, and then the terminal payload *additionally* fills `answerText`. The final `chat.update` renders both buffers concatenated.

## Expected

For `claude-code` harness, the final answer should replace the live-streamed content, not append to it. Possible directions:

- Pass `streamFinalUpdates: false` when the active harness is `claude-code`, or
- Suppress `commentaryText` flushes once `answerText` is non-empty, or
- Treat the `claude-code` stream as a single replace-on-write buffer (drop the commentary/answer split altogether for this harness).

## Affects

- Self-hosted deployments running `claude-code` as the default harness (which is the chart default since #170).
- Reproduces on `main-sha-3588318`. Has been present since at least `main-sha-6bf95df` (the earlier pin we ran).
- Not specific to a particular Slack workspace or channel — reproduced in two of ours independently.

oddharsh:
Fixes #268.

### Root cause
`CodexSessionRenderer.publishPendingAssistantText` computes the live answer delta by raw length offset:

```ts
const delta = state.answerText.slice(state.streamedAnswerText.length)
```

That assumes `answerText` only ever grows by append. It doesn't: `recomposeBuffers()` rebuilds `answerText` from scratch on every event, and the harness answer buffer (claude-code `assistant` snapshots) can be **replaced wholesale** by a later canonical snapshot (the `assistantEventLooksCanonical` branch in `applyAgentMessageUpdate`). Once the buffer is replaced, slicing it by the *previously streamed* length splices at the wrong byte boundary:

- **Character corruption** — the reported `loops` → `loours`, `you are` → `youare`, `L2` → `L2k`.
- **Duplication** — the stale streamed prefix no longer matches the canonical answer, so the renderer's `startsWith()`-based final-block dedup (`unstreamedMarkdownAfterLivePrefix`) misses and falls back to its own `markdown.slice(streamedTextSourceChars)`, re-emitting the body.

So both symptoms in #268 trace to the same unguarded offset slice.

### Fix
Mid-turn, stream **only the codex per-item answer**. Each per-item buffer gets a clean canonical replacement on `item.completed`, so the composed item text only ever grows by append and is safe to offset-slice. The replaceable **harness** answer is held back until `done()` (`opts.force`), when it's final and nothing can contradict it. A `startsWith` guard keeps the slice honest in both paths — if the text we're about to stream doesn't extend what we already streamed, we skip rather than splice at a bad offset.

```ts
const streamableAnswer = opts.force ? state.answerText : compose(state.answerByItemId, '')
if (!streamableAnswer.startsWith(state.streamedAnswerText)) return
const delta = streamableAnswer.slice(state.streamedAnswerText.length)
```

### Trade-off (your call)
With this change, the claude-code answer **text** renders when the turn completes rather than streaming token-by-token mid-turn. Tool calls, plan updates, and thinking still stream live; only the final answer body waits for `done()`. I picked correctness over live-typing as the safe default — but if you'd rather preserve live answer streaming, the alternative is a heavier renderer-side change (detect "streamed prefix no longer matches canonical" and *replace* the message instead of appending). Happy to take that direction instead.

### Tests
- New regression test deterministically reproduces the splice (interim snapshot streamed live, then a diverging canonical snapshot): fails on `main`, passes here.
- Full slackbot suite green: `96 pass, 0 fail`. `check:types` clean.

### DIFF
--- a/services/slackbot/src/slack/codex-session.test.ts
+++ b/services/slackbot/src/slack/codex-session.test.ts
@@ -1840,6 +1840,65 @@ describe('CodexSessionRenderer', () => {
       task_count: 1
     })
   })
+
+  it('does not splice-corrupt or duplicate when a canonical snapshot diverges from streamed text (#268)', async () => {
+    // #268: live answer streaming uses a raw length-offset slice
+    // (state.answerText.slice(state.streamedAnswerText.length)). That assumes
+    // answerText only ever GROWS BY APPEND. But recomposeBuffers() rebuilds
+    // answerText from scratch each event, and the assistant-event merge can
+    // REPLACE the harness buffer (canonical branch). When the recomposed answer
+    // no longer extends what we already streamed, slicing by length splices at
+    // the wrong byte boundary -> the observed character corruption
+    // ("loops" -> "loours"), and the stale streamed prefix then defeats the
+    // renderer's startsWith()-based final-block dedup, duplicating the body.
+    const calls: Array<{ method: string; params: any }> = []
+    const client = makeFakeSlackClient(calls)
+    const { sessionId } = await new AgentSessionRenderer(client as any).open({
+      channel: 'C123',
+      parentTs: '1778866921.505479',
+      recipientTeamId: 'T123',
+      recipientUserId: 'U123',
+      title: 'Centaur execution'
+    })
+    const renderer = new CodexSessionRenderer(client as any)
+
+    // A running command creates a plan, which enables live answer streaming
+    // immediately (no 500ms pre-stream grace) — matching a turn that ran tools.
+    await renderer.event(sessionId, {
+      type: 'item.started',
+      item: { id: 'cmd-1', type: 'commandExecution', command: 'echo hi' }
+    })
+
+    // Opus-4-8 emits an interim (non-canonical) answer snapshot, streamed live.
+    await renderer.event(sessionId, {
+      type: 'assistant',
+      message: { content: [{ type: 'text', text: 'I run durable loops. ' }] }
+    })
+    // The canonical snapshot (carries message.model/usage) does NOT extend the
+    // interim text — the harness re-tokenized/normalized it. This is the event
+    // that triggers the buggy offset slice.
+    const canonical = 'I run resilient loops that survive restarts and catch CI failures.'
+    await renderer.event(sessionId, {
+      type: 'assistant',
+      uuid: 'u1',
+      request_id: 'r1',
+      session_id: 's1',
+      message: {
+        id: 'm1',
+        model: 'claude-opus-4-8',
+        usage: { input_tokens: 1, output_tokens: 1 },
+        content: [{ type: 'text', text: canonical }]
+      }
+    })
+    await renderer.event(sessionId, { type: 'result', result: canonical })
+
+    const visible = visibleMarkdown(calls)
+    // The canonical answer must appear verbatim, exactly once, with no
+    // mid-word splice from the superseded interim snapshot.
+    expect(visible).toContain(canonical)
+    expect(countOccurrences(visible, 'I run')).toBe(1)
+    expect(visible).not.toContain('durable loops.  that')
+  })
 })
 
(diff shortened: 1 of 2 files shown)


oddharsh:
Closing this — live-testing on our cluster showed it targets the wrong layer.

Reproduced a corrupted/duplicated reply on `claude-opus-4-8` and pulled the slackbot trace: `streamed_answer_chars_before_event: 0` (nothing streamed mid-turn) and the done()-time slice is `answerText.slice(0)` (a clean whole-string copy), yet `state.answerText` still rendered garbled — fragments of different sentence positions welded together, plus a link URL appearing once raw and once post-processed. So the corruption is baked into the buffer when the ~25 opus `assistant` snapshots are *merged* (the cumulative-vs-delta classification in `applyAgentMessageUpdate`), not in the offset-slice streaming this PR guards.

The regression test here still demonstrates the offset slice is unsafe, but it's not the operative cause for the Opus reports. Will follow up with a fix at the merge layer once I've captured the raw assistant-event payloads to confirm Opus's emit pattern.

PR #301, opened 2026-05-29
fix(slack): list_bot_channels via users.conversations to avoid whole-workspace scan

### LINKED ISSUE #291: list_bot_channels walks the entire workspace via conversations.list — Slack read tools time out on large workspaces
## Summary

On a workspace with a large number of channels, every Slack **read-path** tool call (`get_channel_history`, and `search_messages` when it falls back to the local scan) hangs and frequently never returns within the agent's tool-call time budget. The culprit is `SlackClient.list_bot_channels`, which walks the **entire workspace** via `conversations.list` and filters `is_member` client-side, rather than asking Slack only for the bot's channels via `users.conversations`.

## Root cause

[`list_bot_channels`](https://github.com/paradigmxyz/centaur/blob/07c03fbb7e639adcd694ff62b86708cb4e66c45d/tools/productivity/slack/client.py#L439-L503) paginates `conversations.list` over `public_channel,private_channel` and keeps only members:

```python
response = self._retry_on_ratelimit(
    self._client.conversations_list,
    types="public_channel,private_channel",
    limit=min(limit - len(channels), 200),
    cursor=cursor,
    exclude_archived=True,
)
...
for channel in response.get("channels", []):
    if channel.get("is_member", False):
        channels.append({...})

cursor = response.get("response_metadata", {}).get("next_cursor")
if not cursor or len(channels) >= limit:
    break
```

Because a bot is typically a member of only a handful of channels, `len(channels)` almost never reaches `limit`, so the loop's only real exit condition is **cursor exhaustion** — i.e. it pages through every public + private channel in the workspace. On a large workspace this is thousands of `conversations.list` calls, which:

1. exceeds the agent tool-call time budget, so the tool returns nothing; and
2. never reaches [`self._save_channel_cache(result)`](https://github.com/paradigmxyz/centaur/blob/07c03fbb7e639adcd694ff62b86708cb4e66c45d/tools/productivity/slack/client.py#L502) (cache is only written after the full walk completes), so **the next call re-scans from scratch** — the cache never warms.

This is reached on common paths even when the caller already has a channel ID:
- [`_resolve_channel_name`](https://github.com/paradigmxyz/centaur/blob/07c03fbb7e639adcd694ff62b86708cb4e66c45d/tools/productivity/slack/client.py#L231-L240) calls `list_bot_channels(limit=1000)` purely to turn an ID into a display name;
- the `_search_messages_local` fallback (used when the token lacks `search:read`) scans this list;
- `gather_context`'s Slack grab walks it.

## Evidence (self-hosted deploy)

Captured at our egress proxy over a ~3h window, broken down per Slack method:

| Method | Calls | Outcome |
|---|---|---|
| `conversations.history` | 8 | **all 200**, max 222 ms |
| `conversations.list` | **1236** | 1066 × 200 + **170 × 429** (rate-limited) |
| `users.conversations` | 0 | — |

So the actual message fetch (`conversations.history`) is healthy and fast; effectively 100% of the latency and all the throttling come from the full-workspace `conversations.list` walk. (Note Slack's 2025 tightening of `conversations.list` rate limits makes this much worse than it used to be.)

## Suggested fix

Use [`users.conversations`](https://api.slack.com/methods/users.conversations) instead of `conversations.list` + client-side `is_member`. `users.conversations` returns **only the channels the bot is a member of**, so it's normally a single page regardless of workspace size, and removes the need for the `is_member` filter entirely. `types="public_channel,private_channel"` carries over directly.

One caveat worth confirming: the dicts `list_bot_channels` returns today include `purpose`, `topic`, and `member_count`. If `users.conversations` objects don't carry those fields in a given workspace, callers that need them would require a `conversations.info` hydration — though in the hot paths above (`_resolve_channel_name`, local search) only `id`/`name`/`is_private` are actually used, so the cheap version is sufficient there.

## Questions

- Is anyone else seeing Slack history/search tool calls hang or time out on larger workspaces? Curious whether this is widespread or specific to channel-count.
- Was `conversations.list` chosen over `users.conversations` deliberately (e.g. to surface channels the bot is *not* in)? If so, would you take a PR switching the membership-listing path to `users.conversations` while preserving current return-field parity?

Happy to open a PR if the direction looks right.

kmx411:
Fixes #291.

## Problem

`SlackClient.list_bot_channels` lists the bot's channels by paginating `conversations.list` over the **entire workspace** and filtering `is_member` client-side. Because the loop only exits on cursor exhaustion (a bot is a member of far fewer channels than `limit`), it walks every public + private channel in the workspace on every cold call.

On a large workspace this is thousands of `conversations.list` calls (we measured 1,236 in a ~3h window, 170 of them `429`-throttled). It exceeds the agent tool-call budget, so read-path tools — `get_channel_history` name resolution, the `_search_messages_local` fallback, and `gather_context`'s Slack grab — hang and return nothing. And because `_save_channel_cache` is only reached after the full walk completes, the cache never warms, so the next call re-scans from scratch.

## Fix

Use [`users.conversations`](https://api.slack.com/methods/users.conversations), which returns only the conversations the bot is a member of — normally a single page regardless of workspace size. This removes the whole-workspace pagination and the client-side `is_member` filter entirely.

- `types="public_channel,private_channel"` carries over unchanged, so private channels are still included.
- The returned dict shape is unchanged; field extraction already uses `.get(..., default)`, so any field a given workspace doesn't hydrate degrades gracefully rather than breaking callers.
- `_list_etl_channels` (which intentionally lists *all* public channels visible to the ETL user token, not just member channels) is left on `conversations.list`.

## Test

Adds `test_list_bot_channels_uses_users_conversations`: asserts the call goes through `users.conversations` with the expected params, that `conversations.list` is not called, and that every returned conversation is kept (membership implied) with fields and sort preserved. Full `tools/productivity/slack/tests/test_client.py` suite passes (34 tests); `ruff check` and `ruff format --check` clean.

### DIFF
--- a/tools/productivity/slack/client.py
+++ b/tools/productivity/slack/client.py
@@ -513,6 +513,13 @@ class SlackClient:
         as the fallback when native search lacks ``search:read``, and
         ``gather_context``'s Slack grab walks the bot's member channels).
 
+        Uses ``users.conversations``, which returns only the conversations the
+        bot is a member of. ``conversations.list`` would instead page through
+        every channel in the workspace and filter ``is_member`` client-side —
+        an O(workspace) scan that, on large workspaces, never finishes inside
+        the tool-call budget (so the cache below never warms and every call
+        re-scans the whole workspace from scratch).
+
         Args:
             limit: Maximum channels to return
             force_refresh: Ignore cache and fetch fresh data
@@ -533,7 +540,7 @@ class SlackClient:
         while True:
             try:
                 response = self._retry_on_ratelimit(
-                    self._client.conversations_list,
+                    self._client.users_conversations,
                     types="public_channel,private_channel",
                     limit=min(limit - len(channels), 200),
                     cursor=cursor,
@@ -542,22 +549,24 @@ class SlackClient:
             except SlackApiError as e:
                 self._raise_slack_api_error(
                     e,
-                    slack_method="conversations.list",
+                    slack_method="users.conversations",
                     access_path="bot_token",
                 )
 
+            # users.conversations only returns conversations the bot belongs
+            # to, so membership is implied — no client-side is_member filter
+            # (and no whole-workspace pagination) needed.
             for channel in response.get("channels", []):
-                if channel.get("is_member", False):
-                    channels.append(
-                        {
-                            "id": channel.get("id", ""),
-                            "name": channel.get("name", ""),
-                            "purpose": channel.get("purpose", {}).get("value", ""),
-                            "topic": channel.get("topic", {}).get("value", ""),
-                            "member_count": channel.get("num_members", 0),
-                            "is_private": channel.get("is_private", False),
-                        }
-                    )
+                channels.append(
+                    {
+                        "id": channel.get("id", ""),
+                        "name": channel.get("name", ""),
+                        "purpose": channel.get("purpose", {}).get("value", ""),
+                        "topic": channel.get("topic", {}).get("value", ""),
+                        "member_count": channel.get("num_members", 0),
+                        "is_private": channel.get("is_private", False),
+                    }
+                )
 
             cursor = response.get("response_metadata", {}).get("next_cursor")
             if not cursor or len(channels) >= limit:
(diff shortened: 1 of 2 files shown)


gakonst:
good catch ty

PR #308, opened 2026-05-30
Add create-linear-project skill + create_project tool method

shwniscool:
## What

Adds a gated intake **skill** for creating Linear projects under the **Darkbloom** team, backed by a new `create_project` method on the Linear tool.

### Why a skill (not a workflow)
Project creation is interactive + on-demand: present a template, take freeform replies across turns, slot them in, prompt for gaps, then create. Skills drive human-in-the-loop conversation; workflows are for unattended/recurring/durable jobs.

## Changes
- **`tools/productivity/linear/client.py`** — new `create_project(name, team_ids, content, description, lead_id, priority, start_date, target_date)` → Linear `projectCreate` mutation.
- **`tools/productivity/linear/cli.py`** — `linear create-project` command (resolves team keys + lead name, maps priority label → int).
- **`.agents/skills/create-linear-project/SKILL.md`** — the skill.

## Intake (all 9 required before creation)
1. What is this? · 2. Hypothesis · 3. Metrics to improve · 4. Evidence · 5. Impact timing · 6. Priority · 7. Lead · 8. Start date · 9. Target date

Mapping: fields 1–5 → project overview document (`content`); 6–9 → native fields (`priority`, `lead`, `startDate`, `targetDate`); always `teamIds: [DAR]`. Admin-only. Creates directly — no preview step.

## Notes
- Enforcement lives in the skill (guidance), not the tool. If hard server-side gating is wanted later, add required-field validation in `create_project`.
- Not yet exercised end-to-end (no test project created in Darkbloom); `create_project` becomes visible in `call linear` after merge + redeploy.

### DIFF
--- a/.agents/skills/create-linear-project/SKILL.md
+++ b/.agents/skills/create-linear-project/SKILL.md
@@ -1,0 +1,112 @@
+---
+name: create-linear-project
+description: "Create a Linear project under the Darkbloom team through a required intake flow. Use when an admin asks to create/start/spin up a new Linear project, a new bet, or a new initiative to go after. Enforces a structured questionnaire (hypothesis, metrics, evidence, impact timing, priority, lead, dates) before anything is created."
+---
+
+# Create Linear Project (Darkbloom)
+
+We don't just create projects — we run a flow. Every new Darkbloom project must answer a
+fixed set of questions before it exists. This skill enforces that intake, then creates the
+project directly in Linear with no preview step.
+
+## Who can use this
+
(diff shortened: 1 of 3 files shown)


gakonst:
down to extend the linear tool - do you mind removing the darkbloom specifics?

goksu (changes requested):
We can't land this as it has too many Darkbloom specific context.

What you are attempting to achieve should be doable at the overlay layer for your organization.

Creating project through Linear client does make sense. Happy to review if you want to split that one and open a different diff.

PR #315, opened 2026-05-31
feat(api-rs): add sandbox core

Zygimantass:
## Summary

This is the bottom PR in the `api-rs` control-plane stack. It creates the Rust workspace and defines the sandbox core abstraction, without committing to any concrete sandbox implementation.

## What changed

- Adds `services/api-rs` as a Rust workspace with `centaur-sandbox-core`.
- Defines the shared sandbox backend contract: create, inspect, attach/exec, suspend/resume, and stop semantics.
- Adds core lifecycle/spec types for sandbox identity, desired state, observed status, resource shape, harness selection, and I/O streams.
- Adds typed errors and byte-stream helpers used by later backend implementations.
- Adds `rfcs/0001-sandbox-abstraction.md` to capture the abstraction boundary and the expected runtime lifecycle.

## Review notes

The goal here is to keep the bottom of the stack policy-light. This PR should answer whether the trait and data model are the right stable interface for local, Kubernetes, and future sandbox backends.

## Out of scope

- No local process backend.
- No Kubernetes AgentSandbox backend.
- No session store, HTTP API, or CLI.
- No integration with the existing Python API yet.

## Testing


services/api-rs/crates/centaur-sandbox-core/src/backend.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    async fn write_bytes(&self, id: &SandboxId, bytes: Bytes) -> SandboxResult<WriteAck>;
+
+    async fn close_stdin(&self, id: &SandboxId) -> SandboxResult<()>;
+
+    async fn status(&self, id: &SandboxId) -> SandboxResult<SandboxStatus>;
+
+    async fn observe(&self, id: &SandboxId) -> SandboxResult<ObservedSandbox>;
+
+    async fn list_observed(&self) -> SandboxResult<Vec<ObservedSandbox>>;
```
gakonst, line 42: what is observe

services/api-rs/crates/centaur-sandbox-core/src/backend.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    async fn stop(&self, id: &SandboxId) -> SandboxResult<()>;
+
+    async fn pause(&self, id: &SandboxId) -> SandboxResult<()>;
+
+    async fn resume(&self, id: &SandboxId) -> SandboxResult<()>;
+
+    async fn exec(&self, id: &SandboxId, command: ExecCommand) -> SandboxResult<ExecResult>;
+
+    async fn interrupt(&self, id: &SandboxId) -> SandboxResult<()>;
```
gakonst, line 61: what is this

services/api-rs/crates/centaur-sandbox-core/src/lifecycle.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ObservedSandbox {
+    pub id: SandboxId,
+    pub backend: String,
+    pub status: SandboxStatus,
+    pub generation: Option<String>,
+    pub reason: Option<String>,
+}
```
gakonst, line 107: ?

services/api-rs/crates/centaur-sandbox-core/src/backend.rs
Zygimantass: `observe` is meant to be the richer runtime inspection call, distinct from `status`.

`status(id)` gives the cheap portable enum (`Running`, `Suspended`, `Gone`, etc). `observe(id)` gives the reconciliation snapshot: id, backend, status, plus backend-owned metadata like an opaque generation/resource-version and a diagnostic reason. The manager should use that after lifecycle operations instead of trusting in-memory state.

So the intended split is:
- `status`: quick state check
- `observe`: full observed state for one sandbox
- `list_observed`: full observed state for all sandboxes this backend owns

This probably wants doc comments, and I would also be fine renaming it to `inspect`/`describe` if that reads better.

services/api-rs/crates/centaur-sandbox-core/src/backend.rs
Zygimantass: This is intended as “interrupt the currently running foreground work, but keep the sandbox alive”. Concretely, for a process/pod backend it maps to Ctrl-C/SIGINT-ish behavior; for a harness-backed sandbox it maps to canceling the active turn/command without deleting the runtime.

That said, this is the most debatable method in the trait. It is included here to preserve the behavior the Python sandbox backend exposes today, but the RFC also calls it out as an open question because it may be too harness/policy-specific for the core sandbox layer. A backend that cannot implement it should return `Unsupported`.

services/api-rs/crates/centaur-sandbox-core/src/lifecycle.rs
Zygimantass: This is the desired-vs-observed split for reconciliation.

`ObservedSandbox` is what the runtime backend says exists right now. `DesiredSandboxState` is what the control plane wants to exist. A manager/reconciler compares the two and decides whether to create, resume, pause, stop, or report drift.

The extra fields are intentionally backend-opaque:
- `generation`: a version/fencing token, e.g. Kubernetes `resourceVersion`, observed generation, UID, or a local backend counter.
- `reason`: a human/debug reason for states like `Unknown`, failed readiness, stopped, etc.

This should get doc comments; `generation` could maybe be named `version` or `observed_version` if that is clearer.

PR #316, opened 2026-05-31
feat(api-rs): add session store

Zygimantass:
## Summary

This PR adds the durable session data model and SQLx-backed store. It is the first control-plane PR above the sandbox runtime stack.

## What changed

- Adds `centaur-session-core` with session, message, execution, event, harness, and thread-key domain types.
- Makes `thread_key` the public primary session identifier, matching the RFC direction to avoid exposing a separate public `session_id`.
- Adds validation for thread keys and harness names.
- Adds `centaur-session-sqlx`, a Postgres-backed implementation of the session store.
- Adds the initial migration for sessions, messages, executions, and durable session events.
- Adds append/replay semantics for event storage, including monotonically increasing event ids for reconnectable consumers.
- Adds `rfcs/0002-session-control-plane.md` describing the simplified session API shape and SSE replay contract.

## Review notes

This is the right PR to review persistence boundaries: schema shape, idempotent session creation, event ordering, and whether the store stays format-oblivious enough to act as a pipe between producers and consumers.

## Out of scope

- No HTTP endpoints yet.
- No CLI yet.
- No concrete sandbox execution wiring through the API server yet.
- No Codex App Server JSONL parsing. Event payloads are stored as opaque JSON values.

## Testing

### DIFF
--- a/services/api-rs/crates/centaur-session-core/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-core/src/lib.rs
@@ -11,7 +11,7 @@ use thiserror::Error;
 use time::OffsetDateTime;
 
 pub const MAX_THREAD_KEY_BYTES: usize = 512;
-pub const MAX_HARNESS_TYPE_BYTES: usize = 64;
+pub const SUPPORTED_HARNESS_TYPES: &[&str] = &["codex", "amp", "claudecode"];
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub struct ThreadKey(String);
@@ -116,21 +116,34 @@ fn validate_thread_key(value: &str) -> Result<(), ThreadKeyError> {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
-pub struct HarnessType(String);
+pub enum HarnessType {
+    Codex,
+    Amp,
+    ClaudeCode,
+}
 
 impl HarnessType {
     pub fn parse(value: impl Into<String>) -> Result<Self, HarnessTypeError> {
         let value = value.into();
-        validate_harness_type(&value)?;
-        Ok(Self(value))
+        match value.as_str() {
+            "" => Err(HarnessTypeError::Empty),
+            "codex" => Ok(Self::Codex),
+            "amp" => Ok(Self::Amp),
+            "claudecode" => Ok(Self::ClaudeCode),
+            _ => Err(HarnessTypeError::Unsupported),
+        }
     }
 
     pub fn as_str(&self) -> &str {
-        &self.0
+        match self {
+            Self::Codex => "codex",
+            Self::Amp => "amp",
+            Self::ClaudeCode => "claudecode",
+        }
     }
 
     pub fn into_string(self) -> String {
-        self.0
+        self.as_str().to_owned()
     }
 }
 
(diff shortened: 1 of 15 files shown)


services/api-rs/rfcs/0002-session-control-plane.md
```diff
@@ -0,0 +1,405 @@
...
+```text
+POST /api/session/{thread_key}
+```
+
+Creates a session for `thread_key` or returns the existing session. The
+`thread_key` path segment must be URL-encoded by clients when it contains
+reserved characters.
+
+Example request:
+
```
gakonst, line 153: explicit endpoint for spawning sandbox seems fine

services/api-rs/crates/centaur-session-sqlx/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub fn new(pool: PgPool) -> Self {
+        Self { pool }
+    }
+
+    pub async fn connect(database_url: &str) -> Result<Self, SessionStoreError> {
+        let pool = PgPoolOptions::new()
+            .max_connections(10)
+            .connect(database_url)
+            .await?;
+        Ok(Self::new(pool))
```
gakonst, line 32: maybe we should have this limit be as high as possible and not worry about it

services/api-rs/crates/centaur-session-sqlx/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+        .bind(thread_key.as_str())
+        .bind(ExecutionStatus::Queued.as_str())
+        .bind(metadata)
+        .fetch_one(&self.pool)
+        .await?;
+
+        row.try_into()
+    }
+
+    pub async fn claim_next_execution(
```
gakonst: what is this for

services/api-rs/crates/centaur-session-core/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        return Err(ThreadKeyError::MissingNamespace);
+    }
+    Ok(())
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct HarnessType(String);
+
+impl HarnessType {
+    pub fn parse(value: impl Into<String>) -> Result<Self, HarnessTypeError> {
```
gakonst: Just make harness type an enum of the supported harnesses

services/api-rs/crates/centaur-session-core/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum SessionStatus {
+    Active,
+    Idle,
+    Executing,
+    Failed,
+    Archived,
+}
+
+impl SessionStatus {
```
gakonst: derive_more(Display) or something here to reduce manualc ode

services/api-rs/crates/centaur-session-sqlx/src/lib.rs
Zygimantass: Done in cd6ab9d3: bumped the default SQLx pool cap from 10 to 50 for now. I still think we should make this configurable before production, but 50 is fine as the current local/default value.

services/api-rs/crates/centaur-session-core/src/lib.rs
Zygimantass: Done in f81a6df3. I used strum rather than SQLx derives so centaur-session-core stays DB-agnostic: the enums now derive Display / EnumString / AsRefStr, and the SQLx adapter binds via AsRef<str>. ThreadKey stays manual because it is a validated newtype.

PR #317, opened 2026-05-31
feat(api-rs): add session HTTP API

Zygimantass:
## Summary

This PR adds the Rust session HTTP API on top of the session store. The API remains intentionally simple and treats event payloads as opaque JSON.

## What changed

- Adds `centaur-api-server` with an Axum-based HTTP server.
- Adds idempotent session creation at `POST /api/session/:thread_key`.
- Adds message append support at `POST /api/session/:thread_key/messages`.
- Adds execution creation at `POST /api/session/:thread_key/execute`.
- Adds reconnectable SSE streaming at `GET /api/session/:thread_key/events` with `after_event_id` replay.
- Emits event ids in the stream so clients can resume from the last observed event.
- Wires local and AgentSandbox backend selection behind server configuration.

## Review notes

The important API contract is that public clients address sessions by `thread_key`, not an internal session id. The API should stay oblivious to producer-specific event formats; it stores and streams opaque JSON lines/events through the durable store.

## Out of scope

- No user-facing CLI; that is the next PR.
- No production deployment wiring.
- No Python API replacement or integration yet.
- No Codex App Server schema coupling.

## Testing

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/lib.rs
+++ b/services/api-rs/crates/centaur-api-server/src/lib.rs
@@ -1,4 +1,9 @@
-use std::{collections::VecDeque, convert::Infallible, sync::Arc, time::Duration};
+use std::{
+    collections::{HashSet, VecDeque},
+    convert::Infallible,
+    sync::Arc,
+    time::Duration,
+};
 
 use axum::{
     Json, Router,
@@ -23,16 +28,19 @@ use futures_util::stream;
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 use thiserror::Error;
-use tokio::time::{Instant, sleep};
+use tokio::{sync::Mutex, time::sleep};
+use tracing::warn;
 
 const SESSION_OUTPUT_LINE_EVENT: &str = "session.output.line";
 const DEFAULT_IDLE_TIMEOUT_MS: u64 = 1_000;
 const DEFAULT_MAX_DURATION_MS: u64 = 60_000;
+type SandboxSpecFactory = Arc<dyn Fn(&ThreadKey, &str) -> SandboxSpec + Send + Sync>;
 
 #[derive(Clone)]
 pub struct AppState {
     store: PgSessionStore,
     sandbox_runtime: SandboxRuntime,
+    stdout_pumps: Arc<Mutex<HashSet<String>>>,
 }
 
 pub fn build_router(store: PgSessionStore) -> Router {
@@ -43,6 +51,7 @@ pub fn build_router_with_runtime(store: PgSessionStore, sandbox_runtime: Sandbox
     let state = AppState {
         store,
         sandbox_runtime,
+        stdout_pumps: Arc::new(Mutex::new(HashSet::new())),
     };
     Router::new()
         .route("/healthz", get(healthz))
@@ -58,13 +67,24 @@ pub enum SandboxRuntime {
     Mock,
     Backend {
         backend: Arc<dyn SandboxBackend>,
-        spec: SandboxSpec,
+        spec_factory: SandboxSpecFactory,
     },
 }
 
 impl SandboxRuntime {
     pub fn backend(backend: Arc<dyn SandboxBackend>, spec: SandboxSpec) -> Self {
-        Self::Backend { backend, spec }
+        let spec_factory = move |_thread_key: &ThreadKey, _execution_id: &str| spec.clone();
+        Self::backend_with_spec_factory(backend, spec_factory)
+    }
+
+    pub fn backend_with_spec_factory<F>(backend: Arc<dyn SandboxBackend>, spec_factory: F) -> Self
+    where
+        F: Fn(&ThreadKey, &str) -> SandboxSpec + Send + Sync + 'static,
+    {
+        Self::Backend {
+            backend,
+            spec_factory: Arc::new(spec_factory),
+        }
     }
 }
 
@@ -148,17 +168,31 @@ async fn execute_session(
         )
         .await?;
 
-    let output_line_count = match run_session_pipe(
-        &state.store,
-        &state.sandbox_runtime,
-        &thread_key,
-        &execution.execution_id,
-        &sandbox_id,
-        &request.input_lines,
-        pipe_options,
-    )
-    .await
-    {
+    let run_result = match &state.sandbox_runtime {
+        SandboxRuntime::Mock => {
+            run_mock_session_pipe(
+                &state.store,
+                &thread_key,
+                &execution.execution_id,
+                &sandbox_id,
+                pipe_options,
+            )
+            .await
+        }
+        SandboxRuntime::Backend { backend, .. } => {
+            match ensure_stdout_pump(&state, &thread_key, &sandbox_id).await {
+                Ok(()) => write_input_lines(
+                    backend.as_ref(),
+                    &SandboxId::new(&sandbox_id),
+                    &request.input_lines,
+                )
+                .await
+                .map(|()| 0),
+                Err(error) => Err(error),
+            }
+        }
+    };
+    let output_line_count = match run_result {
         Ok(output_line_count) => output_line_count,
         Err(error) => {
             let error_message = error.to_string();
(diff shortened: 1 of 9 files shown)


services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+#[tokio::main]
+async fn main() -> Result<(), ServerError> {
+    init_tracing();
+
+    let database_url = env::var("DATABASE_URL").map_err(|_| ServerError::MissingDatabaseUrl)?;
+    let bind_addr = env::var("BIND_ADDR").unwrap_or_else(|_| "127.0.0.1:8080".to_owned());
+    let bind_addr: SocketAddr = bind_addr.parse()?;
+    let run_migrations = env::var("RUN_MIGRATIONS")
+        .map(|value| matches!(value.as_str(), "1" | "true" | "TRUE" | "yes"))
+        .unwrap_or(false);
```
gakonst: make the CLI use clap instead

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+        "mock" => Ok(SandboxRuntime::Mock),
+        "local" => Ok(SandboxRuntime::backend(
+            Arc::new(LocalSandboxBackend::new()),
+            local_mock_app_server_spec(),
+        )),
+        "agent-k8s" => {
+            let namespace = env::var("SESSION_SANDBOX_K8S_NAMESPACE")
+                .unwrap_or_else(|_| "centaur-sandbox-e2e".to_owned());
+            let image =
+                env::var("SESSION_SANDBOX_IMAGE").unwrap_or_else(|_| "busybox:1.36".to_owned());
```
gakonst: also use clap here

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+        .args([mock_app_server_script()])
+}
+
+fn agent_k8s_mock_app_server_spec(image: &str) -> SandboxSpec {
+    SandboxSpec::new(image)
+        .command(["/bin/sh", "-lc"])
+        .args([mock_app_server_script()])
+}
+
+fn mock_app_server_script() -> &'static str {
```
gakonst, line 226: this should just be normal rust mock server

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+
+    let execution = state
+        .store
+        .complete_execution(&execution.execution_id)
+        .await?;
+    Ok(Json(ExecuteSessionResponse {
+        ok: true,
+        execution_id: execution.execution_id,
+        thread_key: execution.thread_key,
+        status: execution.status.as_str().to_owned(),
```
gakonst: shoudl all of these be atomic? so maybe the API can be somehow dumber and instead makes a single call vs this i think risks doing a action, committing it, and then being in partial state or something

vs it could be open tx, do a bunch of things, then commit at the end

this could just live in the postgres sqlx wiring and then the api is just dumb calling it vs orchsetrating hte commits

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+    }))
+}
+
+async fn ensure_session_sandbox(
```
gakonst, line 247: ensure-style functions are slop - shouldn't this just be handled by the other functions somehow inline? idk

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+async fn run_session_pipe(
+    store: &PgSessionStore,
+    runtime: &SandboxRuntime,
+    thread_key: &ThreadKey,
+    execution_id: &str,
+    sandbox_id: &str,
+    input_lines: &[String],
+    options: PipeOptions,
+) -> Result<usize, ApiError> {
+    match runtime {
```
gakonst: this feels smelly - especially the loop for streaming data in/out? not sure what the canonical way to implement this would be but we want the API to be forwarding data in/out - maybe this can be rolled with a Sink / Stream type and be done with normal async code

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+) -> Result<(), ApiError> {
+    store
+        .append_event(
+            thread_key,
+            Some(execution_id),
+            SESSION_OUTPUT_LINE_EVENT,
+            Value::String(line.to_owned()),
+        )
+        .await?;
+    Ok(())
```
gakonst, line 465: these also feel like dumb wrappers that should be handled type-safely deeper in the api struct which owns the session store

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+}
+
+fn mock_app_server_output_lines(
+    thread_key: &ThreadKey,
+    execution_id: &str,
+    sandbox_id: &str,
+) -> Vec<String> {
```
gakonst, line 472: mock stuff should be in explicit mock module

gakonst: ideally the API is instantiated with a generic or a box dyn or something vs this garbage

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+    let data = if event.event_type == SESSION_OUTPUT_LINE_EVENT {
+        event.payload.as_str().unwrap_or_default().to_owned()
+    } else {
+        serde_json::to_string(&event.payload).unwrap_or_else(|_| "{}".to_owned())
+    };
+    Event::default()
+        .id(event.event_id.to_string())
+        .event(event.event_type)
+        .data(data)
+}
```
gakonst, line 594: these things and ohte rconversions should just be From trait implementations

and the let data stuff should be strictly typed w/ enums and stuff, and if you need do an Other variant

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+
+#[derive(Clone, Copy, Debug)]
+struct PipeOptions {
+    idle_timeout: Duration,
+    max_duration: Duration,
+}
+
+impl PipeOptions {
+    fn from_request(request: &ExecuteSessionRequest) -> Result<Self, ApiError> {
```
gakonst, line 618: `TryFrom`

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+}
+
+fn nonzero_duration_millis(value: u64) -> Result<Duration, ApiError> {
+    if value == 0 {
+        return Err(ApiError::BadRequest(
+            "duration values must be greater than zero".to_owned(),
+        ));
+    }
+    Ok(Duration::from_millis(value))
+}
```
gakonst, line 650: slop

services/api-rs/crates/centaur-api-server/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+
+#[derive(Debug, Deserialize)]
+struct CreateSessionRequest {
```
gakonst, line 653: all this stuff should be refactored in types , routes and stuff

services/api-rs/crates/centaur-api-server/src/main.rs
Zygimantass: Done in dfd8e6ab. The server binary now uses clap for all the manual env-backed config here: DATABASE_URL, BIND_ADDR, RUN_MIGRATIONS, sandbox backend selection, k8s namespace/context/image, and ready timeout. I kept the same env var names and exposed equivalent flags in --help.

PR #318, opened 2026-05-31
feat(api-rs): add session CLI

Zygimantass:
## Summary

This PR adds a small CLI for exercising the session API from a terminal. It can create or attach to a session, trigger execution, and stream durable SSE output.

## What changed

- Adds `centaur-session-cli`.
- Supports creating a session by generated or provided `thread_key`.
- Prints the created/used thread key so sessions can be reattached later.
- Supports attaching to an existing session without requiring a separate public session id.
- Supports `--after-event-id` so callers can replay from a known event boundary.
- Prints stream event ids alongside event data for manual reconnect testing.
- Keeps the stream open for long-lived sessions instead of assuming a fixed number of turns.

## Review notes

This is primarily a smoke-test and operator ergonomics tool for the early Rust control-plane slice. It should stay thin: the CLI should not understand producer-specific payload formats, only the API envelope and SSE event ids.

## Out of scope

- No TUI behavior.
- No auth story beyond the current local development path.
- No Codex App Server JSONL parsing or rendering.
- No automatic terminal-state interpretation beyond displaying streamed events.

## Testing

### DIFF
--- a/services/api-rs/crates/centaur-session-cli/src/main.rs
+++ b/services/api-rs/crates/centaur-session-cli/src/main.rs
@@ -1,0 +1,349 @@
+use anyhow::{Context, Result, bail};
+use centaur_session_core::ThreadKey;
+use clap::Parser;
+use futures_util::StreamExt;
+use reqwest::Client;
+use serde_json::{Value, json};
+use tokio::task::JoinHandle;
+use uuid::Uuid;
+
+const DEFAULT_MESSAGE: &str = "Reply with exactly PONG and nothing else.";
+
+#[derive(Debug, Parser)]
+#[command(about = "Create, execute, or attach to a Centaur session")]
+struct Args {
+    #[arg(long, env = "CENTAUR_API_URL", default_value = "http://127.0.0.1:8080")]
+    api_url: String,
+
+    #[arg(long)]
+    thread_key: Option<String>,
+
+    #[arg(long)]
+    attach: bool,
+
+    #[arg(long, default_value = "codex")]
+    harness_type: String,
+
+    #[arg(long)]
+    message: Option<String>,
+
+    #[arg(long = "input-line")]
+    input_lines: Vec<String>,
+
+    #[arg(long, default_value_t = 1_000)]
+    idle_timeout_ms: u64,
+
+    #[arg(long, default_value_t = 60_000)]
+    max_duration_ms: u64,
+
+    #[arg(long, default_value_t = 0)]
+    after_event_id: i64,
+
+    #[arg(long)]
+    all_events: bool,
+
+    #[arg(long)]
+    exit_on_terminal: bool,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+    let attach_mode = attach_mode(&args);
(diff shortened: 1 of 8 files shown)


services/api-rs/crates/centaur-session-cli/src/main.rs
```diff
@@ -0,0 +1,349 @@
+use anyhow::{Context, Result, bail};
```
gakonst, line 1: use `eyre` instead of anyhow

services/api-rs/crates/centaur-session-cli/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    #[arg(long)]
+    exit_on_terminal: bool,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+    let attach_mode = attach_mode(&args);
+    validate_mode(&args, attach_mode)?;
+    let (thread_key, generated_thread_key) = thread_key_arg(&args, attach_mode)?;
```
gakonst, line 71: instead of this stuff you should be able to just do enums or newtypes with clap

services/api-rs/crates/centaur-session-cli/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+
+fn thread_key_arg(args: &Args, attach_mode: bool) -> Result<(String, bool)> {
+    match (&args.thread_key, attach_mode) {
+        (Some(thread_key), _) => Ok((thread_key.clone(), false)),
+        (None, true) => bail!("--attach requires --thread-key"),
+        (None, false) => Ok((format!("cli:{}", Uuid::new_v4().simple()), true)),
+    }
+}
+
+async fn post_json(client: &Client, url: &str, payload: Value) -> Result<Value> {
```
gakonst, line 210: all the Rust API client should be in the API Server Rust code crate - and this CLI should be just calling that typed thing with a struct CentaurClient { client: ... }

services/api-rs/crates/centaur-session-cli/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+
+async fn stream_output_lines(
+    response: reqwest::Response,
+    all_events: bool,
+    exit_on_terminal: bool,
+) -> Result<()> {
+    let mut chunks = response.bytes_stream();
+    let mut buffer = String::new();
+
+    while let Some(chunk) = chunks.next().await {
```
gakonst, line 400: similarly the client i said above should just return a stream, and then the CLI is just printing that to stdout

services/api-rs/crates/centaur-session-cli/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    event: String,
+    data: String,
+}
+
+impl SseFrame {
+    fn parse(frame: &str) -> Option<Self> {
+        let frame = frame.replace("\r\n", "\n");
+        let mut id = None;
+        let mut event = "message".to_owned();
+        let mut data = Vec::new();
```
gakonst, line 601: slop - use upstream sse libraries from axum

PR #319, opened 2026-05-31
feat(api-rs): add local sandbox backend

Zygimantass:
## Summary

This PR adds the first concrete sandbox implementation on top of the core abstraction: a local process-backed backend plus a small manager layer that owns reconciliation behavior.

## What changed

- Adds `centaur-sandbox-local`, a local backend that can spawn and manage sandbox-like child processes.
- Adds `centaur-sandbox-manager`, a backend-independent manager for desired-state reconciliation.
- Covers lifecycle operations for local runtimes, including pause/resume/stop behavior and exec attachment behavior.
- Adds manager tests for drift handling, partial backend loss, and failure-preserving desired state.

## Why this layer is separate

The local backend gives us a fast implementation target for CLI and API development without requiring Kubernetes. The manager crate is also intentionally backend-neutral so it can be shared by the AgentSandbox backend in the next PR.

## Review notes

This PR is the right place to review manager semantics: how desired state is stored, how drift is reported, and how backend failures are surfaced without silently rewriting intent.

## Out of scope

- No Kubernetes AgentSandbox backend.
- No session store, HTTP API, or CLI.
- No real Codex App Server protocol assumptions; this layer only moves bytes through the sandbox abstraction.

## Testing


services/api-rs/crates/centaur-sandbox-manager/src/reconcile.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum DriftReason {
+    NoDesiredState,
+    MissingWhileRunning,
+    MissingWhileSuspended,
+    UnknownObservedState(String),
+}
+
```
gakonst, line 19: this feels a bit slop but im not sure

PR #330, opened 2026-06-01
docs: document macos local setup and Slack quickstart guidance

samlaf:
The existing local setup docs do not work smoothly on a macOS laptop, especially on Apple Silicon: the currently published Centaur images are x86-only, so the GHCR path fails on local arm64 clusters. I spent a lot of time debugging image pull failures, kind image loading, local Slack webhook access, and required Slack scopes before getting a working end-to-end smoke test.

Document the reproducible path so other developers can test Centaur quickly without rediscovering the same issues:
- add macOS/kind local image loading guidance to the Mac Mini setup doc
- explain why local Docker images must be loaded into kind/containerd
- make Slack quickstart setup more generic around the public webhook URL
- document minimal Slack app event/scopes for channel mention testing
- add local tunnel guidance for exposing Slackbot during laptop testing

Reproduced docs in the md and mdx sections, but not sure how you guys keep those in sync. Just manually like this?

### DIFF
--- a/docs/pages/mac-mini-setup.mdx
+++ b/docs/pages/mac-mini-setup.mdx
@@ -14,6 +14,38 @@ Centaur publishes development images to GHCR. On a single small host, the
 simplest setup is to point the local chart at those images instead of building
 and importing images into k3s' container runtime.
 
+If you are evaluating on macOS, especially Apple Silicon, use the local-build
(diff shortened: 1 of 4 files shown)


gakonst (approved):
thank you!

PR #334, opened 2026-06-01
refactor(api-rs): use owned sandbox io streams

Zygimantass:
## Summary

- replace the sandbox polling byte API with `open_io`, returning owned Tokio stdin/stdout/stderr handles
- frame session stdin/stdout with `tokio_util::codec::{FramedWrite, FramedRead, LinesCodec}` instead of hand-rolled byte reads and line buffering
- remove backend-held read/write stream mutexes from local and Agent Sandbox implementations
- move session pipe handling into `centaur-session-runtime`, keeping the API server as a thin HTTP adapter
- update sandbox manager/e2e invariants and RFC 0001 for the owned I/O model

## Split

- Session event notification work moved to #345.

## Validation

- `cargo fmt --all --check`
- `cargo test -p centaur-sandbox-core -p centaur-sandbox-local -p centaur-sandbox-agent-k8s -p centaur-sandbox-manager -p centaur-sandbox-e2e -p centaur-api-server -p centaur-session-cli`
- `cargo clippy -p centaur-sandbox-core -p centaur-sandbox-local -p centaur-sandbox-agent-k8s -p centaur-sandbox-manager -p centaur-sandbox-e2e -p centaur-api-server -p centaur-session-cli --all-targets -- -D warnings`


Zygimantass:
Follow-up on the API-server boundary: moved the live stream/sink coordination, sandbox pipe registry, stdout/stderr pumps, execution state transitions, and event replay loop out of `centaur-api-server` into a new `centaur-session-runtime` crate. The API crate is now just HTTP request parsing, runtime calls, error mapping, and SSE event adaptation.

Zygimantass:
Removed the last runtime type branch: `SessionRuntime` now consumes a manager-backed `SandboxRuntime` for all executions. The old `SandboxRuntime::Mock` path and direct event synthesis are gone; mock server mode now runs the same local scripted sandbox through `SandboxBackend -> SandboxManager -> open_io` as the rest of the flow.

Zygimantass:
Moved the remaining mock workload details out of `centaur-api-server`. The API server now only maps CLI config to a backend plus `SandboxWorkloadMode`; the mock app-server script/spec and codex spec generation live in `centaur-session-runtime`. Also removed the `mock` backend variant: backend is now `local` or `agent-k8s`, and `mock` is only a workload mode.

services/api-rs/crates/centaur-api-server/src/types.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+impl From<&str> for SessionEventName {
+    fn from(value: &str) -> Self {
+        Self::from(value.to_owned())
+    }
+}
+
+pub fn session_event_to_sse(event: SessionEvent) -> Result<Event, SessionEventConversionError> {
```
Zygimantass: can't this be an `impl TryFrom<SessionEvent> for Event`?

services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
```diff
@@ -232,22 +208,12 @@ impl AgentSandboxBackend {
...
-    async fn drop_streams(&self, id: &SandboxId) {
-        if let Some(streams) = self.streams.lock().await.remove(id) {
-            streams.attached.abort();
-        }
+        let stdin = stdin.ok_or_else(|| SandboxError::Io("stdin was not attached".to_owned()))?;
+        let stdout =
+            stdout.ok_or_else(|| SandboxError::Io("stdout was not attached".to_owned()))?;
+        let stderr =
+            stderr.ok_or_else(|| SandboxError::Io("stderr was not attached".to_owned()))?;
+        Ok(SandboxIo::with_guard(stdin, stdout, stderr, attached))
```
Zygimantass, line 217: what is the purpose of the guard herE?

services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
```diff
@@ -395,27 +298,6 @@ impl SandboxBackend for AgentSandboxBackend {
...
-    } else {
-        reader
-            .read(buf)
-            .await
-            .map(Some)
-            .map_err(|err| SandboxError::Io(format!("failed to read output: {err}")))
-    }
-}
-
 fn sandbox_status_from_pod(replicas: i32, pod: Option<&Pod>) -> SandboxStatus {
```
Zygimantass, line 302: is there any better way to do this? it feels like this is flaky

services/api-rs/crates/centaur-sandbox-local/src/lib.rs
```diff
@@ -266,28 +217,18 @@ fn command_parts(spec: &SandboxSpec) -> SandboxResult<(&str, Vec<&str>)> {
...
 
-async fn read_with_timeout<R>(
-    reader: &mut R,
-    buf: &mut [u8],
-    timeout_ms: Option<u64>,
-) -> SandboxResult<Option<usize>>
+fn box_read<R>(reader: R) -> Pin<Box<dyn AsyncRead + Send>>
 where
-    R: AsyncReadExt + Unpin,
+    R: AsyncRead + Send + 'static,
```
Zygimantass: why do these need to be functions? feels like we could inline

services/api-rs/crates/centaur-api-server/src/types.rs
Zygimantass: Done as `SessionSseEvent: TryFrom<SessionEvent>`. The exact `TryFrom<SessionEvent> for axum::response::sse::Event` impl is blocked by Rust orphan rules because `TryFrom`, `Event`, and `SessionEvent` are all foreign to this crate, so the local wrapper gives us the typed conversion without keeping the free function.

services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
Zygimantass: Added a note in the helper. This intentionally keys `Running` off the backing Pod `Ready=True` condition rather than phase alone, because phase can be `Running` before the container is usable for attach/I/O. A future explicit Agent Sandbox ready condition would be nicer, but the current portable boundary we can validate is the Pod Ready condition.

services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
Zygimantass: Added a code comment here. The guard keeps kube `AttachedProcess` alive for the lifetime of the returned stdin/stdout/stderr handles; dropping it tears down the Kubernetes attach connection.

services/api-rs/crates/centaur-sandbox-local/src/lib.rs
Zygimantass: Done, inlined the `Box::pin` casts and removed the helper functions.

PR #340, opened 2026-06-01
feat: add slackbot thread follow mode

### LINKED ISSUE #246: Slackbot ignores direct messages that do not mention the bot
## Summary
The Slackbot acknowledges direct-message webhook events with HTTP 200 but does not hand them off to Centaur unless the message text explicitly mentions the bot user.

In Slack IM/MPIM channels, users reasonably expect messages sent directly to the bot to be actionable without including an explicit `<@bot>` mention.

## Impact
- Users can DM the Slackbot and see no response.
- Slack receives a successful webhook response, so it does not retry.
- No `slack_thread_turn` workflow run is created for the ignored direct message.
- The system appears healthy from pod, ingress, and webhook status checks while user-facing Slack responses fail.

## Expected behavior
Slack events with `channel_type: "im"` or `channel_type: "mpim"` should be treated as actionable bot-directed messages, even when the text does not contain an explicit bot mention.

## Actual behavior
Direct-message `message` events are normalized as non-mentions unless the text contains the bot user mention, so the event processing path returns before creating a workflow run.

## Suggested fix
Update Slack event normalization so IM/MPIM message events are considered actionable in the same way as app mentions. Add regression coverage for a direct-message `message` event with `channel_type: "im"` and no textual bot mention, asserting that it creates an actionable normalized event.

ragnorc:
## Summary

This adds an opt-in Slackbot thread-follow mode behind `SLACKBOT_THREAD_FOLLOW`, defaulting off so upstream behavior remains mention-only unless a deployment explicitly enables it.

When enabled, Slackbot now dispatches three explicit trigger reasons into the existing `slack_thread_turn` workflow:

- `mention` for current app mentions and explicit bot mentions.
- `direct_message` for 1:1 Slack DM `message` events with `channel_type: "im"`.
- `active_thread_reply` for non-mention replies in channel/group threads that already have an active Centaur runtime assignment.

Unrelated channel root messages are still ignored. Multiparty DMs remain out of scope and are ignored by the new follow logic.

## Related Issues and Prior Art

Fixes #246 for 1:1 DMs. That issue describes the current behavior where Slackbot acknowledges direct-message webhook events with HTTP 200 but does not create a `slack_thread_turn` workflow unless the DM text explicitly mentions the bot.

This also revisits the active-thread continuation idea from #234 and #235, but uses a stricter gate. Those PRs allowed continuation based on bot presence in Slack thread history and #235 was closed because that could make Centaur respond to every message in a multi-person Slack thread. This PR instead requires an active Centaur runtime assignment from `/agent/runtime?key=...`, and the whole behavior is disabled unless `SLACKBOT_THREAD_FOLLOW=1` is set.

## Root Cause

Slackbot already normalized both `app_mention` and `message` events, but the webhook dispatch path dropped every normalized event unless `normalized.is_mention` was true. Slack sends 1:1 DMs as `message.im` events rather than `app_mention`, so the existing gate made DMs and active-thread follow-up replies impossible without another explicit mention.

## Fix

- Add strict parsing for `SLACKBOT_THREAD_FOLLOW`.
- Preserve `is_mention` as a real-mention signal and add `trigger_reason` metadata instead of overloading mention semantics.
- Expose Slack `channel_type` on normalized events.
- Backfill Slack thread history for DM thread replies, while avoiding extra history fetches for non-mention channel replies.
- Add a small Slackbot runtime lookup client for `/agent/runtime?key=...` and fail closed when active-thread state cannot be proven.
- Include `trigger_reason` in Slackbot workflow metadata and tracing.
- Document the flag and Slack Events API scope/subscription requirements.

## Validation

- `cd services/slackbot && bun test src/*.test.ts src/**/*.test.ts`
- `cd services/slackbot && bun tsgo --project tsconfig.json --noEmit`
- `cd services/slackbot && bun test test/emulate/slack-e2e.test.ts`
- `just build-one slackbot`
- `just deploy`
- `kubectl rollout status deployment/centaur-centaur-slackbot -n centaur --timeout=120s`
- `kubectl exec -n centaur deploy/centaur-centaur-api -- curl -sS http://centaur-centaur-slackbot:3001/health`

The local deployed Slackbot health check returned:

```json
{"ok":true,"service":"slackbot","commit":"local"}
```

## Notes

The Slackbot and API deployments rolled out locally. The unrelated local `centaur-centaur-token-broker` pod was in `CreateContainerConfigError`; this PR does not touch token broker.


Zygimantass:
closed, since we refactored the slackbot - feel free to reopen the PR against services/slackbotv2 and ping me to take a look:)

PR #342, opened 2026-06-01
fix(slackbot): avoid corrupting final delivery when the streamed-char offset lands mid-token

axis-guzus:
## Problem

Agent answers delivered to Slack are intermittently corrupted — characters are dropped or words merge (observed in production: "market cap" → "marketountries", "defillama.com" → "vama.com", "authenticate" → "thenticate"). The content is correct server-side; only the Slack render is mangled.

## Root cause

`continuationText` (`services/slackbot/src/centaur/final-delivery.ts`) blind-slices the canonical `result_text` at `slackbot_streamed_answer_chars`:

```ts
return text.slice(offset).trimStart();
```

That offset counts the **source characters streamed live** (the per-delta source-char accumulator in `agent-session.ts` / `codex-session.ts`), which is a *different string-space* than `result_text`:

- markdown normalization adds boundary characters to the live stream that aren't in the source count;
- whitespace / paragraph reflow changes lengths;
- the harness can revise the final answer — the code even logs `slack_codex_canonical_answer_correction` when the canonical answer length differs from what was streamed.

When the offset is misaligned, `text.slice(offset)` lands mid-token and drops characters. The offset is accumulated server-side with `max(...)`, so overshoot is sticky.

## Fix

`continuationText` only has the offset count, not the streamed string, so it can't verify the prefix. This guards the slice: only treat the offset as a split point when it lands on a **whitespace boundary** in `result_text`; when it's misaligned, fall back to posting the full answer (re-showing the already-streamed prefix is strictly preferable to emitting corrupted text). The aligned suffix-continuation path (live delivery cut off at a clean boundary) is unchanged.

## Tests

Adds a regression test (offset lands mid-token → the full answer is posted, not a corrupted slice). The existing "posts only the missing suffix when live delivery was cut off" test still passes.

```
bun test src/centaur/final-delivery.test.ts  →  6 pass, 0 fail
```

## Note

A more complete fix would plumb the streamed prefix text through (or compute the continuation server-side, where both strings are available) so the split point can be verified exactly rather than heuristically. Happy to follow up if you'd prefer that direction — this change is a minimal, safe guard against the user-visible corruption.

### DIFF
--- a/services/slackbot/src/centaur/final-delivery.test.ts
+++ b/services/slackbot/src/centaur/final-delivery.test.ts
@@ -454,6 +454,75 @@ describe("final delivery polling", () => {
       globalThis.fetch = originalFetch;
     }
   });
+
+  it("posts the full answer instead of corrupting it when the streamed-char offset lands mid-token", async () => {
+    // `slackbot_streamed_answer_chars` counts streamed *source* characters, which
+    // can drift from `result_text` (normalization / reflow / answer revisions). An
+    // offset that lands inside a word must not be blindly sliced — that drops
+    // characters (e.g. "market cap" -> "marketountries"). The full, uncorrupted
+    // answer should be posted instead.
+    const originalFetch = globalThis.fetch;
+    const resultText = "USDT ~$190B and USDC ~$78B are ~90% of the stablecoin market.";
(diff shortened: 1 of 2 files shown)


Zygimantass:

PR #345, opened 2026-06-01
refactor(api-rs): wake session streams with postgres notify

Zygimantass:
## Summary

- add a `session_events` insert trigger that publishes thread/event IDs through Postgres `LISTEN/NOTIFY`
- add a SQLx session-event listener and typed notification payload parsing
- wake durable session event streams from notifications, with the existing DB replay path retained as the source of truth and a slower safety tick as fallback

## Stack

- Stacked on #334. This PR contains only the session event notification delta.

## Validation

- `cargo fmt --all --check`
- `cargo test -p centaur-session-sqlx -p centaur-session-runtime -p centaur-api-server -p centaur-session-cli`
- `cargo clippy -p centaur-session-sqlx -p centaur-session-runtime -p centaur-api-server -p centaur-session-cli --all-targets -- -D warnings`

### DIFF
--- a/services/api-rs/crates/centaur-session-runtime/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-runtime/src/lib.rs
@@ -12,17 +12,24 @@ use centaur_sandbox_manager::SandboxManager;
 use centaur_session_core::{
     HarnessType, Session, SessionEvent, SessionExecution, SessionMessageInput, ThreadKey,
 };
-use centaur_session_sqlx::{PgSessionStore, SessionStoreError, default_metadata};
+use centaur_session_sqlx::{
+    PgSessionStore, SessionEventListener, SessionStoreError, default_metadata,
+};
 use futures_util::{SinkExt, Stream, StreamExt, stream};
 use serde_json::{Value, json};
 use thiserror::Error;
-use tokio::{io, sync::Mutex, time::sleep};
+use tokio::{
+    io,
+    sync::Mutex,
+    time::{Instant, Interval, MissedTickBehavior, interval_at},
+};
 use tokio_util::codec::{FramedRead, FramedWrite, LinesCodec, LinesCodecError};
 use tracing::warn;
 
 pub const SESSION_OUTPUT_LINE_EVENT: &str = "session.output.line";
 
 const MAX_SESSION_OUTPUT_LINE_BYTES: usize = 1024 * 1024;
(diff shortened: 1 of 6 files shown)


services/api-rs/crates/centaur-session-runtime/src/lib.rs
```diff
@@ -424,6 +445,39 @@ fn session_event_stream(
     )
 }
 
+fn event_stream_safety_tick() -> Interval {
```
Zygimantass: inline this

services/api-rs/crates/centaur-session-runtime/src/lib.rs
```diff
@@ -424,6 +445,39 @@ fn session_event_stream(
...
+    let mut tick = interval_at(
+        Instant::now() + EVENT_STREAM_SAFETY_POLL_INTERVAL,
+        EVENT_STREAM_SAFETY_POLL_INTERVAL,
+    );
+    tick.set_missed_tick_behavior(MissedTickBehavior::Delay);
+    tick
+}
+
+async fn wait_for_session_event_notification(
```
Zygimantass: inline this

services/api-rs/crates/centaur-session-runtime/src/lib.rs
```diff
@@ -424,6 +445,39 @@ fn session_event_stream(
...
+                if notification_matches(&notification, &state.thread_key, state.after_event_id) {
+                    return Ok(());
+                }
+            }
+            _ = state.safety_tick.tick() => return Ok(()),
+        }
+    }
+}
+
+fn notification_matches(
```
Zygimantass: inline this

services/api-rs/crates/centaur-session-runtime/src/lib.rs
Zygimantass: Done, inlined the notification wait loop into the empty-events branch.

services/api-rs/crates/centaur-session-runtime/src/lib.rs
Zygimantass: Done, inlined the safety tick construction into the stream state initialization.

services/api-rs/crates/centaur-session-runtime/src/lib.rs
Zygimantass: Done, inlined the thread/event-id predicate into the notification match guard.

PR #344, opened 2026-06-01
feat: python no more

Zygimantass:

PR #329, opened 2026-06-01
feat: add canonical renderer pipeline

goksu:
## Target

This targets the renderer split described in the [PR #328 architecture comment](https://github.com/paradigmxyz/centaur/pull/328#issuecomment-4589040797):

- Rust/session API remains below rendering and emits [`session.*` events](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/services/api-rs/crates/centaur-session-core/src/lib.rs#L346) plus opaque output lines.
- [`@centaur/rendering`](https://github.com/paradigmxyz/centaur/tree/ca8e3a5d360e180ea0944332e82763e1605cc440/packages/rendering) owns source mapping from Rust/Codex notification shapes to [`renderer.*` events](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/packages/rendering/src/types.ts#L25).
- [`ChatSDKRenderer`](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/packages/rendering/src/chat-sdk.ts#L42) converts `renderer.*` events into Chat SDK-shaped messages and stream chunks.
- Slackbot owns Slack target validation and Slack side effects in [`SlackChatSDKRenderer` / `SlackStreamRenderer`](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/services/slackbot/src/rendering/slack.ts#L61) above the Chat SDK layer.

## Renderer layers

This PR keeps the rendering path split into caller-neutral layers:

1. **Session event layer**: Rust and harness packages emit transport/session events without Slack-specific concepts. Code: [`SessionEvent`](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/services/api-rs/crates/centaur-session-core/src/lib.rs#L346), [`RustSessionStreamEvent`](https://github.com/paradigmxyz/centaur/blob/ca8e3a5d360e180ea0944332e82763e1605cc440/packages/harness-events/src/types.ts#L71).
[...]

PR #322, opened 2026-05-31
[codex] add Chat SDK slackbot v2

gakonst:
## Summary

- adds `services/slackbotv2`, a from-scratch Chat SDK Slackbot wired to the Rust session API from #321
- idempotently creates/gets the Rust session with `POST /api/session/{thread_key}` before writing `/messages`, avoiding FK failures in `session_messages`
- forwards Slack thread history and subscribed messages to `POST /api/session/{thread_key}/messages`, and calls `POST /api/session/{thread_key}/execute` only when the bot is mentioned
- consumes `GET /api/session/{thread_key}/events` SSE, converts `session.output.line` Codex app-server events into Chat SDK chunks, and preserves Slack assistant status/title updates
- adds generated Codex app-server TypeScript event types plus a Chat SDK stream mapper for plan/task/command/intermediate rendering
- adds a runnable Bun entrypoint for `slackbotv2`, with `dev` and `start` scripts, so the service can be booted locally instead of only imported in tests

## Running locally

```bash
PORT=3012 \
SLACK_BOT_TOKEN=xoxb-... \
SLACK_SIGNING_SECRET=... \
DATABASE_URL=postgresql://... \
CENTAUR_API_URL=http://127.0.0.1:8080 \
pnpm --filter slackbotv2 start
```

The service exposes:

```text
GET  /health
POST /api/webhooks/slack
```

Optional env:

```text
SLACKBOT_API_KEY or CENTAUR_API_KEY
SLACK_API_URL
SLACK_BOT_USER_ID
SLACKBOTV2_ASSISTANT_STATUS
SLACKBOTV2_DATABASE_URL
SLACKBOTV2_STATE_KEY_PREFIX
SLACKBOTV2_USER_NAME
SESSION_IDLE_TIMEOUT_MS
SESSION_MAX_DURATION_MS
```

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm --filter @centaur/harness-events typecheck`
- `pnpm --filter @centaur/harness-events test`
- `pnpm --filter slackbotv2 check:types`
- `pnpm --filter slackbotv2 test`
- local boot smoke: started `slackbotv2` on `PORT=3012` with local Postgres + Rust API, then verified `curl http://127.0.0.1:3012/health` returned `{"ok":true,"service":"slackbotv2"}`
[...]

PR #346, opened 2026-06-01
[codex] stack Chat SDK Slackbot on api-rs control plane

gakonst:
## Summary
- stack the Chat SDK Slackbot v2 work from #322 and #327 on top of #344
- add the canonical `@centaur/rendering` pipeline from #329 and make Slackbotv2 consume that renderer directly
- remove the duplicate Chat SDK stream mapper from `@centaur/harness-events`; that package now stays focused on shared event shapes
- keep V1 Slackbot untouched in the PR delta

## Validation
- `pnpm --filter @centaur/harness-events test`
- `pnpm --filter @centaur/harness-events typecheck`
- `pnpm --filter @centaur/rendering test`
- `pnpm --filter @centaur/rendering typecheck`
- `pnpm --filter slackbotv2 test`

## Local deploy notes
- built and rolled out `centaur-slackbotv2-pr344-chat-sdk:18f27b5a` to `centaur-slackbotv2-pr327`
- API-rs remains on `centaur-api-rs-pr344-chat-sdk:9939d6ab-linux`; local API health is passing
- local API-rs is running with `RUN_MIGRATIONS=false` against the reused PR327 DB because that DB has the older migration checksum
- Funnel is on: `https://gak.tail388b2e.ts.net` -> `127.0.0.1:3002`


services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
...
+    traceLog(input.options, 'slackbotv2_forward_context_skipped', trace, {
+      message_count: 1
+    })
+  }
+
+  let lastEventId = state.lastEventId ?? 0
+
+  const forwardInput: ForwardSessionInput = {
+    afterEventId: lastEventId,
+    executeMessage: input.execute ? serializedMessage : undefined,
```
Zygimantass: This still queues a new execution while the thread is already streaming if the follow-up is also a mention. `shouldOpenStream` is false when `state.activeExecution` is true, but `executeMessage` stays set from `input.execute`, so `forwardToSessionApi()` calls `/execute` and then returns without opening a stream. That leaves the second execution unrendered, or races API-rs execution serialization. We should gate this on `shouldOpenStream` (or explicitly queue the mention as context only) and add a test for `app_mention` during an active stream; the current test only covers a non-mention subscribed message.

services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
...
+  }
+}
+
+function attachmentDescription(attachment: SlackbotV2ApiAttachment): string {
+  const fields = [
+    `name=${attachment.name ?? 'attachment'}`,
+    `type=${attachment.type}`,
+    attachment.mimeType ? `mime=${attachment.mimeType}` : undefined,
+    attachment.url ? `url=${attachment.url}` : undefined,
+    attachment.dataBase64 ? `base64=${attachment.dataBase64}` : undefined,
```
Zygimantass: For non-image attachments this puts the full base64 payload into a text block. A modest PDF/doc can become megabytes of prompt text, and these are private Slack file bytes that the model will see as opaque text instead of a usable attachment. Can we avoid embedding raw base64 here and either pass an attachment ref/file through the API-rs message path, or include only metadata/url until document handling is implemented?

services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
...
+    ...fields
+  })
+}
+
+function errorMessage(error: unknown): string {
+  if (error instanceof Error) return error.message
+  return String(error)
+}
+
+type RawSlackEvent = Record<string, unknown> & {
```
Zygimantass: what is this record<string, unknown> stuff? isn't there a better safer typescript stuff?

services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
...
+    adapters: { slack },
+    state: options.state ?? createDefaultState(options, logger),
+    onLockConflict: 'force',
+    logger
+  })
+
+  chat.onNewMention(async (thread, message) => {
+    if (!isAllowedSlackMessage(message, options, logger)) return
+    await thread.subscribe()
+    await forwardAndMaybeRender(thread, message, {
```
Zygimantass: what is forwardAndMaybeRender? explain it and make the naming better. also execute and includeContext is always the same (either both true or both false, let's just collapse them into one param)?

services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
...
+
+function isRawSlackEvent(value: unknown): value is RawSlackEvent {
+  return Boolean(value && typeof value === 'object' && !Array.isArray(value))
+}
+
+function isRawSlackEnvelope(value: unknown): value is RawSlackEnvelope {
+  return Boolean(value && typeof value === 'object' && !Array.isArray(value))
+}
+
+async function forwardAndMaybeRender(
```
Zygimantass: i think there's a bug here, find it

services/slackbotv2/src/index.ts
```diff
@@ -0,0 +1,1136 @@
+import { Hono } from 'hono'
```
Zygimantass, line 1: split it up into a few logical files please

services/slackbotv2/src/index.ts
Zygimantass: we'll fix this later, all good

Zygimantass: agent: add a todo item here, attachments should call the api - POST /session/{thread_key}/attachments

services/slackbotv2/src/index.ts
Zygimantass: Found and fixed: the active-stream mention path could still pass an `executeMessage` while `openStream` was false. That is now derived from `shouldStartExecution`, and the new test covers a second `app_mention` arriving while the first stream is open.

services/slackbotv2/src/index.ts
Zygimantass: Split the file up. `index.ts` now handles app wiring/session orchestration, `slack-events.ts` owns Slack filtering and raw event narrowing, `session-api.ts` owns serialization/API/SSE handling, `types.ts` owns shared types, and `utils.ts` owns small shared helpers.

services/slackbotv2/src/index.ts
Zygimantass: Renamed this path to `syncThreadMessageToSession`, added a doc comment explaining that it persists Slack updates and conditionally renders an execution stream, and collapsed the duplicated `execute`/`includeContext` flags into a single `mode: "append" | "execute"` input.

services/slackbotv2/src/index.ts
Zygimantass: Added the TODO next to the current attachment fallback in `session-api.ts`: future file handling should upload through `POST /session/{thread_key}/attachments` and pass refs instead of inlining bytes.

services/slackbotv2/src/index.ts
Zygimantass: Moved the raw Slack payload handling into `slack-events.ts` and replaced the `Record<string, unknown>` raw event/envelope types with explicit `RawSlackEnvelope`/`RawSlackEvent` shapes over `JsonValue`. The app code now narrows through those guards instead of carrying arbitrary records around.

services/slackbotv2/src/index.ts
Zygimantass: Fixed in `04fb4851`: `executeMessage` is now gated by `shouldStartExecution`, so a mention that arrives during an active stream only appends context and does not call `/execute`. Added a regression test for a second `app_mention` during an active stream.

PR #349, opened 2026-06-01
fix(warm-pool): reap orphaned warm sandboxes after restarts

arjunblj:
Reaping only walked the in-memory pool, so warm sandboxes from prior processes leaked (hundreds of stale Sandbox pods). Parallelize bulk stops and add a grace-gated periodic orphan sweep.


Zygimantass:

PR #362, opened 2026-06-02
feat(web): web client renderer for Centaur

goksu:
## Summary
- add a Regen-based Centaur web service that streams Rust V2 sessions
- add a browser renderer output target for the generic renderer contract
- wire api-rs and web into Helm, image publishing, and local dev secrets
- add Rust API Docker/runtime fixes for the local V2 flow

## Validation
- pnpm --filter @centaur/rendering test
- pnpm --filter @centaur/rendering typecheck
- pnpm --filter centaur-web test
- pnpm --filter centaur-web check:types
- pnpm --filter centaur-web build
- cargo test -p centaur-sandbox-agent-k8s -p centaur-session-runtime -p centaur-api-server
- helm template centaur contrib/chart -n centaur -f contrib/chart/values.dev.yaml
- just build-one api-rs
- just build-one web
- just deploy
- in-app browser smoke at http://127.0.0.1:3003 returned assistant PONG

### DIFF
--- a/contrib/chart/templates/networkpolicy.yaml
+++ b/contrib/chart/templates/networkpolicy.yaml
@@ -1,6 +1,5 @@
 {{- if .Values.networkPolicy.enabled }}
 {{- $ingressSourceNamespaces := default .Values.networkPolicy.ingressControllerNamespaces .Values.networkPolicy.ingressSourceNamespaces }}
-{{- $webIngressSourceNamespaces := default $ingressSourceNamespaces .Values.networkPolicy.webIngressSourceNamespaces }}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
@@ -286,7 +285,7 @@ spec:
   ingress:
     # Only health probes reach discordbot; nothing routes traffic to it (the Gateway is outbound).
     - from:
-{{- range $webIngressSourceNamespaces }}
+{{- range $ingressSourceNamespaces }}
         - namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: {{ . | quote }}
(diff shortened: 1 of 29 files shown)


services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -32,6 +34,10 @@ async fn main() -> Result<(), ServerError> {
     Ok(())
 }
 
+fn install_rustls_crypto_provider() {
+    let _ = rustls::crypto::ring::default_provider().install_default();
+}
+
```
goksu: Why did we have to make this change? Investigate further.

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -49,6 +55,9 @@ async fn sandbox_runtime_from_args(args: &Args) -> Result<SandboxRuntime, Server
         )),
         SandboxBackendKind::AgentK8s => {
             let mut config = AgentSandboxConfig::new(args.session_sandbox_k8s_namespace.clone());
+            config
+                .labels
+                .insert("centaur.ai/managed".to_owned(), "true".to_owned());
```
goksu: Also feels odd that we would have to do this for a web renderer.

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -90,12 +99,29 @@ fn container_workload_mode(args: &Args) -> SandboxWorkloadMode {
...
+                        secret_name: secret_name.to_owned(),
+                    },
+                    args.session_sandbox_firewall_ca_mount_path.clone(),
+                )
+                .read_only(),
+            );
+        }
     }
+    workload
 }
```
goksu: If we had a separate API layer for web to call would we be able to avoid adding this CA cert stuff?

goksu (commented):
Design updates:

- Update it to be dark mode only.
- Use Regen but with minimal rounding on the next pass.
- Only existing threads on the left panel with "New Thread" button top left.
- Streaming message chatbox flow can stay as is but polish it a bit more.

goksu:
Closing for now. Will bring it back later when Slackbot is more stable.

PR #326, opened 2026-06-01
[codex] add Rust iron-proxy config renderer

gakonst:
## Summary

Draft checkpoint for the Rust API iron-proxy work, stacked on #325.

- Adds a small `centaur-iron-proxy` crate that composes native iron-proxy YAML fragments instead of parsing tool `pyproject.toml`.
- Supports operator-owned secret sources: fragments can define placeholders/rules and omit `source`; the renderer fills env / 1Password / 1Password Connect bindings from `SourcePolicy`.
- Preserves the Python-supported managed transform families through fragments (`secrets`, `gcp_auth`, `oauth_token`, `hmac_sign`) plus `postgres` listeners and other top-level iron-proxy config.
- Handles both current nested `replace.proxy_value` and legacy top-level `proxy_value` secret shapes when deriving sandbox placeholder env and filling missing operator-owned `source` blocks.
- Reads both unprefixed and Kubernetes-prefixed proxy source envs, so `KUBERNETES_FIREWALL_MANAGER_SECRET_SOURCE` / TTL fallbacks behave like the existing deployment knobs.
- Discovers `iron.yaml` fragments from `SESSION_SANDBOX_IRON_PROXY_FRAGMENT_DIRS`, falling back to `TOOL_DIRS`, so policy can live next to local CLI tools without a tool manager.
- Enables stock per-sandbox iron-proxy config when the chart-style Kubernetes firewall CA secrets are present, so built-in infra/harness fragments apply even when no tool-local `iron.yaml` exists.
- Adds source-free built-in fragments for common infra placeholders plus Codex and Claude Code API-key / access-token modes, with token-broker TTL filled by `FIREWALL_MANAGER_TOKEN_BROKER_TTL`.
- Adds fragment-level `broker_credentials` entries and renders `iron-token-broker.yaml` from those fragments, including built-in Codex / Claude Code access-token credentials and the broker rule that writable refresh-token stores cannot be env-backed.
- Reconciles the chart-managed iron-token-broker ConfigMap from Rust and patches the broker Deployment config-hash when rendered credentials change, matching the Python API broker rollout path without pyproject parsing.
[...]

PR #365, opened 2026-06-02
[codex] add focused api-rs iron proxy integration

gakonst:
## Summary

- add a focused `centaur-iron-proxy` Rust crate that renders iron-proxy and token-broker YAML from typed serde models plus small YAML fragments
- wire api-rs clap config into typed `IronProxyConfig` without the broader CLI/module refactor from #326/#363
- add Kubernetes sandbox iron-proxy integration: per-sandbox proxy pod/service/configmap, CA mount, proxy env injection, network policies, and token-broker configmap reconciliation
- add built-in infra and Codex/Claude harness fragments for API-key and access-token modes
- ignore `services/api-rs/target` so Docker builds do not slurp the local Rust target dir

## Scope notes

- based directly on `api-rs-control-plane`
- no generated Agent Sandbox CRD replacement/removal
- no session-cli split or broader non-iron refactor
- leaves existing PR #326 untouched

## Validation

- `cargo check -p centaur-api-server -p centaur-sandbox-agent-k8s -p centaur-iron-proxy`
- `cargo fmt -p centaur-api-server -p centaur-sandbox-agent-k8s -p centaur-iron-proxy`
- `cargo clippy -p centaur-api-server -p centaur-sandbox-agent-k8s -p centaur-iron-proxy --all-targets -- -D warnings`
- `cargo test -p centaur-api-server -p centaur-sandbox-agent-k8s -p centaur-iron-proxy`
- `cargo build -p centaur-api-server -p centaur-session-cli`

Live Codex/1Password `gh auth status` e2e was not completed before this branch split request; this PR is opened separately so that proof can happen against this clean branch without mutating #326.


services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -66,85 +83,430 @@ async fn sandbox_runtime_from_args(args: &Args) -> Result<SandboxRuntime, Server
...
+            _ => Err(ServerError::MissingIronProxyCaSecret),
+        }
+    }
+}
+
+#[derive(Debug, ClapArgs)]
+struct IronProxySourceArgs {
+    #[arg(
+        long = "kubernetes-firewall-manager-secret-source",
+        env = "KUBERNETES_FIREWALL_MANAGER_SECRET_SOURCE",
```
Zygimantass: This env name drifts from the existing Helm/Python contract. The chart still sets `FIREWALL_MANAGER_SECRET_SOURCE`, `FIREWALL_MANAGER_SECRET_TTL`, and `FIREWALL_MANAGER_TOKEN_BROKER_TTL`, but this parser only reads the new `KUBERNETES_*` variants. With the current chart, `ironProxy.secretSource=onepassword` is ignored, the Rust path falls back to `env`, and token-broker reconciliation will fail for access-token fragments because broker stores cannot use env sources. Please either keep accepting the existing env names here or update the chart/docs in this PR.

services/api-rs/crates/centaur-sandbox-agent-k8s/src/iron_proxy.rs
```diff
@@ -0,0 +1,1003 @@
...
+    for (name, value) in &iron_proxy.extra_env {
+        env.insert(name.clone(), env_var(name, value));
+    }
+    if let Some(name) = &iron_proxy.token_broker_name {
+        env.insert(
+            "IRON_BROKER_URL".to_owned(),
+            env_var(
+                "IRON_BROKER_URL",
+                &format!(
+                    "http://{name}:{}",
```
Zygimantass: This reconstructs the broker URL from the name and the library default port, ignoring `KUBERNETES_TOKEN_BROKER_URL` and the chart's configurable `tokenBroker.service.httpPort`. A cluster with a non-default broker port, or any future host override, will render `IRON_BROKER_URL` to the wrong endpoint. Carry the configured broker URL through `IronProxyConfig` and use that value directly.

services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
```diff
@@ -226,13 +235,23 @@ impl SandboxBackend for AgentSandboxBackend {
 
     async fn create(&self, spec: SandboxSpec) -> SandboxResult<SandboxHandle> {
         let id = SandboxId::new(next_sandbox_name());
+        let mut spec = spec;
+        let resolved_iron_proxy = self.resolve_iron_proxy(&id, &spec)?;
+        if let Some(resolved) = &resolved_iron_proxy {
+            iron_proxy::apply_proxy_env(&mut spec, resolved);
+        }
+        self.create_iron_proxy_resources(&id, resolved_iron_proxy.as_ref())
```
Zygimantass: If `create_iron_proxy_resources` fails after creating any resources, for example a NetworkPolicy create succeeds but the proxy pod never becomes Ready, this returns without deleting the ConfigMap/Service/NetworkPolicies/proxy pod. The cleanup below only handles failures after the Sandbox CR create. Please wrap this call the same way and delete the proxy resources before returning the error so partial create failures do not leak resources.

services/api-rs/crates/centaur-sandbox-agent-k8s/src/iron_proxy.rs
```diff
@@ -0,0 +1,1003 @@
...
+
+fn token_broker_pod_labels() -> BTreeMap<String, String> {
+    BTreeMap::from([(
+        "app.kubernetes.io/component".to_owned(),
+        "token-broker".to_owned(),
+    )])
+}
+
+fn proxy_password() -> String {
+    format!("pg-{}", unique_suffix())
```
Zygimantass: This proxy password is derived from the current millisecond, so it is predictable and can collide when multiple password envs are generated in the same millisecond. The existing Python path used a random token for this shared sandbox-to-proxy Postgres password. Please generate this with a CSPRNG-backed token instead of `unique_suffix()`.

Zygimantass (commented):
First-pass structure read: the crate split is pointed in the right direction. Keeping YAML rendering/source-policy logic in `centaur-iron-proxy` and Kubernetes object construction in `centaur-sandbox-agent-k8s` is a workable boundary.

The risky part is the K8s side: most behavior is in one large helper and several contracts are string-built against the existing Helm/Python configuration surface. The inline comments are the concrete bugs I found on this pass. I also think this needs Rust-side tests for the rendered K8s resources, especially env compatibility, broker wiring, cleanup-on-partial-create, and PG password generation.

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -37,22 +49,27 @@ fn init_tracing() {
     tracing_fmt().with_env_filter(filter).json().init();
 }
 
+fn init_crypto_provider() {
```
Zygimantass: this feels slop, figure out how to avoid this

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -37,22 +49,27 @@ fn init_tracing() {
...
+            args.sandbox.local_workload_mode()?,
         )),
         SandboxBackendKind::AgentK8s => {
-            let mut config = AgentSandboxConfig::new(args.session_sandbox_k8s_namespace.clone());
-            config.image_pull_policy = args.session_sandbox_image_pull_policy.clone();
-            config.ready_timeout = Duration::from_secs(args.session_sandbox_ready_timeout_secs);
+            let mut config = AgentSandboxConfig::new(args.sandbox.k8s_namespace.clone());
+            config.image_pull_policy = args.sandbox.agent_image_pull_policy.clone();
+            config.ready_timeout = Duration::from_secs(args.sandbox.ready_timeout_secs);
+            config.iron_proxy = args.sandbox.iron_proxy.to_config()?;
```
Zygimantass: is there a way to have `impl From<Args> for AgentK8sConfig` instead of handling it here?

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -66,85 +83,441 @@ async fn sandbox_runtime_from_args(args: &Args) -> Result<SandboxRuntime, Server
...
+    agent_image: Option<String>,
+    #[arg(
+        long = "session-sandbox-image-pull-policy",
+        alias = "kubernetes-agent-image-pull-policy",
+        env = "SESSION_SANDBOX_IMAGE_PULL_POLICY"
+    )]
+    agent_image_pull_policy: Option<String>,
+    #[arg(
+        long = "session-sandbox-ready-timeout-secs",
+        alias = "kubernetes-sandbox-ready-timeout-s",
```
Zygimantass: no aliases please, just call it kubernetes_xxx

Zygimantass: no env vars starting with SESSION_

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -66,85 +83,441 @@ async fn sandbox_runtime_from_args(args: &Args) -> Result<SandboxRuntime, Server
...
+                .or(self.centaur_api_url.as_deref())
+                .unwrap_or("http://api:8000")
+                .to_owned(),
+        );
+        if let Some(api_key) = self
+            .centaur_api_key_override
+            .as_deref()
+            .or(self.centaur_api_key.as_deref())
+        {
+            push_env(&mut envs, "CENTAUR_API_KEY", api_key.to_owned());
```
Zygimantass: remove, we dont have centaur_api_keys yet

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -66,85 +83,441 @@ async fn sandbox_runtime_from_args(args: &Args) -> Result<SandboxRuntime, Server
...
+                "centaur-firewall-ca-key".to_owned(),
+            ))),
+            (None, None) => Ok(None),
+            _ => Err(ServerError::MissingIronProxyCaSecret),
+        }
+    }
+}
+
+#[derive(Debug, ClapArgs)]
+struct IronProxySourceArgs {
```
Zygimantass: move all args related stuff to either config.rs or args.rs or something

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -168,32 +541,43 @@ fn default_sandbox_image(workload: SandboxWorkloadKind) -> &'static str {
...
-    if let Some(api_key) = args
-        .session_sandbox_centaur_api_key
-        .as_deref()
-        .or(args.centaur_api_key.as_deref())
-    {
-        push_env(&mut envs, "CENTAUR_API_KEY", api_key.to_owned());
+fn harness_auth_mode_env(engine: &str) -> Option<String> {
+    match engine {
+        "codex" => env::var("CODEX_AUTH_MODE").ok(),
+        "claude-code" | "claudecode" => env::var("CLAUDE_CODE_AUTH_MODE").ok(),
```
Zygimantass: can we use the harness enum or whatever?

services/api-rs/crates/centaur-api-server/src/main.rs
```diff
@@ -168,32 +541,43 @@ fn default_sandbox_image(workload: SandboxWorkloadKind) -> &'static str {
...
+            return Err(format!("label selector item {item:?} must be key=value"));
         }
+        labels.insert(key.to_owned(), value.to_owned());
     }
+    Ok(labels)
+}
 
-    envs
+fn insert_non_empty(values: &mut BTreeSet<String>, value: &str) {
+    let value = value.trim();
```
Zygimantass: inline

services/api-rs/crates/centaur-iron-proxy/src/model/postgres.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub upstream: Option<PostgresUpstream>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub client: Option<PostgresClient>,
+    #[serde(default, skip_serializing)]
+    pub sandbox_env: Option<SandboxEnv>,
+    #[serde(default, flatten)]
+    pub extra: BTreeMap<String, Value>,
+}
+
+impl PostgresListener {
```
Zygimantass, line 27: what is this for? can't we use the sqlx stuff?

services/api-rs/crates/centaur-iron-proxy/src/fragment.rs
```diff
@@ -0,0 +1,120 @@
+use std::{
+    collections::BTreeMap,
+    fs,
+    path::{Path, PathBuf},
+};
+
+use crate::{IronProxyConfigError, ProxyFragment, Result};
+
+pub const DEFAULT_PROXY_BASE_CONFIG: &str =
```
Zygimantass: why do we hard code these instead of discovering them dynamically?

PR #370, opened 2026-06-02
feat: arm docker images

OisinKyne:
## Problem to be solved

I'm trying to run centaur on a Mac running k3d (specifically the [Obol-stack](https://github.com/ObolNetwork/obol-stack)), and have to build the four images locally currently. 

## Proposed solution

Build ARM images during CI.

I have not tested this fully because its not my GH action to run, but it looks simple enough to submit, and I checked the action SHA is [real](https://github.com/docker/setup-qemu-action/commit/06116385d9baf250c9f4dcb4858b16962ea869c3) and that all dockerfiles are multi-arch ready. 

This will slow the action considerably due to emulation, its possible that splitting runners and using a dedicated arm one might be faster. Not super precious here about how best to go about it. Cheers.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -62,7 +62,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


Zygimantass:
closed in favor of #525

PR #378, opened 2026-06-02
feat(api-rs): integrate iron-control for per-principal proxy credential grants

mslipper:
Integrates iron-control as the control plane for per-sandbox egress credentials. A new `centaur-iron-control` crate provides the typed admin client, fragment→resource translation, and principal derivation. Sessions upsert an iron-control principal (Slack user for DMs, channel otherwise) and assign infra + tool/harness roles; per-sandbox iron-proxies are registered to that principal and sync config over `IRON_CONTROL_URL` with an `iprx_` token rather than a rendered static config.

The whole path is gated on iron-control being configured, so the legacy rendered-proxy path stays as a fallback (true hard-replace is making sync unconditional later). Helm wiring (api-rs env + egress NetworkPolicy) is behind `ironControl.enabled`.

Not yet validated against a live cluster; the egress data path needs a smoke test. Warm-pool principal rebinding via PATCH is implemented in the client but dormant until api-rs grows a warm pool.


gakonst:
If we integrate iron control can we delete a bunch of code from the underlying PR regarding the iron proxy work?

mslipper:
Yes this PR will do a lot of that by the time I'm done.

PR #381, opened 2026-06-03
fix: make overlay dep grep non-fatal under pipefail (#281)

### LINKED ISSUE #281: api entrypoint.sh crashes when an overlay tool dir yields zero pip deps (fatal grep under pipefail)
## Summary

`services/api/entrypoint.sh` crashes the api container (silent exit 1, no logs) when a bind-mounted **overlay** tool directory exists but yields **zero** pip dependencies. The api never reaches `exec uvicorn`, so the pod CrashLoopBackOffs with 0-byte logs.

## Root cause

The script runs under `set -euo pipefail`. It collects overlay tool deps, then:

```bash
if [[ -n "${_extra_deps}" ]]; then
  echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt
  uv pip install -r /tmp/_extra_deps.txt --quiet 2>/dev/null || true
  rm -f /tmp/_extra_deps.txt
fi
```

When an overlay tool dir exists, the per-dir loop appends the python helper's output (empty, when there are no `project.dependencies`) followed by a literal newline:

```bash
_extra_deps+=$(python3 -c "... print('\n'.join(sorted(deps)))" 2>/dev/null || true)
_extra_deps+=$'\n'
```

So `_extra_deps` becomes a lone newline. `[[ -n "${_extra_deps}" ]]` is **true** (a newline is non-empty), the block runs, and `grep -v '^$'` matches nothing and **exits 1**. That line has no `|| true`, so `pipefail` + `set -e` abort the script before `exec "$@"`.

## Reproduction

Mount an overlay whose `tools/` dir contains a single tool with `dependencies = []` (or just a placeholder package with no `pyproject.toml` deps). The api pod will CrashLoopBackOff; the `overlay-bootstrap` init container succeeds (exit 0) and container logs are empty.

## Suggested fix

Make the grep non-fatal, e.g.:

```bash
echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt || true
```

or only install when there is at least one non-blank line:

```bash
mapfile -t _deps < <(echo "$_extra_deps" | sort -u | grep -v '^$')
if (( ${#_deps[@]} )); then
  printf '%s\n' "${_deps[@]}" > /tmp/_extra_deps.txt
  uv pip install -r /tmp/_extra_deps.txt --quiet 2>/dev/null || true
  rm -f /tmp/_extra_deps.txt
fi
```

Either makes an overlay with dependency-free tools safe.

jeremylongshore:
Fixes #281.

`services/api/entrypoint.sh` runs under `set -euo pipefail`. When an overlay tool dir exists but yields zero pip deps, `_extra_deps` is just a newline, so the `[[ -n "${_extra_deps}" ]]` guard passes but `grep -v '^$'` matches nothing and exits 1. With no `|| true`, the script aborts before `exec "$@"` and the api pod CrashLoopBackOffs with empty logs.

One-line guard on the grep, as suggested in the issue:

```diff
-    echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt
+    echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt || true
```

No effect on the success path: with real deps `grep` exits 0 and the guard is a no-op; with none, the empty file is already handled by the guarded `uv pip install -r ... || true` on the next line.

Tested: `shellcheck` clean, `bash -n` ok, and an isolated repro of the fragment crashes before / passes after. I couldn't run the full overlay-mount path end-to-end without the cluster.

### DIFF
--- a/services/api/entrypoint.sh
+++ b/services/api/entrypoint.sh
@@ -27,7 +27,7 @@ print('\n'.join(sorted(deps)))
     _extra_deps+=$'\n'
   done
   if [[ -n "${_extra_deps}" ]]; then
-    echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt
+    echo "$_extra_deps" | sort -u | grep -v '^$' > /tmp/_extra_deps.txt || true
     uv pip install -r /tmp/_extra_deps.txt --quiet 2>/dev/null || true
     rm -f /tmp/_extra_deps.txt
   fi


Zygimantass:

PR #384, opened 2026-06-03
fix(slackbotv2): harden api-rs handoff and stream recovery

Zygimantass:
## Summary
- wait for Slackbot v2's durable api-rs handoff before returning Slack 200 for message/app_mention events
- return 503 on retryable api-rs/network failures so Slack retries the same event
- clear Chat SDK dedupe on retryable failure and persist append/execute progress separately to avoid duplicate appends on retry
- store unfinished Slack stream render obligations on the existing Chat SDK `thread-state:*` record and replay them on Slackbot startup
- keep a small Chat SDK state-list index of candidate thread ids plus short recovery leases so restarted Slackbot instances can reopen api-rs `/events` from the last event id without a separate Slackbot table

## Tests
- bun run --cwd services/slackbotv2 check:types
- bun run --cwd services/slackbotv2 test
- just build-one slackbotv2
- curl -fsS http://127.0.0.1:3002/health
- curl -fsS http://127.0.0.1:8080/healthz

## Notes
- Did not run `just deploy` because the active kubectl context was `prd-centaur-na.tail388b2e.ts.net`; local service health was checked instead.

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from 'node:async_hooks'
 import { Hono } from 'hono'
 import {
   Chat,
@@ -17,6 +18,8 @@ import {
 import {
   collectInitialContext,
   forwardToSessionApi,
+  isRetryableSessionApiError,
+  openSessionEventStream,
   serializeMessage,
   sessionStreamError,
   startingStreamNotification
@@ -62,6 +65,13 @@ type SlackAssistantAdapter = {
   setAssistantTitle?(channelId: string, threadTs: string, title: string): Promise<void>
 }
 
+type SlackbotV2RequestContext = {
+  retryableErrors: unknown[]
+  waitUntil(promise: Promise<unknown>): void
+}
+
+const requestContext = new AsyncLocalStorage<SlackbotV2RequestContext>()
+
 export function createSlackbotV2(options: SlackbotV2Options): SlackbotV2 {
   const userName = options.userName ?? 'centaur'
   const logger = options.logger ?? noopLogger
@@ -73,10 +83,11 @@ export function createSlackbotV2(options: SlackbotV2Options): SlackbotV2 {
     userName,
     logger
   })
+  const state = options.state ?? createDefaultState(options, logger)
   const chat = new Chat({
     userName,
     adapters: { slack },
-    state: options.state ?? createDefaultState(options, logger),
+    state,
     onLockConflict: 'force',
     logger
   })
@@ -86,7 +97,8 @@ export function createSlackbotV2(options: SlackbotV2Options): SlackbotV2 {
     await thread.subscribe()
     await syncThreadMessageToSession(thread, message, {
       mode: 'execute',
-      options
+      options,
+      state
     })
   })
 
@@ -94,7 +106,8 @@ export function createSlackbotV2(options: SlackbotV2Options): SlackbotV2 {
     if (!isAllowedSlackMessage(message, options, logger)) return
     await syncThreadMessageToSession(thread, message, {
       mode: message.isMention === true ? 'execute' : 'append',
-      options
+      options,
+      state
     })
   })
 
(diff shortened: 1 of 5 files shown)


services/slackbotv2/src/index.ts
```diff
@@ -289,6 +645,53 @@ async function* executeAndStreamSession(
...
+      error: errorMessage(deleteError)
+    })
+  }
+  traceLog(options, 'slackbotv2_webhook_retry_marked', trace, {
+    error: errorMessage(error)
+  })
+  return true
+}
+
+function backgroundWaitUntil(promise: Promise<unknown>): void {
```
Zygimantass, line 727: inline this please

services/slackbotv2/src/index.ts
```diff
@@ -289,6 +645,53 @@ async function* executeAndStreamSession(
   }
 }
 
+async function* streamError(error: unknown): AsyncIterable<SlackbotV2RendererSource> {
+  yield sessionStreamError(error)
+}
+
+async function requestSlackRetry(
```
Zygimantass: what is this meant for

services/slackbotv2/src/index.ts
```diff
@@ -172,77 +228,380 @@ async function syncThreadMessageToSession(
...
+    })
+    traceLog(input.options, 'slackbotv2_forward_execution_committed', trace, {
+      executed_message_count: Math.min(latestExecutedMessageIds.size, 1000)
     })
   }
 
   if (!shouldStartExecution) {
-    await forwardToSessionApi(input.options, forwardInput)
-    await commitForwardedState()
+    try {
```
Zygimantass, line 296: why did we add retries here?

services/slackbotv2/src/index.ts
```diff
@@ -172,77 +228,380 @@ async function syncThreadMessageToSession(
...
+): AsyncIterable<SlackbotV2RendererSource> {
+  yield startingStreamNotification(input.threadId)
+  for await (const event of stream) yield event
+}
+
+function renderRecoveryLeaseKey(threadId: string): string {
+  return `slackbotv2:render:lease:${threadId}`
+}
+
+function canRecoverRenderObligation(value: SlackbotV2RenderObligation): boolean {
```
Zygimantass: this seems like a codesmell

services/slackbotv2/src/index.ts
```diff
@@ -172,77 +228,380 @@ async function syncThreadMessageToSession(
...
+  return (
+    typeof value.afterEventId === 'number' &&
+    Boolean(message) &&
+    typeof message.id === 'string' &&
+    typeof message.text === 'string' &&
+    typeof message.threadId === 'string'
+  )
+}
+
+function validDateOrNow(value: string): Date {
```
Zygimantass: also codesmell, we should be able to do this using typescript type safety

services/slackbotv2/src/index.ts
```diff
@@ -172,77 +228,380 @@ async function syncThreadMessageToSession(
...
+  }
+): Promise<void> {
+  await state.appendToList(RENDER_OBLIGATION_INDEX_KEY, input.threadId, {
+    maxLength: RENDER_OBLIGATION_INDEX_MAX_LENGTH,
+    ttlMs: RENDER_INDEX_TTL_MS
+  })
+  traceLog(input.options, 'slackbotv2_render_obligation_indexed', input.trace)
+}
+
+function recoveryThread(
```
Zygimantass: codesmell

Zygimantass: Fixed in `1b73cf14`: removed the manual `ThreadImpl` reconstruction. Recovery now uses the existing `chat.thread(threadId)` and calls `thread.adapter.stream(...)` directly with typed Slack recipient context from the normalized Slack message (`teamId` + author user id).

services/slackbotv2/src/index.ts
Zygimantass: Fixed in `c25888da`: removed the extra defensive recovery checks/helper and switched this path to typed state plus serialization-time coercion only.

services/slackbotv2/src/index.ts
Zygimantass: This is not an internal retry loop. It only marks the current Slack webhook response as retryable when api-rs is temporarily unavailable, then clears the Slack-event dedupe key so Slack’s next delivery attempt is accepted instead of dropped. Terminal execution/render failures still fall through to the normal Slack-visible error path.

PR #396, opened 2026-06-03
fix: proxy lowercase OpenAI auth header for Codex

tarrencev:
## Summary

- Include lowercase `authorization` alongside `Authorization` when proxying the Codex `OPENAI_API_KEY` placeholder.
- Add a regression assertion for Codex API-key harness secret selection.

## Why

Codex's OpenAI Responses client can send the placeholder credential in a lowercase authorization header. The per-sandbox iron-proxy config only scanned `Authorization`, so the placeholder could leak through unchanged and OpenAI returned 401.

## Validation

- `uv --directory services/api run pytest tests/test_tool_manager.py -k codex_api_key`
- `uv --directory services/api run pytest tests/test_sandbox_kubernetes_backend.py -k codex_api_key`
- `uv --directory services/api run ruff check api/tool_manager.py tests/test_tool_manager.py`
- Live downstream validation in the C7E overlay deployment: a Codex-backed memory workflow dry-run completed successfully after adding this behavior as an overlay shim.

## Known Test Gap

- `uv --directory services/api run pytest tests/test_tool_manager.py` currently fails on `test_tool_rest_router_lists_describes_and_invokes_tools` because `/tools` returns `401 Missing API key`; that appears unrelated to this patch and reproducible from current router auth behavior.

### DIFF
--- a/contrib/chart/templates/workloads.yaml
+++ b/contrib/chart/templates/workloads.yaml
@@ -378,6 +378,12 @@ spec:
             - name: KUBERNETES_SANDBOX_RUNTIME_CLASS_NAME
               value: {{ .Values.sandbox.runtimeClassName | quote }}
 {{- end }}
+{{- if .Values.sandbox.serviceAccountName }}
+            - name: KUBERNETES_SANDBOX_SERVICE_ACCOUNT_NAME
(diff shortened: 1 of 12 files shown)


gakonst:
closing / outdated lmk if still a bug

PR #401, opened 2026-06-04
feat: Discord chat ingress (discordbot)

0xdiid:
Adds Discord as a chat ingress for centaur. It's deliberately a close clone of slackbotv2 - same session lifecycle against the api-rs control plane, same streaming/rendering loop - with Discord-shaped edges. Every deliberate divergence is commented in-code at its site, and an end-to-end regression harness (ported from slackbotv2's) keeps the two in lockstep.

## How a mention flows

```mermaid
sequenceDiagram
    participant U as User
    participant D as Discord
    participant B as discordbot
    participant A as api-rs

    U->>D: @centaur question
    D->>B: gateway event
    B->>B: guild allowlist (fail closed, checked before any thread is created)
    B->>D: create thread + 👀 on the triggering message
    B->>A: create session, forward context, execute
    A-->>B: event stream
    B->>D: reasoning posted as small subtext lines
    B->>D: answer streamed into its own message(s), split at Discord's length limit
    B->>D: 👀 → ✅ (or ❌)
```

## Compared to slackbotv2

| | slackbotv2 | discordbot |
|---|---|---|
| Transport | webhooks, failed deliveries are retried by Slack | persistent gateway connection, at-most-once |
| Auth boundary | fail-open org check | fail-closed guild allowlist (inert until configured) |
| Progress UX | assistant status + thread title | 👀/✅/❌ reactions + subtext reasoning lines |
| Answer rendering | one streamed message (generous length limit) | split across messages at the 2k limit, honestly truncated when forced |
| Lost-delivery story | Slack redelivers | bounded in-process retries, reaction feedback when a message is dropped |
| Context quirks | history includes everything | thread starters live in the parent channel and alerts arrive as embeds, so both are fetched/flattened in |

## Testing
In addition to automated tests, we've been running this branch for about a week and a half. Here's a preview of what that looks like:

Screenshot 2026-06-11 at 6 34 55 AM


services/api/api/iron-proxy.base.yaml
```diff
@@ -0,0 +1,93 @@
+dns:
```
0xdiid: did this file get added back in by a merge or something? I don't think we should need to edit this

services/api/api/runtime_control.py
```diff
@@ -253,7 +253,7 @@ def _agent_session_title(
 _DEFAULT_CLAUDE_MODEL = "claude-fable-5"
 
 _CLAUDE_MODEL_ALIASES: dict[str, str] = {
-    "fable": "claude-fable-5",
+    "fable": _DEFAULT_CLAUDE_MODEL,
```
0xdiid: weird change, lets undo

services/api/api/iron-proxy.base.yaml
0xdiid: [CLAUDE] Not a merge artifact — this restore is deliberate (commit fb12a5f4). PR #378 (the iron-control integration on `api-rs-control-plane`) deleted this file but left `proxy_config.py` reading it unconditionally: `load_base_config()` is the default seed for `render_proxy_yaml()`, which `sandbox/kubernetes.py` calls without a `base_config` when rendering the proxy ConfigMap, and most of `test_proxy_config.py` does the same. On the base branch as it stands, that path raises `FileNotFoundError`. If the intent in #378 was to drop the file for good, the `load_base_config` default would need to be removed or repointed first — happy to do that, but it felt out of scope for this PR, so I restored the file instead.

patches/@chat-adapter__discord@4.30.0.patch
```diff
@@ -0,0 +1,434 @@
+diff --git a/dist/index.d.ts b/dist/index.d.ts
```
0xdiid, line 1: would it make sense to make this change upstream? I made a fork (https://github.com/0xSplits/chat) that we can submit a draft pr on if so.

tools/infra/grafana/pyproject.toml
```diff
@@ -26,8 +26,18 @@ build-backend = "hatchling.build"
 
 [tool.centaur]
 module = "client.py"
-hosts = []
+# Egress allowlist + the host(s) the GRAFANA_API_KEY Bearer token is scoped to.
+# MUST be non-empty: an empty list renders the iron-proxy `secrets` transform
+# with no host rules, so the token is never injected AND egress to Grafana is
+# denied (every call 405s/401s through the CONNECT-only proxy).
+hosts = ["splits.grafana.net"]
```
0xdiid: can we make this more general and/or include it in our overlay? this shouldn't be committed to the upstream

services/api/api/iron-proxy.base.yaml
0xdiid: Yeah lets drop it. I think it was necessary because we were running the python api in tandem, but we don't need to do that anymore.

0xdiid: [CLAUDE] Dropped in b2e71419 — the file is deleted again, so this PR no longer touches it. Heads up that `proxy_config.py` (`load_base_config`) still reads it unconditionally, so the Python api's `render_proxy_yaml()` default path and most of `test_proxy_config.py` are broken on this branch line — pre-existing on `api-rs-control-plane`, fine if the Python api is fully retired, but worth a follow-up to strip that dead path.

services/api/api/runtime_control.py
0xdiid: [CLAUDE] Undone in b2e71419 — the file is now untouched by this PR.

tools/infra/grafana/pyproject.toml
0xdiid: [CLAUDE] Done in b2e71419 — `hosts` is back to `[]` with a comment that deployments must override it (e.g. via a tool overlay); `splits.grafana.net` needs to move into our overlay as a follow-up. Kept the general fix in this file: `GRAFANA_URL` moved out of `secrets` because the client reads it in-process, so it can never be a wire-swapped placeholder.

patches/@chat-adapter__discord@4.30.0.patch
0xdiid: [CLAUDE] Yes — everything in this patch is general Discord-platform safety, nothing Splits-specific: @everyone/@here no longer counts as a bot mention, mention gating before thread creation, `allowed_mentions` defanged on outbound create/edit, removal of the blanket @word→mention rewrite, 429 retry with `retry_after`, and embeds preserved on live gateway messages. Caveat: this pnpm patch targets the published `dist/` output, so an upstream PR means porting the changes to the adapter's TypeScript source in the fork — and we keep carrying this patch until a release with the fixes ships. Happy to put up that draft PR on 0xSplits/chat if you want to go ahead.

tools/infra/grafana/pyproject.toml
0xdiid: [CLAUDE] Overlay side is done: https://github.com/0xSplits/gerard/pull/39 (draft) adds a copy of the grafana tool that shadows the upstream definition by name and pins `hosts` to our Grafana stack.

0xdiid: [CLAUDE] Better approach landed in b96564da: `hosts = ["*.grafana.net"]` — generic for upstream (covers any Grafana Cloud stack; self-hosted overrides via overlay) and the iron-proxy host rules already support wildcards (`*.googleapis.com`, `logs.*.amazonaws.com` use the same path). The gerard shadow-copy PR is closed as superseded.

patches/@chat-adapter__discord@4.30.0.patch
0xdiid: Yeah lets go ahead and submit a draft PR that addresses it upstream, but like you said this wouldn't change this PR

Issue #358, opened 2026-06-02
Centaur Slack live delivery suppresses final reply when Codex completes with empty result


GoonMachine:
## Summary

When Codex emits an empty `turn.done`, Centaur can mark the execution completed and suppress fallback Slack delivery even though Slack streamed zero answer text. The user sees "Thinking" briefly and then no response.

## Observed

Execution: `exe_048db3295bf34737`
Thread key: `slack:T03USE6Q064:C083QLFNXPY:1780358702.780989`

Facts from the execution:

```text
duration: 0.32s
raw_event_count: 1
only raw event: { "type": "turn.done", "result": "" }
assistant_text_chars: 0
command_events: 0
tool_result_events: 0
status: completed
terminal_reason: completed
```

Slack delivery logs showed:

```text
slackbot_streamed_answer_chars: 0
result_size_bytes: 2
event: final_delivery_skipped_without_live_answer
event: final_delivery_skipped
reason: slackbot_live_delivery
```

Slackbot terminal event also showed:

```text
result_text_chars: 0
answer_chars_before_event: 0
streamed_answer_chars_before_event: 0
task_count: 0
```

## Expected

If live Slack delivery streamed zero answer chars and the final Codex result is empty, Centaur should not treat the response as successfully delivered.

It should do one of:

- retry or requeue the turn,
- mark the execution as no_output / failed / degraded,
- post a visible Slack status explaining that the turn produced no answer,
- or fall back to non-live final delivery when possible.

## Why It Matters

From Slack, this looks like Centaur started working and then silently disappeared. In the database, the run is recorded as successful, which makes the failure hard to detect.

## Possible Root Cause

The live-delivery suppression logic appears to treat empty result text as "covered" by Slack live delivery, even when `slackbot_streamed_answer_chars` is zero.

Relevant code paths to inspect:

- `services/slackbot/src/slack/codex-session.ts`
- `services/api/api/runtime_control.py`

goksu:
Closing as this is no longer applicable after the Rust API / common Codex harness updates.

PR #408, opened 2026-06-04
Fix Slack live delivery status and empty fallback

Geeknerd1337:
# PR Description

## Problem

Long-running Slack sessions could look dead even while Centaur was still running.

The tool/task stream continued internally, but Slack's assistant status was cleared after the first visible streamed output. For long executions, that meant the thread could stop showing an active "Thinking..." state even though tool calls and chain-of-thought style progress were still being processed.

We also found a related blank-delivery edge case. If Slack live delivery was enabled but the model produced zero visible answer characters, the API treated live delivery as already covering the result. That suppressed the durable final-delivery fallback, so a no-output execution could complete without posting a useful fallback message.

## Fix

- Keep Slack assistant status active for the whole agent session.
- Refresh the `Thinking...` status every 30 seconds while the session is active.
- Clear assistant status only when `done()` runs.
- Stop retrying status refreshes if Slack reports that assistant status is unsupported for the channel/thread.
- Avoid opening a blank Slack stream when a session completes with no visible output.
- Change API final-delivery suppression so live Slack delivery only suppresses fallback delivery when there is actual result text and Slack streamed enough visible answer characters to cover it.
- Add regression coverage for both sides:
  - Slackbot renderer does not create blank streams and keeps status behavior sane.
  - API final-delivery fallback is queued when live delivery streamed zero answer characters.

## Relevant Issues Fixed

- Long Slack runs can appear silent/dead after early streamed output because `Thinking...` disappears too soon.
- No-output or failed/no-input executions can skip final fallback delivery when live Slack delivery reported zero streamed answer characters.
- Blank Slack streams can be opened during session close even when there is no visible content to render.

## Test Plan

- `bun test src/*.test.ts src/**/*.test.ts` in `services/slackbot`
- `bun run check:types` in `services/slackbot`
- `/Users/fin-sa/centaur/services/api/.venv/bin/python -m pytest tests/test_agent_control_plane.py -k "slackbot_streamed_answer_chars or live_answer"` in `services/api`
- `/Users/fin-sa/centaur/services/api/.venv/bin/python -m ruff check api/runtime_control.py tests/test_agent_control_plane.py` in `services/api`

### DIFF
--- a/services/api/api/runtime_control.py
+++ b/services/api/api/runtime_control.py
@@ -1039,7 +1039,7 @@ def _slackbot_streamed_answer_chars(value: Any) -> int:
 def _slackbot_live_delivery_covers_result(result_text: str, streamed_chars: int) -> bool:
     text = result_text.strip()
     if not text:
-        return True
+        return False
     return streamed_chars >= len(text)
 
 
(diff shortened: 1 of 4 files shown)


Geeknerd1337:
This fixes https://github.com/paradigmxyz/centaur/issues/358

Zygimantass:

PR #414, opened 2026-06-05
fix(slackbot): stop splicing stale residue into streamed replies

lyoungblood:
## Symptom

Slackbot live replies occasionally arrive with garbled, misformatted residue spliced into an otherwise-correct answer — a misaligned fragment of text jammed in, then the real content. Each turn has its own `CodexSessionState`, so it isn't cross-turn buffer reuse; it's an offset bug in the live-reply assembler.

Two related sites both assumed the answer text only grows by **pure append**, while `recomposeBuffers` can **rewrite** already-streamed content (an `item.completed` replaces an item's accumulated deltas with reformatted canonical text, or a canonical `assistant` event). Slack's `appendStream` accumulates and **cannot be retracted**, so any misaligned text becomes permanent.

## Fix 1 — live streaming (`codex-session.ts`)

`publishPendingAssistantText` streamed `answerText.slice(streamedAnswerText.length)`. When the recomposed answer no longer **extends** what was already streamed, that slice is misaligned. Guard it: if `answerText` no longer `startsWith` `streamedAnswerText`, the stream has diverged — **freeze the answer stream** (logging `slack_codex_answer_stream_diverged` once) instead of appending misaligned text. The coherent already-streamed text stands.

## Fix 2 — long-answer finalize path (`agent-session.ts`)

For **long** answers (streamed past the 30k live cap), `closeTextStream` appends only the *unstreamed remainder* as the final composed block (Slack already shows the live chunks). `finalMarkdownForFinalBlocks` computed that remainder, but fell back to `markdown.slice(segment.streamedTextSourceChars)` when the streamed text wasn't a prefix of the canonical answer — a char-count slice that's **misaligned on divergence**, splicing garbled residue into the final block.

`unstreamedMarkdownAfterLivePrefix` now always returns the answer beyond the **longest common (normalized) prefix** with the streamed text: the exact unstreamed tail when the streamed text is a prefix, and a clean continuation (no misaligned slice) when the model reformatted already-streamed content.

Both fixes are Slack-semantics-agnostic — they simply never emit content that can't be cleanly appended to the accumulating stream. (A "re-render the canonical at finalize" approach was rejected: `stopStream` blocks are additive to the `appendStream` chunks — the long-answer path slices off the streamed prefix precisely to avoid re-sending it — so re-rendering would duplicate, not replace.)

## Tests (`*.test.ts`)

- **divergence (streaming):** a divergent `item.completed` no longer leaks a misaligned canonical fragment into the delivered message; the normal append-only path still streams live.
- **finalize remainder:** `unstreamedMarkdownAfterLivePrefix` returns the exact tail for an (optionally whitespace-normalized) prefix, empty when fully covered, and a clean longest-common-prefix continuation on divergence (no mid-token residue).

```
bun test src/*.test.ts src/**/*.test.ts   # 103 pass, 0 fail
bunx tsc --noEmit                          # clean
```

### DIFF
--- a/services/slackbot/src/slack/agent-session.test.ts
+++ b/services/slackbot/src/slack/agent-session.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from 'bun:test'
-import { AgentSessionRenderer, withAgentSessionLock } from './agent-session'
+import {
+  AgentSessionRenderer,
+  unstreamedMarkdownAfterLivePrefix,
+  withAgentSessionLock
+} from './agent-session'
 
 describe('AgentSessionRenderer', () => {
   it('stops calling assistant.threads.setStatus after the channel returns user_not_found', async () => {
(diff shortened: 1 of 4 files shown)


lyoungblood:
⚠️ **Converted to draft — the streaming guard regressed in production.**

Deploying this revealed that `answerText` is **not monotonic**: during recompose it transiently shrinks then grows (observed live: `streamed_answer_chars: 101 → canonical_answer_chars: 67 → result 2869`). The Fix-1 guard (`!answerText.startsWith(streamedAnswerText)` → freeze the stream) treats that transient shrink as divergence and **freezes**, never resuming — so long replies were truncated to the pre-divergence fragment (~101 of 2869 chars). That's worse than the original occasional-garble.

Root issue is deeper than an offset bug: the assembler streams *provisional* answer content to Slack's append-only `appendStream`, and the canonical answer can diverge from / replace it — which an additive stream can't cleanly reconcile. Reworking toward either (a) replacing the message at finalize via `chat.update`, or (b) not streaming answer text live and rendering it once from canonical at `stopStream`. Needs maintainer input on Slack streaming semantics + a non-prod Slack test before redeploy.

lyoungblood:
Superseded by the option-B PR above: deliver the answer once at finalize from canonical text, instead of guarding the incremental-streaming math (which regressed by truncating long replies). Closing.

PR #417, opened 2026-06-05
fix(slackbot): deliver the agent answer once at finalize (not incrementally)

lyoungblood:
## Symptom

Slackbot live replies were consistently mangled: the **first ~150 characters dropped and back-filled with a duplicated later slice**. (An earlier intermittent variant of the same root cause was explored in #414.)

## Root cause

The reply assembler streams provisional answer deltas live in `CodexSessionRenderer.publishPendingAssistantText` via `answerText.slice(streamedAnswerText.length)`, assuming the answer only ever grows by append. But `recomposeBuffers` can **reorder/reformat already-streamed content** (a canonical `item.completed`, multi-item recompose). When the recomposed answer no longer matches what was already streamed, the stale slice offset drops the real opening and leaves the already-streamed provisional text (a later slice) at the top. Slack's `appendStream` is append-only, so the bad content can't be retracted.

## Fix

Deliver the answer **once, at turn finalize**, from the canonical text — never stream it incrementally. Answer streaming is gated on `opts.force`, which only the finalize calls pass; at finalize `streamedAnswerText` is still empty, so the delta is the whole final answer streamed once → no offset arithmetic, no divergence possible. Activity / plan / thinking still stream live.

```ts
if (state.commentaryText.length > state.streamedCommentaryText.length) return
if (!opts.force) return   // answer is delivered once at finalize, from canonical text
if (state.answerText.length <= state.streamedAnswerText.length) return
const delta = state.answerText.slice(state.streamedAnswerText.length)
```

**Trade-off:** the answer appears when the turn completes rather than token-by-token. Chosen over the alternative (`chat.update`-replacing the message at finalize) as the simpler, lower-risk option that doesn't depend on Slack `stopStream`/`update` replace-vs-append semantics.

## Tests

- A divergent `item.completed` no longer drops the opening or duplicates a later slice.
- The answer is not streamed mid-turn even once `hasPlan`/grace would have allowed it (updated the mid-turn-ordering test to match).
- `bun test src/*.test.ts src/**/*.test.ts` → **98 pass, 0 fail**; `bunx tsc --noEmit` → clean.

## Supersedes #414

#414 tried to fix this inside the incremental-streaming math (freeze the stream on divergence). That regressed in production: `answerText` is non-monotonic — it transiently shrinks then grows during recompose — so the guard froze the stream and **truncated** long replies to the pre-divergence fragment. Option B avoids the incremental path entirely.

### DIFF
--- a/services/slackbot/src/slack/codex-session.test.ts
+++ b/services/slackbot/src/slack/codex-session.test.ts
@@ -747,7 +747,9 @@ describe('CodexSessionRenderer', () => {
     )
     expect(firstTaskIndex).toBeGreaterThanOrEqual(0)
     expect(contextBlockIndex).toBe(-1)
-    expect(firstTextIndex).toBeGreaterThan(firstTaskIndex)
+    // The answer is delivered once at finalize (not streamed incrementally), so the activity task
+    // has streamed but no answer text has streamed yet before the terminal event.
+    expect(firstTextIndex).toBe(-1)
     expect(thinkingBlockText(calls)).toBe('')
 
     await renderer.event(sessionId, { type: 'turn.completed', result: 'Done.' })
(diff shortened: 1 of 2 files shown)


gakonst:
nice thanks!

PR #423, opened 2026-06-05
feat(chart,api): allow resource requests/limits for API-managed pods

### LINKED ISSUE #420: API-managed pods (iron-proxy, tool-server, workflow-run) can't set resource requests/limits
## Summary

Several pods the API creates at runtime have no way to set Kubernetes resource requests/limits. The chart exposes `resources` for `api`, `postgres`, `slackbot`, `tokenBroker`, `repoCache`, `ironControl`, and the `sandbox` container, but **not** for the iron-proxy Pods (the API self-proxy and every per-sandbox proxy), the `tool-server` sidecar, or the workflow-run pod. The container specs in `services/api/api/sandbox/kubernetes.py` are hardcoded with no `resources` field and no env knob, so these pods run with no scheduler reservation and no OOM/throttle ceiling (the workflow-run pod only incidentally inherits the sandbox sizing).

## Steps to reproduce

1. Deploy Centaur with `toolServer.enabled=true` and the kubernetes sandbox backend.
2. Inspect container resources, e.g. `kubectl get pod <pod> -o jsonpath='{.spec.containers[*].resources}'`:
   - `centaur-api-proxy-*` and per-sandbox `*-proxy-*` (container `iron-proxy`) → `{}`
   - sandbox `tool-server` sidecar → `{}`
   - workflow-run pods → inherit the sandbox sizing, not independently configurable
3. Contrast the `sandbox` container, which is sized via `sandbox.resources` → `KUBERNETES_SANDBOX_*` env → `_pod_resources()`.

## Root cause

`_build_proxy_pod_spec`, `_build_tool_server_container`, and `_build_workflow_run_pod_spec` build their containers without a configurable `resources`, and the chart's `ironProxy`/`toolServer` have no `resources` keys. There is no env knob analogous to the sandbox's `_pod_resources()` for these containers, so they cannot be sized and the workflow-run pod can only inherit the sandbox's values.

## Expected outcome

- Each API-managed pod — the API self-proxy, every per-sandbox proxy, the tool-server sidecar, and the workflow-run pod — can be sized independently via chart values.
- Leaving the values unset preserves current behavior: the proxies and tool-server stay unconstrained, and the workflow-run pod keeps its prior sandbox-equivalent sizing.
- Partial specs work (e.g. memory request==limit, no CPU request).

pkobielak:
## Summary

The pods the API creates at runtime — the iron-proxy Pods (the API self-proxy and every per-sandbox proxy), the `tool-server` sidecar, and the workflow-run pod — had no way to set Kubernetes resource requests/limits. The chart exposed no knob and the container specs in `services/api/api/sandbox/kubernetes.py` were hardcoded, so these pods ran with no scheduler reservation or OOM/throttle ceiling (the workflow-run pod only incidentally inherited the sandbox sizing).

This wires per-pod resources through the existing sandbox pattern (chart values → API env → Pod spec), sizing each API-managed container independently:

- **chart values**: `ironProxy.apiResources`, `ironProxy.sandboxResources`, `toolServer.resources`, `workflowRun.resources`.
- **workloads template**: emit `KUBERNETES_{API_PROXY,SANDBOX_PROXY,TOOL_SERVER,WORKFLOW_RUN}_*` resource env from the api Deployment, guarded so unset keys emit nothing.
- **api**: `_resources_from_env()` (no defaults) for the proxies and tool-server; `_resources_from_env_with_default_limits()` for workloads that historically ran with implicit cpu=2/memory=4Gi limits (the sandbox and workflow-run pods). The shared proxy spec selects API vs per-sandbox values by sandbox id.

Behavior is preserved when values are unset: the proxies and tool-server stay unconstrained, and the workflow-run pod keeps its prior sandbox-equivalent sizing.

Fixes #420

## Testing

- `helm template`: each knob renders its `KUBERNETES_*` env independently; partial specs work; unset emits no resource env (and the workflow-run default reproduces the prior sandbox-equivalent sizing).
- `ruff check`: clean.
- `pytest tests/test_sandbox_kubernetes_backend.py`: new full/partial/empty tests for all four knobs pass; existing sandbox/`_pod_resources` tests unchanged.

### DIFF
--- a/contrib/chart/templates/workloads.yaml
+++ b/contrib/chart/templates/workloads.yaml
@@ -309,6 +309,46 @@ spec:
               value: {{ .Values.ironProxy.service.managementPort | quote }}
             - name: KUBERNETES_IRON_PROXY_HEALTH_PORT
               value: {{ .Values.ironProxy.service.healthPort | quote }}
+{{- with .Values.ironProxy.apiResources.limits }}
+{{- if .cpu }}
+            - name: KUBERNETES_API_PROXY_CPU_LIMIT
+              value: {{ .cpu | quote }}
+{{- end }}
+{{- if .memory }}
+            - name: KUBERNETES_API_PROXY_MEMORY_LIMIT
+              value: {{ .memory | quote }}
+{{- end }}
+{{- end }}
+{{- with .Values.ironProxy.apiResources.requests }}
+{{- if .cpu }}
+            - name: KUBERNETES_API_PROXY_CPU_REQUEST
+              value: {{ .cpu | quote }}
+{{- end }}
+{{- if .memory }}
+            - name: KUBERNETES_API_PROXY_MEMORY_REQUEST
+              value: {{ .memory | quote }}
+{{- end }}
+{{- end }}
+{{- with .Values.ironProxy.sandboxResources.limits }}
+{{- if .cpu }}
+            - name: KUBERNETES_SANDBOX_PROXY_CPU_LIMIT
+              value: {{ .cpu | quote }}
+{{- end }}
+{{- if .memory }}
+            - name: KUBERNETES_SANDBOX_PROXY_MEMORY_LIMIT
+              value: {{ .memory | quote }}
+{{- end }}
+{{- end }}
+{{- with .Values.ironProxy.sandboxResources.requests }}
+{{- if .cpu }}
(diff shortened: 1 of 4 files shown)


Zygimantass:
we deprecated the old API in favor of the Rust one, please re-open the PR if there's still interest:)

PR #441, opened 2026-06-07
fix: keep final Slack answers in stopStream layout

snario:
Fixes a Slack rendering bug where the final `chat.stopStream` layout could omit the answer and leave only activity visible.

When the session has an explicit final answer, render that answer in the durable final block layout. After `stopStream`, update the stream message to that final layout so Slack keeps one authoritative copy of the answer instead of leaving stale live chunks beside the final blocks.

Verification:
- `bun test src/slack/agent-session.test.ts src/slack/codex-session.test.ts`


Zygimantass:

PR #442, opened 2026-06-08
Publish multi-arch images and expose pod scheduling values

ecejas:
## Summary

This makes Centaur friendlier to mixed-architecture Kubernetes clusters.

Today the published Centaur images are built by the default GitHub-hosted runner architecture, so the GHCR tags resolve to linux/amd64 manifests only. In clusters with both amd64 and arm64 nodes, chart-managed pods and API-created runtime pods can land on an incompatible node.

Downstream users can try to work around that by pinning Centaur workloads to amd64 nodes, but that is not a clean or complete solution. Some of the relevant pods are not ordinary chart workloads: the API creates sandbox, iron-proxy, and workflow-run pods at runtime. Pinning those from downstream means carrying chart forks, post-render patches, or extra environment plumbing that mirrors Centaur internals. That workaround is brittle, easy to miss during upgrades, and pushes cluster-scheduling behavior out of the upstream chart/API boundary where operators expect to configure it.

This PR addresses that from both sides:

- publish Centaur images for `linux/amd64` and `linux/arm64` using Buildx + QEMU
- expose standard Helm scheduling values (`nodeSelector`, `affinity`, `tolerations`) for chart-managed workloads
- pass scheduling values through the API for runtime-created `iron-proxy`, sandbox, and workflow-run pods
- validate the runtime scheduling env as JSON before building pod specs

## Details

The publish workflow now configures QEMU before Buildx and sets:

```yaml
platforms: linux/amd64,linux/arm64
```

for the image matrix. That should make these GHCR images publish as multi-platform indexes on future main/tag releases:

- `centaur-api`
- `centaur-slackbot`
- `centaur-agent`
- `centaur-iron-proxy`

The chart now supports scheduling values for:

- `api`
- `slackbot`
- `postgres`
- `tokenBroker`
- `ironControl`
- `ironProxy`
- `sandbox`
- `workflowRun`

For normal chart workloads, those values render directly into the pod template. For runtime-created pods, the chart serializes the values into API env vars:

- `KUBERNETES_IRON_PROXY_NODE_SELECTOR`, `KUBERNETES_IRON_PROXY_AFFINITY`, `KUBERNETES_IRON_PROXY_TOLERATIONS`
- `KUBERNETES_SANDBOX_NODE_SELECTOR`, `KUBERNETES_SANDBOX_AFFINITY`, `KUBERNETES_SANDBOX_TOLERATIONS`
- `KUBERNETES_WORKFLOW_RUN_NODE_SELECTOR`, `KUBERNETES_WORKFLOW_RUN_AFFINITY`, `KUBERNETES_WORKFLOW_RUN_TOLERATIONS`

The Kubernetes backend parses those as JSON and applies them to the generated pod specs. This keeps runtime-created pods configurable through the chart, instead of requiring downstream template forks or post-render patches.

## Testing

- `helm lint contrib/chart`
- `helm template centaur contrib/chart --namespace centaur --set tokenBroker.enabled=true --set ironControl.enabled=true ...` with amd64 `nodeSelector` values for chart workloads and runtime pod values
- `python3 -m json.tool contrib/chart/values.schema.json`
- `uv run ruff format api/sandbox/kubernetes.py tests/test_sandbox_kubernetes_backend.py`
- `uv run ruff check api/sandbox/kubernetes.py tests/test_sandbox_kubernetes_backend.py`
- `uv run pytest tests/test_sandbox_kubernetes_backend.py` (`52 passed`)

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -61,10 +61,14 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
 
       - name: Log in to GHCR
+        if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
         uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
         with:
           registry: ${{ env.REGISTRY }}
@@ -98,7 +102,8 @@ jobs:
           context: .
           file: ${{ matrix.dockerfile }}
           target: ${{ matrix.target }}
-          push: ${{ !github.event.pull_request.head.repo.fork }}
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha,scope=${{ matrix.image }}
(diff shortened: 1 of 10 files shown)


Zygimantass:
closed in favor of #525

PR #445, opened 2026-06-08
fix(sandbox): drop foreign-thread events in codex app-server wrapper

### LINKED ISSUE #444: Codex wrapper relays background-thread events, leaking the memories agent's reply and prematurely ending the user's turn
# Codex wrapper relays background-thread events, leaking the memories agent's reply and prematurely ending the user's turn

## Summary

`services/sandbox/codex-app-wrapper.py` forwards **every** notification from `codex app-server` to the Centaur NDJSON stream. But the app-server multiplexes background work onto that same channel — most notably the **memories consolidation agent, which runs in its own thread**. Because the wrapper does not scope notifications to the turn/thread it started, a background agent's output reaches the user and its terminal event ends the user's turn.

## Impact (observed in production)

During a normal task the memories agent woke up in its own thread and, having no user task, emitted a final answer:

> "What do you want me to work on in `~/.codex/memories`? I only have the environment context right now, not a task."

Two things went wrong:

1. **Wrong reply surfaced** — that background message was relayed and shown to the user as if it were the agent's answer.
2. **Real answer dropped** — the background turn's `turn/completed` was also relayed; downstream treats the first `turn/completed` as terminal, so the session closed early and the actual result (which arrived seconds later) was discarded.

The user's task had in fact completed successfully; they just never saw it.

## Root cause

The app-server emits notifications for **all** concurrent threads/turns on one stream, each stamped with its own `threadId`/`turnId`. The wrapper relayed them without checking ownership, so background-thread events were indistinguishable from the user turn's events.

## How we caught it

> **Note on the logs below:** these are real wrapper-emitted NDJSON events from one affected session. All identifiers are replaced with synthetic values — primary turn ids use an `…aaaa…` family, the background (memories) thread an `…bbbb…` family — and user content / task text / repo / URLs are redacted as `<…>`. Event types and field names are unchanged.

The session starts normally; note the app-server's child-process subsystem is already unhealthy at startup (relevant — the memories agent runs as background child work):

```jsonl
{"type":"system","subtype":"wrapper_heartbeat","phase":"startup"}
{"type":"system","subtype":"wrapper_heartbeat","phase":"app_server_started"}
ERROR codex_models_manager::manager: failed to refresh available models: timeout waiting for child process to exit
{"type":"thread.started","thread_id":"0192aaaa-0000-7000-8000-00000000a000"}
{"type":"turn.started","turn_id":"0192aaaa-0000-7000-8000-00000000a001"}
```

The primary turn streams its work, with each event carrying the **primary** `threadId`/`turnId`:

```jsonl
{"type":"item.agentMessage.delta","threadId":"0192aaaa-0000-7000-8000-00000000a000","turnId":"0192aaaa-0000-7000-8000-00000000a001","itemId":"msg_primaryA","delta":"<streamed task commentary>","session_id":"0192aaaa-0000-7000-8000-00000000a000"}
```

Mid-turn, the **memories agent** starts its own thread and turn on the same stream — a different `threadId`/`turnId` — and, with no task, emits a final answer that gets relayed to the user:

```jsonl
{"type":"thread.started","thread_id":"0192bbbb-0000-7000-8000-00000000b000"}
{"type":"turn.started","turn_id":"0192bbbb-0000-7000-8000-00000000b001"}
{"type":"item.agentMessage.delta","threadId":"0192bbbb-0000-7000-8000-00000000b000","turnId":"0192bbbb-0000-7000-8000-00000000b001","itemId":"msg_memB","delta":"What do you want me to work on in `~/.codex/memories`? I only have the environment context right now, not a task.","session_id":"0192bbbb-0000-7000-8000-00000000b000"}
{"type":"item.completed","item":{"type":"agentMessage","id":"msg_memB","text":"What do you want me to work on in `~/.codex/memories`? I only have the environment context right now, not a task.","phase":"final_answer"}}
{"type":"turn.completed","turn":{"id":"0192bbbb-0000-7000-8000-00000000b001","status":"completed"}}
```

That background `turn.completed` is the first terminal event on the stream — downstream treats it as the end of the user's turn. The primary turn's **real** answer arrives afterwards and is dropped because the session is already closed:

```jsonl
{"type":"item.completed","item":{"type":"agentMessage","id":"msg_primaryA","text":"<final result: branch/PR link + validation summary>","phase":"final_answer"}}
{"type":"turn.completed","turn":{"id":"0192aaaa-0000-7000-8000-00000000a001","status":"completed"}}
```

Summarised by turn boundaries (arrows added), the interleaving is clear — the memories turn opens and closes *inside* the primary turn's lifetime:

```
== TURN START  0192aaaa…a001   (thread 0192aaaa…a000)
       msg  msg_primaryA  [commentary]    <agent working on the user's task>
== TURN START  0192bbbb…b001   (thread 0192bbbb…b000)   <-- background agent, DIFFERENT thread
       msg  msg_memB      [final_answer]  "What do you want me to work on in ~/.codex/memories? ..."
== TURN END    0192bbbb…b001   <-- relayed; terminates the user's durable turn early
       msg  msg_primaryA  [final_answer]  <the real result>   <-- dropped: session already closed
== TURN END    0192aaaa…a001
```

Mapping every streamed agent message to its `threadId`/`turnId` confirms the discriminator is the **thread** — one background thread, two of its own turns, distinct from the user's:

```
itemId          threadId          turnId
msg_primaryA    0192aaaa…a000     0192aaaa…a001
msg_memB        0192bbbb…b000     0192bbbb…b001
msg_memB        0192bbbb…b000     0192bbbb…b002
```

## Environment

- Engine: `codex` via `codex app-server` (`codex-app-wrapper.py`)
- Codex memories feature active (background consolidation agent writing to `~/.codex/memories`)

## Fix

PR: filter notifications by `threadId` — the wrapper's own thread id is known from the `thread/start`/`thread/resume` response before any turn runs, so events on a different thread are foreign and are suppressed. Scoped intentionally to the confirmed cause; see the PR for why turn-id-level machinery was deliberately left out.

## Related observations — intentionally NOT addressed in the PR (for maintainer discussion)

We kept the PR scoped to the bug we actually reproduced. We noticed two adjacent things while investigating, but we **only caught the present issue because it surfaced through the memories agent**, and we'd rather discuss with maintainers whether the wrapper should try to detect/cover these before adding speculative code:

1. **`error` notifications ignore `willRetry`.** The wrapper turns any `error` into a terminal `turn.failed`. `ErrorNotification` carries `willRetry`; a retryable error would still end the turn. Whether the wrapper should honour `willRetry` (and how) is worth a design call.
2. **`uuid4()` thread-id fallback as identity.** When `thread/start` returns no id, the wrapper fabricates a uuid that is also used as the emitted `session_id` and the `turn/start` `threadId`. The PR makes the new thread filter fail open in that case (so it can't drop the primary turn), but it does not change the fabricated identity flowing into routing/session — that predates this change and may deserve its own fix.

Happy to follow up on either if maintainers want them covered.

akandic47:
## What

Scope `codex app-server` notifications to the thread the wrapper started, so background-thread work (notably the memories consolidation agent) can no longer reach the user or end the user's turn.

Fixes #444

## Problem

`codex-app-wrapper.py` relayed every notification from the app-server. The app-server multiplexes concurrent threads/turns onto one stream, and the **memories agent runs in its own thread**. In production this caused:

- the memories agent's "What do you want me to work on in `~/.codex/memories`?" final answer to be shown to the user as the reply, and
- its `turn/completed` to be relayed, which terminated the user's durable turn early and **dropped the real answer**.

See #444 for the full write-up and the (sanitized) log evidence showing the background work on a distinct `threadId`.

## Fix

- `_is_foreign_thread_event(params)` — suppress any notification whose `threadId` differs from ours. Our thread id is known from the `thread/start`/`thread/resume` response before any turn is drained, so the check has no blind window.
- `thread/started` only adopts the **first** thread id; a later foreign `thread/started` cannot clobber ours.
- **Authoritative-id guard.** Making `THREAD_ID` a filter key means a wrong value would drop the primary turn's own events. The existing `thread.get("id") or resume or uuid.uuid4()` fallback can produce an id that matches nothing if the server returns none. We now track whether the id is authoritative (from the server or a resume id) and **fail open** (relay everything) when it is not. Events that omit `threadId` also fail open, for protocol compatibility.

## Why thread-only (scope rationale)

We confirmed from production logs that the discriminator is the **thread** — the memories agent's events carry a different `threadId`, and the thread filter alone suppresses all of them, including the premature terminal. We deliberately did **not** add turn-id-level machinery (per-turn id tracking, stale-terminal guards, etc.): in the reproduced incident those guards would have fired on nothing, and we preferred not to add code for failure modes we have not observed. Happy to revisit if maintainers know of same-thread background-turn cases.

## Testing

`services/api/tests/test_codex_app_wrapper.py` (20 passing), including regressions that fail without this change:

- foreign-thread suppression across the turn lifecycle (`turn/started`, `item/agentMessage/delta`, `item/completed`, `turn/completed`)
- foreign-thread `error` suppressed (does not terminate the user's turn)
- primary-thread events relayed normally; process-global `error` (no `threadId`) still terminal
- non-authoritative `THREAD_ID` (uuid fallback) → filter fails open, primary turn's events relayed
- `start_or_resume_thread` marks server/resume ids authoritative and the uuid fallback non-authoritative

## Out of scope — for maintainer discussion (not in this PR)

To keep this PR focused on the reproduced bug, two adjacent observations are **not** addressed here. We only caught this issue because it surfaced via the memories agent, so before adding speculative handling we'd like your read on whether the wrapper should cover them:

1. **`error` ignores `willRetry`** — any `error` becomes a terminal `turn.failed`, even when `ErrorNotification.willRetry` is true.
2. **`uuid4()` thread-id fallback as identity** — the fabricated id also flows into the emitted `session_id` and `turn/start`'s `threadId`; this PR only makes the new filter fail open in that case.

Both are noted in #444.

### DIFF
--- a/services/api/tests/test_codex_app_wrapper.py
+++ b/services/api/tests/test_codex_app_wrapper.py
@@ -344,6 +344,270 @@ def test_emit_notification_collects_agent_message_delta_output(monkeypatch) -> N
     assert emitted[0]["type"] == "item.agentMessage.delta"
 
 
+def test_emit_notification_suppresses_foreign_thread_turn(monkeypatch) -> None:
+    """A background turn on another thread (the memories agent) must not reach
+    the stream or terminate the durable turn.
+
+    Regression test for the incident: the memories consolidation agent ran in
+    its own thread (confirmed from prod pod logs) and its "What do you want me
+    to work on in /home/agent/.codex/memories?" final answer was relayed to
+    Slack while its turn/completed prematurely ended the user's turn, dropping
+    the real answer.
+    """
+    wrapper = _load_wrapper()
+    emitted: list[dict] = []
+    monkeypatch.setattr(wrapper, "emit", emitted.append)
+
+    wrapper.THREAD_ID = "thread-main"
+    wrapper.THREAD_ID_AUTHORITATIVE = True
+    wrapper.ACTIVE_TURN_ID = "turn-main"
+    wrapper.CURRENT_LLM_OUTPUT_TEXT = ""
+    wrapper.LLM_OUTPUTS_BY_TURN_ID = {}
+
+    foreign = {"threadId": "thread-mem", "turnId": "turn-mem"}
+    assert (
+        wrapper.emit_notification({"method": "turn/started", "params": {**foreign}})
+        is False
+    )
+    assert (
+        wrapper.emit_notification(
+            {
+                "method": "item/agentMessage/delta",
+                "params": {
+                    **foreign,
+                    "itemId": "msg-mem",
+                    "delta": "What do you want me to work on",
+                },
+            }
+        )
+        is False
+    )
+    assert (
+        wrapper.emit_notification(
+            {
+                "method": "item/completed",
+                "params": {
+                    **foreign,
+                    "item": {
+                        "id": "msg-mem",
+                        "type": "agentMessage",
+                        "text": "I only have the environment context, not a task.",
+                        "phase": "final_answer",
+                    },
+                },
+            }
+        )
+        is False
+    )
+    done = wrapper.emit_notification(
+        {"method": "turn/completed", "params": {**foreign, "turn": {"id": "turn-mem"}}}
+    )
+
+    assert done is False
+    assert emitted == []
+    assert wrapper.ACTIVE_TURN_ID == "turn-main"
+    assert wrapper.CURRENT_LLM_OUTPUT_TEXT == ""
+    assert wrapper.LLM_OUTPUTS_BY_TURN_ID == {}
+
+
+def test_emit_notification_relays_primary_thread_turn(monkeypatch) -> None:
+    wrapper = _load_wrapper()
+    emitted: list[dict] = []
+    monkeypatch.setattr(wrapper, "emit", emitted.append)
+
+    wrapper.THREAD_ID = "thread-main"
+    wrapper.THREAD_ID_AUTHORITATIVE = True
+    wrapper.ACTIVE_TURN_ID = "turn-main"
+    wrapper.CURRENT_LLM_OUTPUT_TEXT = ""
+    wrapper.LLM_OUTPUTS_BY_TURN_ID = {}
+
+    wrapper.emit_notification(
+        {
+            "method": "item/agentMessage/delta",
+            "params": {
+                "threadId": "thread-main",
+                "turnId": "turn-main",
+                "itemId": "msg-1",
+                "delta": "PONG",
+            },
+        }
+    )
+    done = wrapper.emit_notification(
+        {
+            "method": "turn/completed",
+            "params": {"threadId": "thread-main", "turn": {"id": "turn-main"}},
+        }
+    )
+
+    assert done is True
+    assert [event["type"] for event in emitted] == [
+        "item.agentMessage.delta",
+        "turn.completed",
+    ]
+    assert wrapper.CURRENT_LLM_OUTPUT_TEXT == "PONG"
+    assert wrapper.ACTIVE_TURN_ID is None
+
+
+def test_emit_notification_suppresses_foreign_thread_error(monkeypatch) -> None:
+    """`error` is the protocol's turn-failure path and carries threadId, so a
+    background turn's error must not terminate the durable turn."""
+    wrapper = _load_wrapper()
+    emitted: list[dict] = []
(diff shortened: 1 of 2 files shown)


gakonst:
this code has been replaced by the harness server!

PR #447, opened 2026-06-08
fix: post Slack sends to other channels instead of capturing into active thread

mslipper:
The live-delivery capture in `_capture_live_slack_send` was swallowing `slack.send_message` calls that targeted a *different* channel and folding them into the active thread reply. The guard for "is this bound for the active thread?" only bailed when the channel was a Slack ID that differed, so a different channel referenced by name (or with `thread_ts` omitted) was wrongly captured — the agent would report it had posted nothing to the requested channel.

This inverts the logic to opt-in: capture only when the request is unambiguously bound for the active thread (empty/matching channel and empty/matching thread). Since the active channel from the thread_key is always a Slack ID, a channel name now correctly fails the match and posts normally. Adds `TestCaptureLiveSlackSend` covering the active-thread, different-ID, different-name, and different-thread cases.


Zygimantass:

PR #450, opened 2026-06-09
feat: add opencode as a selectable harness

0xdiid:
Adds opencode as a harness option alongside the existing ones. It runs as its own sandbox process that translates between Centaur's turn protocol and opencode's headless server, so conversations can be driven by opencode the same way they already can by the others.

The model provider and model are fully configurable, and provider credentials are attached the same way they are for the other harnesses — at the proxy boundary, so no real keys ever live inside the sandbox.

Worth a smoke test against a live build before rollout, since the message and event shapes were validated against the pinned version's docs rather than a running server.


0xdiid:
closing for now, changing priorities

PR #451, opened 2026-06-09
feat(api-rs): CloudWatch tool aws_auth via iron-control

mslipper:
Reimplements PR #449's AWS auth support in the Rust api-rs / iron-control control plane instead of the Python `services/api`, which api-rs is replacing.

Adds an `aws_auth` credential type to `centaur-iron-control` (new `AwsAuthSecretInput` model, `aws_auth_secrets` upsert client, `aas_` grant wiring) and to the `centaur-perms` tool-config path, so a tool declaring `type = "aws_auth"` in its pyproject gets a re-signing secret registered in iron-control and granted to its role. iron-proxy re-signs SigV4 requests the tool signed with placeholder credentials, so the real read-only IAM keys never reach the sandbox.

Keeps the CloudWatch tool and the iron-proxy SigV4 header allowlist from #449; drops the Python `services/api` changes. AWS_REGION is non-secret and reaches the sandbox via `passthrough_env`.

Supersedes #449.

### DIFF
--- a/contrib/scripts/bootstrap-k8s-secrets.sh
+++ b/contrib/scripts/bootstrap-k8s-secrets.sh
@@ -23,6 +23,12 @@ Optional local-dev admin key:
                                (envFrom centaur-infra-env). Re-run with --force
                                or kubectl patch to rotate.
 
+Optional repo-cache GitHub token:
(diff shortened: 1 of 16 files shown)


0xdiid:
will close mine 🤝

PR #452, opened 2026-06-09
fix(slackbotv2): paginate task-card streams

Zygimantass:
## Summary
- paginate Slackbot v2 task-card progress streams across multiple Slack stream messages after 24 tasks per page
- persist render pagination state in thread state with pages, task routes, and last rendered event id
- post final answers separately from progress streams, splitting long final text into safe Slack-sized chunks
- treat Slack block-size/count errors as render-size failures and degrade oversized progress pages to compact text

## Validation
- pnpm --filter slackbotv2 check:types
- pnpm --filter slackbotv2 test

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -3,7 +3,6 @@ import { randomUUID } from 'node:crypto'
 import { Hono } from 'hono'
 import {
   Chat,
-  StreamingPlan,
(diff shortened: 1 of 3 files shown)


Zygimantass:
Closing in favor of pinning Chat SDK to upstream Slack stream continuation fix in vercel/chat#601.

PR #459, opened 2026-06-09
fix: auto-recover codex threads when the rollout is gone

goksu:
## Problem

Incident: thread `slack:T092R71U6QY:C0B9BLRNPNE:1781020003.963999` (#ai-gtm) got permanently bricked — every message answered with `no rollout found for thread id 019ead10-eee3-74c3-a9fc-aa3e8bc53783` within 1–3s, including explicit "@ai retry".

Root cause: the codex thread id is persisted durably in `sandbox_sessions.agent_thread_id`, but the rollout file it refers to lives only on the sandbox pod's **ephemeral filesystem**. The first turn hit `silence_deadline_exceeded`, which stopped the pod and destroyed the rollout. Every later turn spawned/attached a sandbox with the dead id as `CODEX_CONTINUE_THREAD_ID`, `thread/resume` failed, and the turn ended `failed_permanent`/`harness_error`. Nothing ever clears the broken id (idle-TTL and generation bumps deliberately preserve it for resume), and the "@ai retry" recovery command only re-hydrates the prompt text.

## Fix (two layers)

1. **Wrapper fallback** (`services/sandbox/codex-app-wrapper.py`): if `thread/resume` fails, emit an observable `thread_resume_failed` system event and fall back to `thread/start`, never reusing the dead resume id. The new thread id flows through the existing `thread.started` event and is persisted by the control plane. Future pod replacements self-heal mid-turn with no user-visible error.

2. **Recovery-command reset** (`services/api/api/workflows/slack_thread_turn.py`): recovery commands ("retry", "continue", …) now check the thread's most recent execution; if it ended `failed_permanent`, the session is reset via a shared `_release_and_reset_session()` helper (extracted from the prompt-switch path): assignment released, runtime stopped, `agent_thread_id` cleared. The fresh sandbox spawns with no resume id, gets Slack history backfill, and runs the re-hydrated original ask. Healthy threads saying "retry" keep their live session. Logs `recovery_session_reset`.

Layer 2 is what un-bricks the existing stuck thread immediately on api deploy (the wrapper fix ships in the sandbox image and only applies to newly spawned pods).

## Tests

- `test_codex_app_wrapper.py`: fallback on resume failure (events + fresh id), dead id never reused even when `thread/start` returns no id, successful resume unchanged.
- `test_workflows.py`: recovery turn resets session after `failed_permanent`; no reset after `completed`; ordinary messages never reset.
- `uv run pytest tests/test_codex_app_wrapper.py tests/test_workflows.py`: all pass except the pre-existing `test_slack_thread_turn_attachment_roundtrip_to_agent` failure (needs `KUBERNETES_SECRET_ENV_NAME` locally; fails on clean main too).

### DIFF
--- a/services/api/api/workflows/slack_thread_turn.py
+++ b/services/api/api/workflows/slack_thread_turn.py
@@ -7,8 +7,12 @@ import json
 import re
 from typing import Any
 
+import structlog
+
 from api.runtime_control import ControlPlaneError
 from api.workflow_engine import Delivery, WorkflowContext
+
+log = structlog.get_logger()
 
 WORKFLOW_NAME = "slack_thread_turn"
 
@@ -247,15 +251,14 @@ def _with_prompt_switch_context_note(
     return [{"type": "text", "text": _PROMPT_SWITCH_CONTEXT_NOTE}, *parts]
 
 
-async def _release_for_prompt_switch(
+async def _release_and_reset_session(
     ctx: WorkflowContext,
     *,
     thread_key: str,
-    message_id: str | None,
+    release_id: str,
 ) -> None:
     from api.runtime_control import release_assignment
 
-    release_id = f"prompt-switch:{message_id or ctx.run_id}"
     await release_assignment(
         ctx._pool,
         thread_key=thread_key,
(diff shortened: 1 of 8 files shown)


services/api/api/workflows/slack_thread_turn.py
```diff
@@ -7,9 +7,13 @@
 import re
 from typing import Any
 
+import structlog
```
goksu: Let's avoid introducing a new log pattern here for the time being.

goksu (commented):
How does this change actually solve the original timeout issue in this thread? Wouldn't restart still run into the same root issue of task taking too long? Any way we can update the task timeouts on retries (exponentially but with some sane max limit).

services/api/api/workflows/slack_thread_turn.py
goksu: Removed in d2c33dd2 — dropped the structlog logger and the log call; the reset is still traceable through the `recovery-reset:<message_id>` release_id that flows into the existing release_assignment logging.

goksu:
> How does this change actually solve the original timeout issue in this thread? Wouldn't restart still run into the same root issue of task taking too long?

It doesn't by itself — the PR as reviewed only fixed the *brick* (the impossible resume after the rollout was lost), not the timeout that started the chain. You're right that a retry of the same prompt could hit the same 600s silence ceiling again.

Addressed in d2c33dd2 with exponential timeout escalation on retries:

- `enqueue_execution` counts `silence_deadline_exceeded` failures since the thread's last **completed** execution (other failure reasons like `harness_error` neither count nor reset the streak, so the incident thread's trailing resume failures don't mask the original timeout).
- The silence window doubles per timeout retry: 600s → 1200s → 2400s, capped by new `EXECUTION_SILENCE_TIMEOUT_MAX_S` (default 2400s) and never above the hard deadline (3600s).
- The escalated value is stored on the execution row (`metadata.silence_timeout_s`) and the claim path now honors it via `COALESCE` instead of resetting to the global constant — previously claim would have clobbered any per-execution value.
- A completed turn resets the streak, and the existing `THREAD_FAILURE_LOOP` guard still stops runaway retry loops within its window.

So for the incident thread: "@ai Retry" → recovery reset → fresh sandbox → re-hydrated original ask running with a 1200s silence window.

goksu:
Added in beb14f10: context replay for the silent fallback path.

Context completeness per recovery path now:

- **"@ai Retry" (recovery reset)**: full context on the first turn — the reset clears `last_delivered_id`, the released assignment triggers Slack history backfill (slackbot collects the entire thread via paginated `conversations.replies`, bot replies included), and `_flush_pending` with a NULL cursor replays the whole durable transcript into the fresh codex thread alongside the re-hydrated original ask.
- **Silent wrapper fallback (resume fails mid-turn)**: previously the fresh thread only ever saw post-cursor messages, since `get_or_spawn` preserves the cursor assuming resume succeeds. Now the control plane handles the wrapper's `thread_resume_failed` event by clearing `last_delivered_id` — the in-flight turn still runs with delta input (it was flushed before the resume attempt), but the next flush replays the full durable transcript (current turn's messages get re-sent once, which is benign).

PR #470, opened 2026-06-10
fix(api): respawn unresumable suspended sandboxes and dedupe runtime-failure posts

samuelclay:
## Why

The Watch Agent spammed a customer-investigation thread in #product-eng with `Failed to start the runtime: failed to resume suspended sandbox: ...` on every message (reported by Ari, 2026-06-09). Root cause chain:

1. PR #9 (cost controls) cut `IDLE_TTL_S` from 24h to 15min, so sessions get suspended after any short lull.
2. The deployed sandbox controller is `KUBERNETES_SANDBOX_CONTROLLER=pod`, whose `pause_by_id` falls back to `stop_by_id` (deletes the pod) and whose `resume_by_id` is the base-class **no-op**. A 'suspended' session can never come back.
3. `get_or_spawn` raised on the failed resume and left the DB row in `suspended`, so **every** subsequent message in the thread re-failed and re-posted the error. 39 sessions are currently stuck in `suspended` on samuel-a100.

## What

- `get_or_spawn`: on suspended-resume failure, log and fall through to the gone-container path — preserve `agent_thread_id`/cursors, delete the row, respawn. Stuck rows self-heal on the next message.
- `workflow_engine`: post `Failed to start the runtime` at most once per thread per distinct error (24h in-process TTL); suppressed repeats are logged as `workflow_spawn_failure_notice_suppressed`.
- Test for the dedupe helper.

## Testing

- `uv run pytest tests/test_agent_control_plane.py tests/test_workflows.py` — same pass/fail set as clean `main` (the failures there are pre-existing DB-fixture flakes); new test passes.
- `uv run ruff check` clean on touched files.

FORK.md updated per fork policy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,4 +81,6 @@ jobs:
           bun test \
             src/centaur/handoff.test.ts \
             src/slack/normalize.test.ts \
+            src/slack/home.test.ts \
+            src/index.test.ts \
             src/centaur/final-delivery.test.ts
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,12 @@
 # Centaur — Developer Guide
 
+## Fork tracking (Tavus)
+
+This repo is Tavus's fork of [paradigmxyz/centaur](https://github.com/paradigmxyz/centaur).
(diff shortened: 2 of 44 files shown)


samuelclay:
Opened against the wrong repo (meant for the Tavus fork) — apologies for the noise.

PR #476, opened 2026-06-10
fix(proxy): allow Datadog API-key auth headers

cjustice:
## Summary

Add `dd-api-key` / `dd-application-key` to iron-proxy's `header_allowlist` so agent tools can authenticate to the Datadog REST API through the per-sandbox proxy.

## Why

Datadog authenticates with two headers — `DD-API-KEY` (org API key) and `DD-APPLICATION-KEY` (application key); it does not accept `Authorization: Bearer <api_key>` for static-key auth. Because neither `dd-`-prefixed header is on iron-proxy's `header_allowlist` (and they don't match the `^x-…(api-key|token|auth|key)$` catch-all), the proxy strips them before egress, so a Datadog tool's requests reach `api.datadoghq.com` unauthenticated and 403. This mirrors the existing per-tool header additions (`x-as-user-email`, etc.).

## Rebased / ported

This PR originally also edited the legacy `services/api` Python proxy renderer and its `test_proxy_config.py`. Both were **removed upstream** when the Python `api` service was deleted (#521), so the rebase dropped them. The change now lands solely in `services/iron-proxy/iron-proxy.yaml` — the base proxy config baked into the iron-proxy image (`Dockerfile` → `proxy.yaml.default`) and the only place `header_allowlist` is defined in the repo. The allowlist is now static YAML (no renderer), so the old render-test is obsolete.

### DIFF
--- a/services/iron-proxy/iron-proxy.yaml
+++ b/services/iron-proxy/iron-proxy.yaml
@@ -66,6 +66,8 @@ transforms:
         - "api-access-key"
         - "api-signature"
         - "api-timestamp"
+        - "dd-api-key"
+        - "dd-application-key"
         - "fx-access-key"
         - "fx-access-sign"
         - "fx-access-timestamp"


cjustice:
@mslipper — would you be able to review this when you have a moment? One-line addition to the iron-proxy `header_allowlist` (your area), now rebased onto `main` and reduced to just `services/iron-proxy/iron-proxy.yaml` after the legacy `services/api` renderer was removed. Thanks!

goksu:
@cjustice:

Thank you for the diff!

Unfortunately, closing this one for now as it is orthogonal to our current design of logging via OTEL. 

We actually use Datadog at Tempo and are utilizing existing logs and mirroring them on the infra deployment side ingestion. This is instead of making API calls directly from source hence iron-proxy is not part of the equation. No API requests are done through the Centaur source to any logging provider (i.e. Datadog) and is generic.

Happy to re-open if you strongly disagree! Otherwise also open to walk through how we mirror logs on our Centaur infra for Datadog if desired!

PR #477, opened 2026-06-10
fix: stop silent Slack answer loss - segment budget, pump reattach, file uploads

Zygimantass:
## Context

Six production incidents on 2026-06-10 (prd-centaur-na) traced to three root causes. Five "no response, only thinking" threads: the agent finished its work, but the Slack streamed render outgrew Slack's message cap because task-card bytes were never counted toward the segment budget; #466's `msg_too_long` suppression then cleared the render obligation and the persisted final answer was silently dropped. One thread additionally lost its answer to api-rs's stdout-EOF handling during the back-to-back rollouts. Separately, agents linked local sandbox paths instead of uploading files.

### 1. slackbotv2 — card bytes count toward the stream segment budget (`patches/@chat-adapter__slack@4.30.0.patch`)

- Structured chunks (task cards, plan) contribute an estimated size to their segment's 11500-char budget, with per-card upsert accounting so status updates don't double-count.
- New cards rotate to a fresh stream message when they'd exceed the budget (previously only a 50-card count cap); `stopBlocks` rotate first when they'd tip the final message over.
- With renders no longer able to outgrow Slack's limit, the #466 suppression path stays as a dead-letter guard only.
- Regression test: 60 command cards + final answer → ≥2 stream messages, each ≤11800 accumulated chars, answer delivered. Without the accounting one message accumulates 13,632 chars and the test fails.

### 2. api-rs — transport EOF ≠ harness failure; SIGTERM; attach race (`centaur-session-runtime`, `centaur-api-server`)

- Stdout pump EOF with an active execution now feeds a reattach loop: if the sandbox is still Running, re-open IO and keep pumping (bounded attempts, reset on progress, `session.stdout_pump_reattached` event). Only a gone/terminal sandbox or exhausted retries fail the execution — with the status detail in the error text. Previously the first EOF failed the execution and flipped the session to `failed`; in incident thread `slack:C0A87C21805:1781034315.961439` the full answer arrived 107ms after that premature failure.
- Per-sandbox open locks serialize `open_io` across execute/SSE/steering/reattach — the unserialized check-then-act race produced 3 competing k8s attaches that disrupted each other and caused those EOFs. Pipe-map cleanup is now pointer-guarded so a dying pump can't clobber a newer pipe.
- `shutdown_signal()` handles SIGTERM (k8s' stop signal); as PID 1 it was previously ignored, so every rollout ended in an abrupt SIGKILL.
- Two PG-backed integration tests (gated on `SESSION_RUNTIME_TEST_DATABASE_URL`, mirroring absurd-sdk) cover reattach-delivers-late-terminal-output and sandbox-gone-fails-with-detail.

### 3. sandbox — deliver files via `slack-upload`, never local-path links (`SYSTEM_PROMPT.md`, `codex-app-wrapper.py`)

- The codex harness base instructions tell the model to render clickable local-file links; nothing overrode that for chat, and the existing `slack-upload` script was dead in the Rust stack because `SLACK_CHANNEL`/`SLACK_THREAD_TS` were never set.
- The wrapper now derives both from the `thread_key` already on every input line (api-rs and legacy key shapes), before the app server first starts so tool subprocesses inherit them.
- SYSTEM_PROMPT: deliver files with `slack-upload`; never reference sandbox paths or `file://` URIs in replies (explicitly overrides harness clickable-link instructions); fixed the garbled "[Slack responses]" line with an upload carve-out.

## Validation

- slackbotv2: `bun test test` 24/24; types clean; negative check confirms the new test fails without the fix
- api-rs: `cargo check`, `clippy`, `fmt`, and `cargo test --workspace` green, incl. the 2 new PG-backed tests run against a local scratch DB
- sandbox: `pytest tests/test_codex_app_wrapper.py` 16/16 (2 new), `tests/test_assemble_prompt.py` green
- Known limitation: k8s attach cannot replay output emitted during a detached gap; the race fix + graceful drain prevent the gap from opening in the observed scenarios

## Deploy

Fix 1 ships with the slackbotv2 image, fix 2 with api-rs, fix 3 with the centaur-agent image (`SESSION_SANDBOX_IMAGE`).

Incident investigation: https://ampcode.com/threads/T-019eb342-f946-750e-89d1-a8f028e80d0e

### DIFF
--- a/patches/@chat-adapter__slack@4.30.0.patch
+++ b/patches/@chat-adapter__slack@4.30.0.patch
@@ -1,8 +1,8 @@
 diff --git a/dist/index.js b/dist/index.js
-index 53a35e40d7dc03dea57f965325728761467ad7c7..3e63db3d328688a104cf408ff97f8e984b25ba93 100644
+index 53a35e40d7dc03dea57f965325728761467ad7c7..f9f5a201e661a5234fd5cd479d8475ca9bfacccd 100644
 --- a/dist/index.js
 +++ b/dist/index.js
-@@ -31,6 +31,123 @@ import {
+@@ -31,6 +31,148 @@ import {
    toModalElement,
    toPlainText
  } from "chat";
@@ -11,6 +11,8 @@ index 53a35e40d7dc03dea57f965325728761467ad7c7..3e63db3d328688a104cf408ff97f8e98
 +var STREAM_CHUNK_LIMIT = 256;
 +var STREAM_TASK_LIMIT = 50;
 +var STREAM_FENCE_RESERVE = 64;
++var STREAM_CARD_OVERHEAD = 64;
++var STREAM_PLAN_SIZE_KEY = "\0plan";
 +var FENCE_PATTERN = /^(`{3,}|~{3,})/;
 +var Fence = class {
 +  constructor() {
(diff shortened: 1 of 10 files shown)


Zygimantass:
Closing this combined PR in favor of smaller follow-ups. The remaining file-delivery fix has been split to #507. The Slack answer-loss portion is superseded by merged #505, and the api-rs rollout/stdout-loss work overlaps with merged #486 for recovery/adoption.

PR #487, opened 2026-06-11
fix(slackbot): route final delivery to Slack Connect team

decofe:
## Summary
- Pass preserved Slack Connect recipient team into final-delivery fallback/aggregate chat.postMessage calls
- Keep normal Slack posts unchanged when no recipient team is present
- Add regression coverage for recipient_team_id on final delivery posts

## Test
- bun test src/centaur/final-delivery.test.ts
- bun test src/*.test.ts src/**/*.test.ts


Zygimantass:

PR #498, opened 2026-06-11
fix(sandbox): report stopped when the agent container terminates so executions don't reclaim-loop

cjustice:
## Problem

When a sandbox's agent container is OOMKilled (or otherwise terminates) mid-turn, the execution gets stuck in an endless reclaim loop until the hard deadline and the user never gets a response.

## Root cause

`KubernetesExecutorBackend.status_by_id` derives session status from the **pod phase** alone. The phase stays `Running` as long as *any* container runs — and each sandbox pod has a per-sandbox `tool-server` sidecar that survives the agent container's death. So after the `sandbox` container is OOMKilled (exit 137), `status_by_id` still returns `running`.

The execution stream consumer (`runtime_control`) then sees the dead container's stdout stream EOF, checks `backend.status(...)`, gets `running`, and sets the execution back to `retry_wait` expecting a reattach to recover. But every reattach to the dead container's stdout EOFs immediately, so it loops — `execute_claimed` roughly every 60s — until the hard-deadline reaper terminalizes it. No failure is ever surfaced to the user.

Observed in a deployed GKE cluster: a memory-heavy build OOM-killed the sandbox container; the attach stream closed (`{"status":"Success"}`) at the exact instant of the OOM kill; `podPhase=Running` (sidecar still alive) → reclaim loop for the entire deadline window.

## Fix

`status_by_id` now inspects the agent (`sandbox`) container's own state. If that container has terminated, it returns `stopped` even when the pod phase is `Running`, so the stream-EOF path finalizes the execution (failed) instead of reclaim-looping on a dead container's stdout.

## Testing

- Added `test_status_by_id_reports_stopped_when_agent_container_terminated` — OOMKilled agent container + `Running` pod phase → `stopped`; live agent container → `running`. Passing (`uv run pytest tests/test_sandbox_kubernetes_backend.py`); `ruff check` clean.
- Root cause reproduced live (a real monorepo build OOM-killed the sandbox container at its memory limit); this change finalizes such executions cleanly instead of looping.

### DIFF
--- a/services/api/api/sandbox/kubernetes.py
+++ b/services/api/api/sandbox/kubernetes.py
@@ -53,6 +53,23 @@ _READY_TIMEOUT_S = int(os.getenv("KUBERNETES_SANDBOX_READY_TIMEOUT_S", "60"))
 _ATTACH_LOG_TAIL_LINES = int(os.getenv("KUBERNETES_ATTACH_LOG_TAIL_LINES", "200"))
 _CONTAINER_NAME = "sandbox"
 _AGENT_UID = 1001
+
+
+def _agent_container_terminated(pod: Any) -> bool:
+    """True when the sandbox (agent) container has terminated.
+
+    The pod's ``phase`` stays ``Running`` as long as any container is alive, so
+    a crashed/OOMKilled agent container is hidden by the surviving per-sandbox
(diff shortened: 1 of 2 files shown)


Zygimantass:

PR #519, opened 2026-06-12
fix(linear): surface mutation success flag from create_issue/update_issue/add_comment

lyoungblood:
## What

`LinearClient.create_issue`, `update_issue`, and `add_comment` now return the mutation's `success` flag merged into the returned entity dict (`{"success": <bool>, **entity_fields}`), via a shared `_mutation_result` helper. The CLI `create`/`update`/`comment` commands check the flag (shared `require_mutation_success` helper) and exit 1 with an error instead of printing a green success line for a failed mutation.

## Why

The GraphQL mutations already requested Linear's `success` field, but the client stripped it and returned only the inner `issue`/`comment` dict. A mutation that fails with `success: false` but no GraphQL error was therefore indistinguishable from success. Callers that detect failure by checking `result.get("success") is False` — e.g. workflow helpers wrapping the linear tool's mutations — silently treated those failures as success. (`add_label`/`remove_label` already returned `{"success": ...}`; this brings the entity-returning mutations in line.)

## Notes for reviewers

- **Return shape**: the flag is merged flat into the entity dict rather than nested (`{"success": ..., "issue": {...}}`) because existing callers read entity fields directly off the result (`result.get("identifier")` etc.); the flat merge is the only shape that breaks no one. On failure the methods return `{"success": False}`.
- The CLI `comment` command's old `if result:` truthiness check is updated in the same change — a failed comment now returns a truthy `{"success": False}` dict, so truthiness alone would no longer catch it.
- `create_issue_relation` is untouched: it already returns the raw payload including `success`, and its caller navigates that nested shape.
- Tests: existing mutation test now asserts the flag passes through; new test covers the failure path for all three methods (`uv run pytest tests/test_linear_integrations.py` in `services/api`, 8 passed).

### DIFF
--- a/tools/productivity/linear/cli.py
+++ b/tools/productivity/linear/cli.py
@@ -33,6 +33,13 @@ def get_client():
     return LinearClient()
 
 
+def require_mutation_success(result: dict, action: str) -> None:
+    """Exit with an error if a mutation result carries success=False."""
+    if not result.get("success"):
+        console.print(f"[red]Linear reported the {action} failed.[/]")
+        raise typer.Exit(1)
+
+
 @app.command()
 def me():
     """Show authenticated user info."""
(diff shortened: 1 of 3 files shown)


Zygimantass:
lgtm

PR #529, opened 2026-06-12
fix(sandbox): restore codex as the default harness

cjustice:
## Problem
The sandbox image `CMD` and the chart's `sandbox.harnessEngine` both default to **claude-code**, carried in from the api-rs migration (#344). Both lines carry a NOTE:

> `claude-code is a preview-463 dogfood default — flip back to codex (together with …) before merging.`

…but it shipped to `main` un-reverted.

Since the harness engine is fixed by the image `CMD` (`harness-server codex|claude-code|amp`) with no chart/api-rs override, this makes claude-code the de-facto default for any deployment tracking the chart at `main`. A deployment provisioned for codex (`OPENAI_API_KEY` via the proxy) then runs claude-code with no Anthropic credential and fails with **`Not logged in · Please run /login`**. codex is also the native protocol (Codex App Server V2) and `harness-server`'s own no-arg default.

## Change
Restore the documented-intended default:
- `services/sandbox/Dockerfile`: `CMD ["harness-server", "codex"]`
- `contrib/chart/values.yaml`: `sandbox.harnessEngine: codex`

Operators who want to dogfood claude-code can still set `sandbox.harnessEngine` (and the image CMD).

## Question
Was shipping claude-code as the `main` default intentional (an ongoing dogfood), or an oversight per the NOTE? If intentional, feel free to close — happy to defer. Either way the comment seemed worth surfacing.

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.52
+version: 0.1.53
 appVersion: "0.1.0"
 dependencies:
   - name: connect
--- a/contrib/chart/values.yaml
+++ b/contrib/chart/values.yaml
@@ -166,9 +166,7 @@ sandbox:
   # Which harness the sandbox runs: codex | amp | claudecode. Selects the
   # iron-proxy auth fragment api-rs registers and must agree with the sandbox
   # image's default CMD (services/sandbox/Dockerfile).
-  # NOTE: claudecode is a preview-463 dogfood default — flip back to codex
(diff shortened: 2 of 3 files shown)


cjustice:
cc @Zygimantass @gakonst @goksu — flagging for review since this touches the harness default you all worked on: @Zygimantass set both defaults to claude-code in #344, and the `preview-463` the NOTE references is @gakonst / @goksu's #463. This restores codex per the in-code "flip back to codex before merging" comment. If keeping claude-code as the `main` default is intentional for the ongoing dogfood, feel free to close — just wanted to surface it.

goksu (approved):
We should be default Codex — thank you for catching it!

cjustice:
@goksu can you reapprove? I had to push a change to fix a failing check

goksu (approved):

PR #531, opened 2026-06-12
fix(harness-server): export SLACK_CHANNEL/SLACK_THREAD_TS for codex tools

cjustice:
## Problem

`harness-server codex` (blocks mode, the default) never sets `SLACK_CHANNEL`/`SLACK_THREAD_TS`, so `slack-upload` aborts with `SLACK_CHANNEL not set` and the agent can't deliver files back into the Slack thread.

The Python `codex-app-wrapper` derived these from the per-turn `thread_key` *before starting the app server* (`export_slack_thread_env`). That was dropped when the Rust harness replaced the wrapper in #344. The Rust bridge also spawns `codex app-server` **eagerly**, before reading any input — so the child (and the tools it spawns, which inherit its env) can't pick up vars set mid-stream.

## Fix (`crates/harness-server/src/codex.rs`)

- Defer the `codex app-server` spawn + `initialize` to the first user turn.
- Before that spawn, parse `thread_key` from the blocks line and export `SLACK_CHANNEL`/`SLACK_THREAD_TS` so the child and its tool subprocesses inherit them. Mirrors `export_slack_thread_env`; supports `slack:<channel>:<thread_ts>` (api-rs) and legacy `slack:<team>:<channel>:<thread_ts>`; no-op for non-Slack threads.

`env::set_var` runs before the child/reader threads exist (single-threaded), so it's sound. Deferring `initialize` matches the wrapper's original ordering and is invisible to api-rs — no harness output flows until the first turn.

## Scope

Only the codex blocks path. The `claude-code` path (`run_blocks_app_server`) has the same gap — separate generic codepath, left for follow-up.

## Testing

- Unit tests for `slack_channel_and_thread` / `thread_key_from_blocks_line` (arity, legacy 4-part keys, non-Slack, empty/blank/absent/invalid).
- Integration test: a fake `codex app-server` records its inherited env; asserts `SLACK_CHANNEL`/`SLACK_THREAD_TS` are set from `thread_key "slack:C123:123.456"`. Confirmed to **fail without the fix**.
- `cargo test` (37 lib + 8 integration), `cargo fmt --check`, `cargo clippy` all clean. K8s E2E not run — change is confined to the harness binary.

### DIFF
--- a/crates/harness-server/src/codex.rs
+++ b/crates/harness-server/src/codex.rs
@@ -66,26 +66,15 @@ impl AppServerRuntime for CodexHarnessServer {
 }
 
 pub(crate) fn run_codex_blocks_server() -> Result<()> {
-    let mut codex = CodexJsonRpcChild::spawn()?;
     let mut stdout = io::stdout().lock();
     let mut request_id = 1_i64;
     let mut thread_id: Option<String> = None;
     let mut blocks_state = BlocksState::default();
-
-    let initialize_id = next_request_id(&mut request_id);
-    codex.send_request(
-        initialize_id,
-        "initialize",
-        json!({
-            "clientInfo": {
-                "name": "centaur-harness-server",
-                "title": null,
-                "version": env!("CARGO_PKG_VERSION"),
-            },
-            "capabilities": null,
-        }),
-    )?;
-    codex.read_response_or_forward(initialize_id, &mut stdout)?;
+    // The codex app-server child is spawned lazily on the first turn so the
+    // Slack thread env (below) is exported *before* the child exists — the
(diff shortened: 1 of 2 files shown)


cjustice:
@akshaan — mind taking a look when convenient? Small harness-server fix to export `SLACK_CHANNEL`/`SLACK_THREAD_TS` for codex tools, in your Slack-thread area and rebased on latest `main`. Thanks!

goksu:
Similar approach landed with https://github.com/paradigmxyz/centaur/pull/633. Closing this as @akshaan  is also actively looking into improving attachment flows with the latest API that is less dependent on env variables.

PR #532, opened 2026-06-12
fix(api-rs): deliver overlay.systemPrompt to sandboxes

cjustice:
## Problem
`overlay.systemPrompt` is a no-op on the Rust control plane. The chart renders it into a ConfigMap (`overlay-configmap.yaml`) and sets `CENTAUR_OVERLAY_DIR` + a `checksum/overlay` annotation on api-rs — but the ConfigMap is **mounted nowhere** (api-rs's only volume is `repo-cache`) and **api-rs never consumes it** (no overlay reader anywhere in `services/api-rs`). So the org overlay never reaches the sandbox: the agent runs on the bare baked base prompt and silently ignores the operator's overlay.

Confirmed live: in the sandbox, `~/workspace/AGENTS.md` is the base prompt only, `CENTAUR_OVERLAY_DIR` is unset, there is no `~/AGENTS_OVERLAY.md`, and `/app/overlay/org` doesn't exist in the api-rs pod — though the `…-overlay` ConfigMap holds the prompt.

## Fix
The legacy Python control plane delivered the prompt via a ConfigMap `subPath` mount (`services/api/api/sandbox/{prompt_assembly,kubernetes}.py`); the Rust rewrite never ported it. This wires it through minimally, reusing the sandbox entrypoint's **existing** `~/AGENTS_OVERLAY.md` contract (`entrypoint.sh` already appends that file to the effective `AGENTS.md` — no entrypoint change):

- **centaur-sandbox-core**: add `MountKind::ConfigMap { name }` + optional `Mount::sub_path` (project a single ConfigMap key as one file).
- **centaur-sandbox-agent-k8s**: render the `configMap` volume (`optional: true`) + `volumeMount.subPath`.
- **centaur-api-server**: add `--overlay-configmap` / `KUBERNETES_OVERLAY_CONFIGMAP`; when set, the codex sandbox mounts that ConfigMap's `SYSTEM_PROMPT.md` at `~/AGENTS_OVERLAY.md`, read-only.
- **chart** (`apirs.yaml`): pass the overlay ConfigMap name to api-rs when `overlay.systemPrompt` is set.

The volume is `optional: true`, so a missing/misnamed ConfigMap degrades gracefully (the entrypoint already guards `~/AGENTS_OVERLAY.md` with `[ -f ]`) rather than wedging the pod.

## Scope
Delivers the overlay **system prompt** only. The `[Active deployment]` block + persona injection that the Python `prompt_assembly` also produced are a related, separate gap left for follow-up — happy to adjust shape if you'd prefer it unified.

## Verification
- `cargo test -p centaur-sandbox-core -p centaur-sandbox-agent-k8s -p centaur-api-server` — green, incl. new tests: `mount_json_renders_configmap_subpath`, `agent_sandbox_crd_round_trips_configmap_overlay_mount`, `codex_workload_mounts_overlay_configmap_as_agents_overlay`, `codex_workload_has_no_overlay_mount_when_unset`.
- `helm lint` + `helm template contrib/chart -f values.dev.yaml` — pass; api-rs renders `KUBERNETES_OVERLAY_CONFIGMAP`.

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.55
+version: 0.1.56
 appVersion: "0.1.0"
 dependencies:
   - name: connect
--- a/contrib/chart/templates/apirs.yaml
+++ b/contrib/chart/templates/apirs.yaml
@@ -133,6 +133,10 @@ spec:
 {{- if .Values.overlay.systemPrompt }}
             - name: CENTAUR_OVERLAY_DIR
               value: {{ .Values.overlay.mountPath | quote }}
+            # api-rs mounts this ConfigMap's SYSTEM_PROMPT.md into each sandbox as
+            # ~/AGENTS_OVERLAY.md (the entrypoint appends it to AGENTS.md).
+            - name: KUBERNETES_OVERLAY_CONFIGMAP
+              value: {{ include "centaur.componentName" (dict "root" . "component" "overlay") | quote }}
 {{- end }}
             - name: SESSION_SANDBOX_BACKEND
               value: {{ .Values.apiRs.sandboxBackend | quote }}
--- a/services/api-rs/crates/centaur-api-server/src/args.rs
+++ b/services/api-rs/crates/centaur-api-server/src/args.rs
@@ -2364,7 +2364,10 @@ mod tests {
             .find(|mount| mount.target_path == SANDBOX_OVERLAY_PROMPT_PATH)
             .expect("overlay prompt mount present");
         assert!(overlay.read_only);
-        assert_eq!(overlay.sub_path.as_deref(), Some(OVERLAY_CONFIGMAP_PROMPT_KEY));
+        assert_eq!(
+            overlay.sub_path.as_deref(),
+            Some(OVERLAY_CONFIGMAP_PROMPT_KEY)
+        );
         assert_eq!(
             overlay.kind,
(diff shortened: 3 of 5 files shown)


cjustice:
@Zygimantass — would you be able to review when you have a moment? It touches the api-rs overlay delivery you've worked on (#443), and it's rebased on latest `main`. Thanks!

gakonst:
Closing this because the repo-backed overlay path now covers the intended prompt delivery contract.

With the repo-cache overlay setup, sandboxes get the overlay repo mounted/available and  already appends  into the workspace . That makes  -> ConfigMap ->  a legacy compatibility path rather than the canonical mechanism.

Keeping both paths would add another source of truth for org prompts and make prompt provenance harder to debug. If we need ConfigMap-based prompt delivery later, we should reintroduce it explicitly as a compatibility fallback with clear precedence against repo-backed overlays.

gakonst:
Closing this because the repo-backed overlay path now covers the intended prompt delivery contract.

With the repo-cache overlay setup, sandboxes get the overlay repo mounted/available and `entrypoint.sh` already appends `${CENTAUR_OVERLAY_DIR}/services/sandbox/SYSTEM_PROMPT.md` into the workspace `AGENTS.md`. That makes `overlay.systemPrompt` -> ConfigMap -> `~/AGENTS_OVERLAY.md` a legacy compatibility path rather than the canonical mechanism.

Keeping both paths would add another source of truth for org prompts and make prompt provenance harder to debug. If we need ConfigMap-based prompt delivery later, we should reintroduce it explicitly as a compatibility fallback with clear precedence against repo-backed overlays.

PR #539, opened 2026-06-12
REV-16: document shared 1Password quota handling

ethanhasclaws:
Fixes REV-16

## Summary
- raise the chart-level `ironProxy.secretTtl` default from `10m` to `1h`
- document that 1Password service-account rate limits are account-wide across service accounts
- add an operator runbook for the proxy `secret_unavailable` / `rate limit exceeded` -> `Invalid or missing API key` failure signature and recovery steps

## Validation
- `bash scripts/build-md.sh`
- `git diff --check`
- `rg -n "secretTtl: 1h|onepassword-quota|account-wide|rate limit exceeded|Invalid or missing API key" contrib/chart/values.yaml docs/pages docs/public/md`

## Notes
- `helm` is not installed in this sandbox, so I validated the chart/default wiring by source inspection and grep instead of `helm lint`/`helm template`.

### DIFF
--- a/contrib/chart/values.yaml
+++ b/contrib/chart/values.yaml
@@ -29,7 +29,10 @@ ironProxy:
     # one port so callers can reach it.
     pgPort: 5432
   secretSource: onepassword
-  secretTtl: 10m
+  # Deployment default. The binary fallback remains 10m, but a longer chart
+  # value keeps steady-state 1Password reads proportional to live proxies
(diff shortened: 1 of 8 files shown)


ethanhasclaws:
@coderabbitai review

Zygimantass:
hi, we prefer that people who run into these ratelimits use 1Password Connect instead of the raw API - we'd prefer not to increase the TTL to avoid staleness.

PR #541, opened 2026-06-12
feat: Linear chat ingress (linearbot)

0xdiid:
Adds Linear as a chat ingress for centaur. Like discordbot, it's a deliberate clone of slackbotv2's session model — same lifecycle against the api-rs control plane — re-shaped to Linear's comment threads: a comment thread maps to one sandbox/context, and the bot answers *in the thread* with a single, live-edited comment. Every deliberate divergence is commented in-code at its site, and an emulate-style end-to-end harness (ported from slackbotv2's) keeps the two in lockstep.

Linear's native **agent sessions** are deliberately **off** — making the app an agent turns an `@`-mention into a session widget that consumes the comment, destroying the interactive thread the product wants. So work is triggered off plain `Comment` and `Issue` webhooks, which also gives Linear two surfaces Slack and Discord don't have: assigning or delegating an issue to the bot makes it *own* the work (it drives the issue's workflow status), and plain comments in a thread it's already active in fold in as context for the next turn.

## How a mention flows

```mermaid
sequenceDiagram
    participant U as User
    participant L as Linear
    participant B as linearbot
    participant A as api-rs

    U->>L: @centaur in a comment
    L->>B: Comment webhook (HMAC-signed)
    B->>B: verify signature + detect mention (fail closed)
    B->>L: 👀 on the triggering comment
    B->>L: post live comment in the thread (keyed linear:{issue}:c:{comment})
    B->>A: create session, prepend issue context, execute
    Note over B,L: handoff awaited before the webhook is acked — a retryable<br/>failure answers 503 so Linear redelivers
    A-->>B: event stream
    B->>L: reasoning as a headline over a collapsed Thinking… section (throttled edits)
    B->>L: same comment swapped in place to the answer over a collapsed Chain of thought
    B->>L: 👀 → ✅ (or ❌)
```

## Testing

In addition to the automated unit suites and the emulate harness (fake Linear GraphQL API + mock api-rs + signed webhooks driving the real patched adapter end to end), this has been running live against real issues.


mslipper:
Something wrong with the image publish job. Linearbot is missing from the `service` array, and needs to be added to the merge matrix so images actually get built.

services/linearbot/src/index.ts
```diff
@@ -0,0 +1,1271 @@
...
+
+/**
+ * Comment-thread responder (primary). A comment that @-mentions the bot is
+ * answered as one visible comment in its thread — the answer with the
+ * chain-of-thought in a collapsed section — running on the thread's sandbox
+ * (1 thread === 1 context stack). Returns null when the webhook is not a
+ * bot-mention comment. Fires whether or not a session was also created, so it
+ * works before and after agent-session events are turned off.
+ */
+function handleCommentMention(
```
goksu, line 456: Looks like thread follow ups has a gate around blocking self-mentions of bots. Potentially you may want the same here.

```
if (input.botUserId && event.authorId === input.botUserId) return null;
```

services/linearbot/src/overrides.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+// Claude model aliases, usable both as bare flags (--opus) and as --model
+// values (--model opus). Bare-flag form also implies the claude-code harness.
+const CLAUDE_MODEL_ALIASES: Record<string, string> = {
+  fable: "claude-fable-5",
+  haiku: "claude-haiku-4-5",
+  opus: "claude-opus-4-8",
+  sonnet: "claude-sonnet-4-6",
+};
```
goksu, line 37: Wonder if there is a way for us to have these shared for all bots. I think we had to the same thing for Slackbot before.

goksu (commented):
My comments are non-blocking. 

Overall I think we are good once @mslipper comment is addressed.

0xdiid:
@mslipper added, the existing jobs are failing on a nonexistent token, maybe because I'm an external contributor?

services/linearbot/src/index.ts
0xdiid: done

services/linearbot/src/overrides.ts
0xdiid: Yeah there's definitely a way, I can look into it later if you want but not sure it's the job of this PR

goksu (approved):
Feel free to land! Thank you.

PR #543, opened 2026-06-12
feat(sandbox): make codex reasoning summaries configurable via env

tarrencev:
## Why

The codex CLI bump to 0.139 (#…recent releases) changed behavior: codex no longer emits reasoning summaries unless `model_reasoning_summary` is set in its config. Since `harness/codex/config.toml` doesn't set it, the app-server now emits reasoning items with empty `summary`/`content`, and downstream renderers (e.g. the Slack adapter, which reads the reasoning item text) silently lost the embedded thinking trace.

Reproduce on codex 0.139.0:

```
codex exec --json --skip-git-repo-check -c model_reasoning_effort='"low"' "What is 17 * 23 - 19?"
# -> no reasoning item, but usage shows reasoning_output_tokens > 0

codex exec --json --skip-git-repo-check -c model_reasoning_effort='"low"' -c model_reasoning_summary='"detailed"' "What is 17 * 23 - 19?"
# -> item.completed {type: reasoning, text: "**Calculating...**"}
```

## What

Make it configurable per deployment instead of baking a default into the image: the sandbox entrypoint merges `CODEX_MODEL_REASONING_SUMMARY` (`auto` | `concise` | `detailed` | `none`) into the generated `~/.codex/config.toml` — replacing an existing top-level `model_reasoning_summary` or inserting one before the first table. Invalid values warn to stderr and are ignored so a typo can't break every sandbox boot.

Deployments opt in through the existing `sandbox.extraEnv` chart value; no chart changes needed.

Tested the merge against `harness/codex/config.toml` for all four cases (unset, insert, replace-existing, invalid) — output stays valid TOML and the existing `[features]` rewrite is unaffected.

### DIFF
--- a/docs/pages/reference/configuration.mdx
+++ b/docs/pages/reference/configuration.mdx
@@ -164,6 +164,7 @@ Sandbox entrypoint and wrappers:
 | `GOOGLE_APPLICATION_CREDENTIALS` | Sandbox entrypoint or `sandbox.extraEnv`. | Google ADC path; entrypoint creates a local stub when unset. |
 | `CODEX_API_KEY`, `CODEX_HOME`, `CODEX_CONTINUE_THREAD_ID` | `sandbox.extraEnv` or runtime resume. | Codex auth/config/resume behavior. |
 | `CODEX_AUTH_MODE` | `sandbox.extraEnv`. | Codex auth flow: `api_key` (default, hits `api.openai.com`) or `access_token` (hits `chatgpt.com` via the brokered ChatGPT login). See [Codex Auth Modes](/deploying-in-production#codex-auth-modes). |
+| `CODEX_MODEL_REASONING_SUMMARY` | `sandbox.extraEnv`. | Sets `model_reasoning_summary` in the Codex config (`auto`, `concise`, `detailed`, `none`). Codex >= 0.139 emits no reasoning summaries unless this is set, so renderers show no thinking trace. |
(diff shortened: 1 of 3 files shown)


tarrencev:
Related in-flight work, flagging to avoid a collision:

- **#552** (`feat/codex-reasoning-effort-env`) adds `CODEX_MODEL_REASONING_EFFORT` via the **same** `entrypoint.sh` inline-Python step that patches the copied `~/.codex/config.toml`. This PR and #552 are siblings touching the same normalization pass and will textually conflict — they'd ideally share one pass (one place that rewrites top-level keys / inserts before the first table), with `model_reasoning_summary` (here) and `model_reasoning_effort` (#552) as two keys.
- **#549** (`feat/codex-rsn-reasoning-flag`) is the per-turn `-rsn` override and is complementary (per-turn on top of the fleet default set here).

Happy to rebase this on top of #552 (or fold both into a single config-patch step) — whichever sequencing the maintainers prefer.

decofe:
Opened a small follow-up PR against the source branch to default `CODEX_MODEL_REASONING_SUMMARY` to `detailed`, preserving the prior visible reasoning-summary behavior when unset: https://github.com/tarrencev/centaur/pull/5

Validation run:
- `bash -n services/sandbox/entrypoint.sh`
- focused `uv run python` config-merge simulation for unset, override, invalid env values

decofe:
Superseded by #597, which combines this change with #552.

PR #551, opened 2026-06-13
restore ETL workflow integrations

akshaan:
## Summary
- restore the deleted read-only clients used by ETL workflows next to their workflow domains
- group Slack ETLs under `workflows/slack/`, Google Workspace ETLs/helpers under `workflows/gsuite/`, and Linear ETL/helpers under `workflows/linear/`
- update workflow discovery to recurse through workflow subdirectories while importing only files that define `WORKFLOW_NAME`
- lazy-import Google/Linear runtime clients so metadata discovery does not require workflow-only HTTP deps
- add the Python package deps these ETLs need in workflow-run sandboxes
- declare the same deps in `services/workflow-python/pyproject.toml`
- remove the now-unneeded integration compatibility fallback from `workflow_host.py`

## Why
The Rust/Absurd workflow host can still run the Python ETL workflows, but the client layer they import was deleted with the legacy Python API cleanup. These helpers now live beside the workflows that use them instead of reviving the old Python API package. Slack also got past the env gate and failed on missing `slack_sdk`, so this includes the missing runtime deps for all ETL workflows.

Restored from pre-removal history (`95698b78^`) and relocated to domain-local imports:
- `workflows.gsuite.drive.GoogleDriveReadonlyClient`
- `workflows.gsuite.calendar.GoogleCalendarReadonlyClient`
- `workflows.linear.LinearReadonlyClient`
- `workflows.slack.shared` for Slack ETL shared code

## Validation
- `git diff --check`
- `python3 -m py_compile` for Slack, GSuite, Linear, company-context workflows, and workflow host
- bare workflow-host discovery with `WORKFLOW_DIRS=/Users/akshaan/Desktop/centaur/workflows`, confirming all ETL workflow names without extra deps
- `uv run --with ...` import smoke for Slack, Google Drive, Google Calendar, Linear, and company-context ETL modules
- `cargo check -p centaur-api-server`
- built local `centaur-api-rs:latest` and `centaur-agent:latest` images
- container smoke: API image discovers nested workflows from `/app/workflows`; agent image imports Slack/Google/Linear deps
- throwaway local Kubernetes deployment in `centaur-pr551` on `orbstack`:
  - API registered enabled schedules from `/app/workflows/{slack,gsuite,linear}/...`
  - workflow-run pod received ETL env passthrough (`SLACK_ETL_ENABLED`, Google/Linear flags, metrics labels, Slack token present)
  - Slack sync ran real workflow code and wrote local DB rows before cancellation
  - Google Drive/Calendar and Linear ran through workflow host and completed as handled ETL failures at external auth because iron-proxy was intentionally disabled for the local smoke
  - local DB showed ETL rows: Slack messages/channels/backfill jobs and one run row each for Drive, Calendar, Linear


akshaan:
+1

PR #549, opened 2026-06-13
feat(slackbotv2): per-turn codex reasoning effort via -rsn flag

cjustice:
## Motivation

Codex reasoning effort is fixed globally by `harness/codex/config.toml` (`model_reasoning_effort`), so every turn pays the same cost regardless of difficulty. A low default keeps routine turns cheap but underserves hard ones; a high default overpays for trivial ones. There's no way to dial it per turn.

## Change

Add a `-rsn <effort>` message flag that overrides reasoning effort for a single turn, mirroring the existing `--model` per-turn override path end to end:

- **`overrides.ts`** — parse and validate `-rsn` (space or `=` form), strip it from the prompt → `MessageOverrides.reasoning`. Unknown values (e.g. `-rsn turbo`) are left untouched rather than forwarded.
- **`index.ts` / `types.ts`** — carry `reasoning` on `ForwardSessionInput` (+ trace log).
- **`session-api.ts`** — thread through `executeSession` → `toCodexInputLines`, emitting a `reasoning` field on the blocks-protocol user line.
- **`crates/harness-server`** — `BlocksLine` / `BlocksCommand::User` parse `reasoning`; the codex runtime sets `turn/start` `effort`. The emulated claude/amp app-server ignores it (no equivalent knob).

Usage (tag the bot):

```
@centaur -rsn high audit this auth flow
@centaur -rsn=medium refactor the parser
@centaur fix this typo            # unchanged — uses the configured default
```

Accepted efforts match codex's `ReasoningEffort` enum, with short aliases:

| value | aliases |
|-------|---------|
| `none` | |
| `minimal` | `min` |
| `low` | |
| `medium` | `med` |
| `high` | `hi` |
| `xhigh` | `xhi`, `x-high` |

## Scope

Reasoning effort only affects the **codex** harness — combining `-rsn` with a claude/amp flag (e.g. `--opus`) silently no-ops, matching how those harnesses already lack a reasoning knob.

## Testing

- `bun test` (slackbotv2) — `overrides.test.ts` 16/16 (6 new `-rsn` cases: space/equals forms, case-insensitivity, aliases, combining with harness flags, unknown + empty values) and `session-api.test.ts` (2 new: `reasoning` lands on / is omitted from the execute blocks line, matching the `model` bar). Full suite: 64 pass, with 5 pre-existing failures in `chat-sdk-emulate.test.ts` (confirmed identical on a clean tree — unrelated Slack stream-sealing).
[...]

PR #552, opened 2026-06-13
feat(sandbox): override codex reasoning effort via CODEX_MODEL_REASONING_EFFORT

cjustice:
## Motivation

The codex `model_reasoning_effort` lives in `harness/codex/config.toml`, which is baked into the agent image and copied into each sandbox at boot (`services/sandbox/entrypoint.sh`). A deployment that runs upstream `main` directly (no image fork) therefore has no way to change the reasoning default — it's static. Orgs that want a different default (e.g. `high` fleet-wide, or a cheaper `low`) currently can't, short of forking.

## Change

Add a deploy-time override. When `CODEX_MODEL_REASONING_EFFORT` is set, the entrypoint patches `model_reasoning_effort` in the copied `~/.codex/config.toml` — reusing the inline-Python step that already normalizes the `[features]` table, so it's one place and one pass:

- Validated against codex's `ReasoningEffort` enum (`none|minimal|low|medium|high|xhigh`); an unknown value is **ignored with a stderr note** rather than written, so a typo can't break codex startup.
- Unset leaves the baked-in config default untouched (no behavior change for existing deployments).
- Rewrites the existing top-level key in place, or inserts it before the first table if absent.

Delivery: this is injectable per-deployment via `sandbox.extraEnv` (→ `SESSION_SANDBOX_EXTRA_ENV` → every sandbox pod), so an org overlay can set the default without a fork:

```yaml
sandbox:
  extraEnv:
    CODEX_MODEL_REASONING_EFFORT: high
```

Documented in `docs/public/md/reference/configuration.md` alongside the other sandbox entrypoint env vars.

## Relationship to the per-turn flag

Complements (doesn't overlap) the per-turn `-rsn` override (#549): this sets the fleet **default**; `-rsn` overrides a **single turn** on top of it.

## Testing

- Extracted the entrypoint's inline-Python patch step and ran it against the real `harness/codex/config.toml`:
  - `CODEX_MODEL_REASONING_EFFORT=high` → `model_reasoning_effort = "high"`; `xhigh` → `"xhigh"`.
  - `=bogus` → unchanged (`"low"`) + stderr `ignoring invalid ...`.
  - unset → unchanged. The existing `[features]` normalization (`multi_agent`/`multi_agent_v2`) still applies in all cases.
- `bash -n services/sandbox/entrypoint.sh` clean.

### DIFF
--- a/docs/public/md/reference/configuration.md
+++ b/docs/public/md/reference/configuration.md
@@ -163,6 +163,7 @@ Sandbox entrypoint and wrappers:
 | `GOOGLE_APPLICATION_CREDENTIALS` | Sandbox entrypoint or `sandbox.extraEnv`. | Google ADC path; entrypoint creates a local stub when unset. |
 | `CODEX_API_KEY`, `CODEX_HOME`, `CODEX_CONTINUE_THREAD_ID` | `sandbox.extraEnv` or runtime resume. | Codex auth/config/resume behavior. |
 | `CODEX_AUTH_MODE` | `sandbox.extraEnv`. | Codex auth flow: `api_key` (default, hits `api.openai.com`) or `access_token` (hits `chatgpt.com` via the brokered ChatGPT login). See [Codex Auth Modes](/deploying-in-production#codex-auth-modes). |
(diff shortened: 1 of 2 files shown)


cjustice:
@Zygimantass — could I get your eyes on this when you get a chance? It's a sandbox entrypoint / codex-config change in your area, rebased on latest `main`. Thanks!

decofe:
Superseded by #597, which combines this change with #543.

PR #557, opened 2026-06-14
Persist session attachments in API

gakonst:
## Summary\n- add a session-scoped attachments table plus store methods\n- materialize incoming session attachment parts with dataBase64 into DB rows and replace them with attachment_ref parts\n- add SDK-compatible /agent/attachments upload/download routes\n\n## Root cause\nThe sandbox prompt and tool SDK expected an attachments DB/API surface, but api-rs had no attachments migration or /agent/attachments routes. Slackbotv2 could inline attachment bytes into session messages, but nothing persisted them as durable DB attachment records.\n\n## Validation\n- cargo fmt --check\n- cargo check -p centaur-api-server\n- cargo test -p centaur-api-server attachment

### DIFF
--- a/services/api-rs/Cargo.lock
+++ b/services/api-rs/Cargo.lock
@@ -457,6 +457,7 @@ name = "centaur-session-runtime"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "base64",
  "centaur-iron-control",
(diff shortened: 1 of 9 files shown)


gakonst:
might be unnecessary now given ETL + attachmetns being materialized

PR #567, opened 2026-06-15
fix(rendering): coalesce streamed reasoning into one Thinking task

tarrencev:
## Problem (regression since the app-server rewrite, #344)

When reasoning summaries are enabled, the Slack/Chat-SDK "Thinking" trace renders as **one task per streamed word** plus a **duplicated block per summary part**, e.g.:

image

Codex ≥ 0.139 emits no reasoning summaries by default, so this path is dormant on a default config — it only surfaces when a deployment opts back in (`model_reasoning_summary = auto|concise|detailed`). That's why it isn't seen day to day, but it's a real bug in the renderer for anyone who enables summaries.

## Root cause

In `packages/rendering/src/codex-app-server.ts`, every event carrying reasoning text minted a **brand-new task**:

```ts
const reasoningMessage = reasoningText(event).trim()
if (reasoningMessage) {
  const task = { id: `reasoning-${++this.state.stepCounter}`, title: 'Thinking', ... }
  this.state.taskByUseId.set(task.id, task)
}
```

`reasoningText()` returns `event.delta` for `item/reasoning/summaryTextDelta` / `textDelta`, and each delta is a tiny fragment (≈ one word). So N deltas → N separate "Thinking" tasks with N different ids. It never accumulates by the reasoning item id (unlike the adjacent *commentary* thinking path, which does). The existing test fed only a **single** delta, so the per-delta id coincidentally equalled the itemId and the explosion was never covered.

It also conflicts with the chat-sdk emission contract: `changedActivityTaskUpdates` emits a task's `details` only once per id (`emittedActivityRunByTaskId`), which assumes a task body is set-once — true for a command string, false for a growing reasoning stream.

## Why this is a regression

The pre-rewrite (Python) normalizer rendered reasoning from the **consolidated item**, not the deltas — `services/api/api/sandbox/normalize.py`:

```python
if item_type == "reasoning" and phase in ("updated", "completed"):
    text = _as_str(item.get("text")) or _as_str(item.get("thinking"))
    return [{"type": "reasoning", "text": text}] if text else []
```

i.e. one coherent block per reasoning item from its full text. The app-server rewrite replaced that with per-delta task creation.

## Fix

Restore the former behavior: accumulate `summaryTextDelta`/`textDelta` by `itemId`, and render a **single** Thinking task from the consolidated reasoning item delivered on `item/completed` (`{ type: "reasoning", id, summary: string[], content: string[] }`), preferring the item's own `summary`/`content` and falling back to the accumulated deltas when the completed item carries no text (finalized at turn end). One Thinking block per reasoning item — which also satisfies the emit-details-once contract the per-delta path violated.

Adds tests for both the consolidated-item path and the deltas-only fallback.

### DIFF
--- a/packages/rendering/src/codex-app-server.test.ts
+++ b/packages/rendering/src/codex-app-server.test.ts
@@ -53,6 +53,34 @@ describe('CodexAppServerRendererEventMapper', () => {
     })
   })
 
+  it('coalesces streamed reasoning summary deltas into a single Thinking task', async () => {
+    // Codex streams a reasoning summary as many small deltas (≈ one word each).
+    // They must accumulate into one Thinking task keyed by the reasoning itemId,
+    // not a fresh task per delta (which rendered the trace one word per line).
+    const chunks = await collect(
+      codexAppServerToChatSdkStream(
+        toAsyncIterable([
+          { method: 'item/reasoning/summaryTextDelta', params: { threadId: 't', turnId: 'turn-1', itemId: 'reasoning-7', summaryIndex: 0, delta: 'Consider' } },
+          { method: 'item/reasoning/summaryTextDelta', params: { threadId: 't', turnId: 'turn-1', itemId: 'reasoning-7', summaryIndex: 0, delta: 'ing the' } },
+          { method: 'item/reasoning/summaryTextDelta', params: { threadId: 't', turnId: 'turn-1', itemId: 'reasoning-7', summaryIndex: 0, delta: ' investor' } },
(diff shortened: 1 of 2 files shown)


tarrencev:
Superseded by #527, which fixes the same per-delta reasoning Thinking-task explosion and goes further: it streams the accumulating thinking text live (re-emitting task details when they change), resolves the Thinking/command flip-flop by keeping the task in_progress until the reasoning item seals, and adds paragraph breaks between summary sections. Closing in favor of #527 to avoid duplicate work.

PR #571, opened 2026-06-15
chore(sandbox): observe instruction-sensitive git actions

decofe:
## Summary
- add non-blocking sandbox wrappers for `git` and `gh`
- log commit and PR-title convention matches when AGENTS.md mentions conventional commits
- warn agents at action time without blocking the command

## Test
- uv run python -m py_compile services/sandbox/instruction-observer.py
- uvx ruff check services/sandbox/instruction-observer.py
- manual observer smoke test with fake git/gh binaries

Prompted by: @goksu


goksu:
This is not something we want to land as it is really over-complicating to find the root cause of issues.

PR #581, opened 2026-06-15
Preserve Slack thread history attachments

gakonst:
## Summary
- preserve Slack `files` from `conversations.replies` history messages instead of converting them with `attachments: []`
- map Slack file metadata into the existing slackbotv2 attachment shape for harness context
- add emulator coverage proving a prior-thread file reaches the harness input when a reply mention refreshes thread context

## Tests
- `bun test test/chat-sdk-emulate.test.ts --test-name-pattern "refreshes Slack thread context"`


gakonst:
fixed https://github.com/paradigmxyz/centaur/commit/18fecd8feec7f5af044593c98e4ef20e707fec2e

PR #590, opened 2026-06-16
Allow sandboxes to reach cluster-local observability backends

Zygimantass:
## Summary
- Add generated per-sandbox NetworkPolicy egress to cluster-local observability pods in the `monitoring` namespace.
- Allow VictoriaMetrics `vmsingle` on `8428`/`8429`, VictoriaLogs `vlsingle` on `9428`, and Loki on `3100`.
- Scope the new peers by namespace and pod labels, and add unit coverage for the generated rules.

## Why
Production overlays inject cluster-local VictoriaMetrics/VictoriaLogs URLs and include those service names in `NO_PROXY`, so sandbox tools bypass `iron-proxy` for those calls. The generated per-sandbox Kubernetes NetworkPolicy only allowed egress to `iron-proxy`, `api-rs`, DNS, and optional OTLP, which caused direct observability calls to time out.

## Validation
- `cargo fmt --package centaur-sandbox-agent-k8s`
- `cargo test -p centaur-sandbox-agent-k8s sandbox_egress_policy_allows`

### DIFF
--- a/services/api-rs/crates/centaur-sandbox-agent-k8s/src/iron_proxy.rs
+++ b/services/api-rs/crates/centaur-sandbox-agent-k8s/src/iron_proxy.rs
@@ -1092,6 +1092,7 @@ fn build_iron_proxy_network_policies(
         ),
         dns_egress_rule(),
     ];
+    sandbox_egress.extend(cluster_local_observability_egress_rules());
     if let Some(target) = otlp_egress {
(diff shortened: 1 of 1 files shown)


Zygimantass:
Superseded by the infra-owned fix in tempoxyz/prd-centaur-infra#406. The sandbox observability egress allowance is being carried as GitOps NetworkPolicy instead of a base Centaur generator change.

PR #598, opened 2026-06-16
feat(sandbox): configure harness config via env overlays

tarrencev:
## What

Adds two deployment-time overlay env vars so an operator can configure the codex and Claude Code harnesses through `sandbox.extraEnv`, without forking the baked config files or maintaining a derived sandbox image:

- **`CODEX_CONFIG_OVERLAY`** — a TOML fragment deep-merged over the baked `~/.codex/config.toml`. The motivating use case is pointing codex at a custom model provider via a `[model_providers.*]` block, e.g.:
  ```toml
  model_provider = "myproxy"
  [model_providers.myproxy]
  name = "openai"
  base_url = "https://proxy.internal/backend-api/codex"
  wire_api = "responses"
  supports_websockets = false
  ```
- **`CLAUDE_SETTINGS_OVERLAY`** — a JSON fragment deep-merged over the baked `~/.claude/settings.json` (symmetric, for the Claude Code harness).

## Why

This generalizes the same "configure the harness from a deployment without forking config" pattern already established by `CODEX_MODEL_REASONING_SUMMARY` / `CODEX_MODEL_REASONING_EFFORT` (#597). Those knobs cover individual keys; an operator running a custom egress/proxy or alternate provider needs to add a whole `[model_providers.*]` table, which today forces either a config fork or a derived image. A deep-merge overlay covers that case with no schema coupling.

## Behavior

- Both are **no-ops when unset**.
- **Invalid input is ignored** (the baked config stands) and logged to stderr, rather than failing the sandbox — same failure posture as the existing reasoning overrides.
- Deep-merge semantics: nested tables/objects merge recursively; scalars and arrays replace. Existing baked keys not named in the overlay are preserved.
- The codex overlay re-serializes via `tomli-w` (added to the sandbox image deps); the claude overlay uses stdlib `json`.

## Notes for review

- The codex overlay runs as the final step of the existing `config.toml` python block, after the reasoning-summary/effort line edits, so it composes with them.
- `tomllib` is stdlib on the sandbox's Python 3.12; only `tomli-w` (writer) is a new dependency.

### DIFF
--- a/services/sandbox/Dockerfile
+++ b/services/sandbox/Dockerfile
@@ -39,6 +39,7 @@ RUN --mount=type=cache,target=/root/.cache/pip,sharing=locked \
     opentelemetry-proto==1.42.1 \
     pysocks>=1.7.1 \
     slack-sdk==3.39.0 \
+    tomli-w==1.2.0 \
     && find /usr/local/lib/python3.12 /usr/lib/python3 -type d -name __pycache__ -prune -exec rm -rf '{}' +
 
 # ── GitHub CLI + Node.js 24 (LTS) + Docker CLI (single layer, cached) ─────────────
--- a/services/sandbox/entrypoint.sh
+++ b/services/sandbox/entrypoint.sh
@@ -220,7 +220,35 @@ if effort:
                 break
         else:
             lines.insert(first_table, override)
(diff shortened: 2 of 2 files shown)


gakonst (approved):
seems fine

PR #608, opened 2026-06-16
feat: tighten ETL context RLS

akshaan:
## Summary
- tighten ETL RLS so Slack channel principals only see Slack rows/docs for their channel unless allowlisted as admin
- keep projected Slack documents visible to their source channel while restricting Google Drive, Google Calendar, and Linear ETL rows/docs to admin channels
- scope live Slack fallback searches to the current Slack channel unless the channel is allowlisted as an ETL admin

## Validation
- `uv run --with ruff ruff check tools/productivity/company_context/client.py tools/productivity/company_context/tests/test_client.py`
- `uv run --with pytest --with asyncpg --with rich pytest tools/productivity/company_context/tests/test_client.py`
- `cargo test -p centaur-api-server postgres_listeners_retain_sandbox_env_name_and_database`

### DIFF
--- a/services/api-rs/crates/centaur-session-sqlx/migrations/0019_etl_context_rls.sql
+++ b/services/api-rs/crates/centaur-session-sqlx/migrations/0019_etl_context_rls.sql
@@ -1,0 +1,234 @@
+create or replace function centaur_current_slack_channel_id()
+returns text
+language sql
+stable
+as $$
+    select nullif(current_setting('centaur.slack_channel_id', true), '')
+$$;
+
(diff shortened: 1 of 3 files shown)


services/api-rs/crates/centaur-session-sqlx/migrations/0019_etl_context_rls.sql
goksu, line 1: If you do end up deciding to land this watch out for the migration order as 0019 I believe is already behind the latest increment.

goksu (commented):
One comment. We should close if no longer relevant. Thank you!

PR #609, opened 2026-06-16
Re-enable sandbox warm pool

gakonst:
## Summary
- Set the Helm chart default `apiRs.sandboxWarmPoolSize` back to `1` so api-rs starts a warm sandbox pool.

## Verification
- `git diff --check`
- Unable to run `helm template`: `helm` is not installed in this sandbox.
- Unable to verify live rollout: `kubectl` has no active cluster context and falls back to localhost:8080, which is refused.

### DIFF
--- a/contrib/chart/values.yaml
+++ b/contrib/chart/values.yaml
@@ -278,7 +278,7 @@ apiRs:
   sandboxBackend: agent-k8s
   sandboxWorkload: codex-app-server
   sandboxReadyTimeoutSecs: 90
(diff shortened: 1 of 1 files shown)


gakonst:
Closing in favor of paradigmxyz/centaur-infra#124, which updates the live Argo values override for production. This PR only changed the chart default.

gakonst:
Reopening: we do want the chart default set to 3 as well. Production override is handled separately in paradigmxyz/centaur-infra#124.

PR #625, opened 2026-06-17
Expose Slack block content to agent context

smartcontracts:
## Summary
- render visible Slack Block Kit text from raw.blocks into session message parts
- include rendered block content in Codex execute input and Slack thread context
- bound rendered block content and avoid forwarding raw block JSON

## Tests
- bun test services/slackbotv2/test/session-api.test.ts
- /Users/k/Projects/work/centaur/services/slackbotv2/node_modules/.bin/tsgo --project tsconfig.json --noEmit


goksu:
Thank you for the diff! We made an effort to move all Slack rendering to Chat SDK as earlier versions of Centaur built on top of Slack Block Kit and was a pain to work consistently. Closing for now. Happy to re-open if there is a specific use case this is enabling. Let me know!

PR #637, opened 2026-06-17
Reduce Docker image build context

gakonst:
## Summary
- Shrink the iron-proxy image build context to services/iron-proxy
- Exclude CI metadata and test directories from root Docker contexts to reduce context uploads and cache churn

## Validation
- Parsed publish-images workflow YAML with PyYAML before dropping the workflow edit blocked by token scope
- Local Docker build could not run because this sandbox has no Docker daemon

## Follow-up
- I also attempted to add pull_request path filters to .github/workflows/publish-images.yml so image builds do not run on unrelated PRs, but GitHub rejected pushing workflow changes from this OAuth token because it lacks workflow scope.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -64,6 +64,8 @@ services/api-rs/target
 logs/
 
 # IDE / editor
+.github
+.github/**
 .claude
 .cursor
 .vscode
@@ -80,6 +82,12 @@ logs/
 !centaur_sdk/README.md
 !services/sandbox/SYSTEM_PROMPT.md
 !.agents/skills/**
+
+# Tests are validated by CI but are not needed in runtime images.
+**/test/
+**/test/**
+**/tests/
+**/tests/**
 
 # Environment (secrets)
 .env
--- a/.dockerignore
+++ b/.dockerignore
@@ -88,6 +88,8 @@ logs/
 **/test/**
 **/tests/
 **/tests/**
+**/examples/
+**/examples/**
 
 # Environment (secrets)
 .env
--- a/services/api-rs/Dockerfile
+++ b/services/api-rs/Dockerfile
@@ -2,7 +2,7 @@
 # Build context is the repo root (see Justfile `_build-api-rs`) so the runtime
 # stage can bake the shared `tools/` tree used for iron-proxy fragment discovery.
 
-FROM rust:1-bookworm AS builder
+FROM lukemathwalker/cargo-chef:latest-rust-1-bookworm AS chef
 # aws-lc-rs (the rustls crypto provider) builds native code, so the builder
 # needs cmake + a C toolchain. No DATABASE_URL is required at build time:
 # centaur-session-sqlx only uses sqlx::migrate! (compile-time file embedding),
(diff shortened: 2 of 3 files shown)


gakonst:
Added an api-rs-specific compile optimization in 076b9b12: the Dockerfile now uses cargo-chef to split dependency compilation from the final application compile. This should make ordinary source-only api-rs changes reuse the dependency layer instead of spending the full compile step rebuilding deps.

Caveat: I could not run a local Docker build in this sandbox because there is no Docker daemon/socket. CI should validate the image build.

gakonst:
Looked at paradigmxyz/reth's Docker build setup and ported the part we can change without workflow-token scope.

What reth does:
- normal Dockerfile: cargo-chef dependency layer
- Dockerfile.depot: installs sccache, sets RUSTC_WRAPPER=sccache and SCCACHE_DIR=/sccache, mounts /sccache during cargo build, and uses Depot WebDAV cache via DEPOT_TOKEN
- docker workflow: depot/bake-action with docker-bake.hcl and DEPOT_TOKEN secret

What I added here in b76a9ba3:
- install sccache in the api-rs builder image
- set RUSTC_WRAPPER=sccache, SCCACHE_DIR=/sccache, CARGO_INCREMENTAL=0
- mount /sccache in both cargo-chef cook and final cargo build
- print sccache stats in CI logs

The remaining reth-style optimization is wiring DEPOT_TOKEN into the image build as a secret for remote sccache. That requires editing .github/workflows/publish-images.yml, which this OAuth token still cannot push because it lacks workflow scope.

gakonst:
Pushed Dockerfile support for Depot-backed sccache in 084a30cb. It now mounts `DEPOT_TOKEN` as `SCCACHE_WEBDAV_TOKEN` and only enables `SCCACHE_WEBDAV_ENDPOINT=https://cache.depot.dev` when the secret is present, so regular Docker builds still work.

The remaining workflow change is blocked by my token lacking `workflow` scope. Apply this in `.github/workflows/publish-images.yml`:

```diff
 permissions:
   contents: read
   packages: write
+  id-token: write

-      - name: Set up Docker Buildx
+      - name: Set up Depot CLI
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: depot/setup-action@15c09a5f77a0840ad4bce955686522a257853461 # v1.7.1
+
+      - name: Set up Docker Buildx
+        if: ${{ github.event.pull_request.head.repo.fork }}
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3

-      - name: Build and push ${{ matrix.image }} (${{ matrix.platform }})
+      - name: Build and push ${{ matrix.image }} (${{ matrix.platform }}) with Depot
         id: build
-        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: depot/build-push-action@98e78adca7817480b8185f474a400b451d74e287 # v1
+        env:
+          DEPOT_TOKEN: ${{ secrets.DEPOT_TOKEN }}
         with:
+          project: ${{ vars.DEPOT_PROJECT_ID }}
           context: ${{ matrix.context }}
           file: ${{ matrix.dockerfile }}
           target: ${{ matrix.target }}
           platforms: ${{ matrix.platform }}
           labels: ${{ steps.meta.outputs.labels }}
+          secrets: |
+            DEPOT_TOKEN=${{ secrets.DEPOT_TOKEN }}
```

Keep the existing Docker fallback build step for fork PRs, because forks will not have Depot secrets.

gakonst:
Update based on “we have Paradigm OSS”: use the existing Paradigm OSS Depot project for `project:`. I could not verify org-level Actions vars from this token (`gh variable list --org paradigmxyz` is 403), and `centaur` has no visible repo-level `DEPOT_PROJECT_ID`/`DEPOT_TOKEN`.

Recommended workflow wiring:
- If org variable exists: keep `project: ${{ vars.DEPOT_PROJECT_ID }}` and point that var at Paradigm OSS.
- If no org variable exists: set repo/org variable `DEPOT_PROJECT_ID` to the Paradigm OSS Depot project ID, or hardcode that ID in the workflow.
- Auth: prefer Depot OIDC with `id-token: write`; keep `DEPOT_TOKEN` only if Paradigm OSS is not configured for GitHub OIDC yet. The Dockerfile already supports `DEPOT_TOKEN` for remote sccache when mounted.

PR #638, opened 2026-06-17
Allow sandboxes readonly Postgres access

gakonst:
## Summary
- add an infra pg_dsn listener that exposes the app database to sandboxes as CENTAUR_DATABASE_URL through iron-proxy
- bind the listener to the existing centaur_readonly role
- cover the infra role wiring in the api-rs args test

Note: main now includes 0020_centaur_readonly_role_only.sql from the centaur_investigator work, so this PR no longer carries a migration. That migration owns the all-table read-only grants.

## Tests
- cargo test -p centaur-iron-proxy
- cargo test -p centaur-iron-control postgres_listener_translates_to_pg_dsn_secret
- cargo test -p centaur-api-server args::tests::iron_control_registers_discovered_tool_secrets_on_infra_role

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/args.rs
+++ b/services/api-rs/crates/centaur-api-server/src/args.rs
@@ -2415,6 +2415,24 @@ mod tests {
                         == Some("TOOL_API_KEY")
             })
         }));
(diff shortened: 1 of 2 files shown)


mslipper:
This is no longer needed.

PR #642, opened 2026-06-17
Replace Centaur QA skill with staging smoke workflow

gakonst:
## Summary
- Replace the old local-stack QA skill with a staging/preview-oriented E2E smoke workflow
- Add required flows from the Slack discussion: concurrent agent turns, tool calls, attachments, thread history/user context, search/internal DB checks, scheduler idempotency, and promotion gating
- Keep the skill concise and prompt-driven so it tests real agent behavior instead of only scripted API curls

## Validation
- Validated skill frontmatter and line count with uv
- Checked the replacement is ASCII-only


mslipper:
superseded by https://github.com/paradigmxyz/centaur/pull/648

PR #652, opened 2026-06-17
[codex] allow shared Postgres database DSN roles

Zygimantass:
## Summary

Fix the staging api-rs startup crash caused by registering two logical `pg_dsn` tool secrets that both target the physical `centaur` database but require different proxy-owned roles/settings.

## Root Cause

`company_context` and `centaur_investigator` correctly declare different DSN env vars and roles in their tool manifests, but centaur-console rejected the second `pg_dsn` secret because it treated `database` as namespace-unique. The stable control-plane identity is `foreign_id`; `database` is the dbname clients use through the proxy and can be shared by distinct logical credentials.

## Changes

- Remove the console model uniqueness validation on `PgDsnSecret#database`.
- Keep same-database PG DSNs distinct in proxy sync by `foreign_id`.
- Add API/model/proxy-sync tests for same-database PG DSNs.
- Add an api-rs tool-discovery regression test showing two tool `pg_dsn` listeners can share `database = "centaur"` while retaining distinct roles/settings.
- Update console docs and api-rs comments so they no longer document `database` as globally unique.

## Validation

- `cargo fmt --check && cargo test -p centaur-api-server tool_discovery` in `services/api-rs`
- `PATH=/opt/homebrew/opt/ruby@3.3/bin:$PATH CENTAUR_CONSOLE_DB_PORT=55432 CENTAUR_CONSOLE_DB_PASSWORD=postgres RAILS_ENV=test bin/rails db:prepare`
- `PATH=/opt/homebrew/opt/ruby@3.3/bin:$PATH CENTAUR_CONSOLE_DB_PORT=55432 CENTAUR_CONSOLE_DB_PASSWORD=postgres bin/rails test test/models/pg_dsn_secret_test.rb test/controllers/api/v1/pg_dsn_secrets_controller_test.rb test/models/principal_test.rb`

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/tool_discovery.rs
+++ b/services/api-rs/crates/centaur-api-server/src/tool_discovery.rs
@@ -1346,6 +1346,95 @@ mod tests {
     }
 
     #[test]
+    fn pg_dsn_listeners_can_share_a_database_with_distinct_roles() {
+        let temp = temp_dir("api-rs-tools-pg-dsn-shared-db");
+        let base = temp.join("base");
+        write_tool(
+            &base.join("infra").join("centaur_investigator"),
+            r#"
+[project]
+description = "investigate centaur"
+
+[tool.centaur]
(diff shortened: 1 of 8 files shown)


mslipper:
Closing in favor of https://github.com/paradigmxyz/centaur/pull/653.

PR #656, opened 2026-06-18
Coalesce commentary Thinking task updates

decofe:
## Summary
- reuse one stable commentary Thinking task id instead of creating one task per commentary message
- show cumulative commentary details on that task
- default Slack task streams to native `dense` display mode to reduce task-card pressure
- add/update regression coverage for commentary coalescing and Slack stream payload shape

## Test plan
- bun test src
- bun run typecheck
- bun test test/chat-sdk-emulate.test.ts
- bun run check:types

Prompted by: @Rjected
Prompted by: @gakonst

### DIFF
--- a/packages/rendering/src/codex-app-server.test.ts
+++ b/packages/rendering/src/codex-app-server.test.ts
@@ -83,7 +83,7 @@ describe('CodexAppServerRendererEventMapper', () => {
     expect(task).toMatchObject({
       type: 'renderer.task.update',
       task: {
-        id: 'thinking-thinking-1',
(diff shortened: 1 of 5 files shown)


goksu:
Removed the preview label as it seems to not have the latest chart config that disables workflows from running on previews. FYI — @Zygimantass if you are still testing feel free to add it back and hopefully it picks the last config.

mslipper:
We've mostly stopped using thinking blocks, so closing this for now. Please reopen if you're still using thinking blocks and need this functionality updated.

PR #657, opened 2026-06-18
Fix Slack current-thread channel inference

decofe:
## Summary
- Cross-check API Slack session context against the bound tool thread key before inferring upload destinations
- Fall back to the tool thread key when API context is stale or mismatched
- Add regression coverage for Slack channel mention destinations and stale API context

## Tests
- `PYTHONPATH=/home/agent/branches/paradigmxyz/centaur/tools/productivity:/home/agent/branches/paradigmxyz/centaur uv run --no-project --with pytest --with slack-sdk --with structlog pytest /home/agent/branches/paradigmxyz/centaur/tools/productivity/slack/tests/test_client.py`

Prompted by: @gakonst

### DIFF
--- a/tools/productivity/slack/client.py
+++ b/tools/productivity/slack/client.py
@@ -1506,26 +1506,43 @@ class SlackClient:
         """Return the active Slack channel/thread, if any.
 
         Prefer API-owned session context so warm pooled sandboxes do not depend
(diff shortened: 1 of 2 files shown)


goksu:
We are going to take a different approach.

Issue #683, opened 2026-06-19
`api-rs` drops overlay DB-migration support and leaves base `company_context_documents` querying a removed `workflow_runs` table


lyoungblood:
## Summary

Migrating an overlay-based deployment from the legacy Python `api` to the Rust control plane (`api-rs`) surfaced two database-initialization regressions that are both still present on `main`. Neither is caught by a startup check, so they manifest at workflow runtime as `relation "..." does not exist`:

1. **`api-rs` no longer applies overlay DB migrations**, despite the Helm chart still setting `CENTAUR_OVERLAY_DIR` on the pod and the legacy `api` having supported overlay migration sets as a first-class, documented feature.
2. **The base workflow `company_context_documents` queries a `workflow_runs` table that the RS control plane never creates** (it was a legacy-`api` table, removed with `services/api`).

Verified against `main` @ `4fae0634` (#674).

---

### Background (the contract that existed)
The legacy Python `api` applied overlay-supplied SQL migrations automatically. `services/api/api/db.py` read `CENTAUR_OVERLAY_DIR` and, when `{overlay}/services/api/db/migrations` existed, appended an `overlay` `dbmate` migration set (its own `schema_migrations_overlay` version table to avoid collisions) and ran it alongside the core set on every startup (`create_pool(..., apply_migrations=True)` in the FastAPI lifespan hook). This was added intentionally in #566 ("Support overlay dbmate migration sets") and is still documented in `AGENTS.md` (`./scripts/dbmate --set overlay`, "overlay repos can extend the shared Postgres database without version collisions").

### What changed
- `services/api` (including `db.py` and the overlay applier) was removed in #521 (`chore: remove legacy api and slackbot`).
- `api-rs` applies **only** its own compile-time-embedded core migrations:
  `services/api-rs/crates/centaur-session-sqlx/src/lib.rs:19` → `static MIGRATOR = sqlx::migrate!("./migrations")`, run via `RUN_MIGRATIONS`. There is **no** code path that reads `CENTAUR_OVERLAY_DIR` for migrations, no `dbmate` invocation, and no second migration set anywhere in `services/api-rs/crates/**`.
[...]

PR #685, opened 2026-06-19
feat(chart): apply overlay DB migrations via an api-rs init-container

lyoungblood:
## What

Makes api-rs apply an overlay's `dbmate`-format DB migrations **after** its own core migrations, so overlay-owned tables get created under the RS control plane. Fixes **gap 1 of #683**. (Gap 2 — base `company_context_documents` querying the removed `workflow_runs` — is addressed in #678.)

## Why

api-rs applies only its compile-time-embedded core sqlx migrations. The legacy Python `api` *also* applied an overlay's `services/api/db/migrations/*.sql` (dbmate, tracked in `schema_migrations_overlay`) via `CENTAUR_OVERLAY_DIR` — added in #566, removed with `services/api` in #521, never reimplemented. So under RS, overlay-owned tables are never created and overlay workflows fail at runtime with `relation "..." does not exist`. The chart still set `CENTAUR_OVERLAY_DIR`, making it look like they'd still apply.

## How

The applier lives in api-rs and runs **immediately after `run_migrations()`**, so the core schema is in place before overlay migrations run (overlay is a layer above core — thanks @goksu).

- **`centaur-session-sqlx::apply_overlay_migrations(pool, dirs)`** — for each opted-in overlay dir, applies each pending `-- migrate:up` section in `NNN_` filename order, recorded in a `schema_migrations_overlay` ledger independent of the core `schema_migrations` (so applied versions are never re-run; an upgrade from the legacy dbmate path skips versions it already recorded).
  - dbmate markers are matched **anchored at column 0** with a whole-token match (an indented `-- migrate:down` in a `$$` body, or `migrate:upgrade`, won't match).
  - The version is the numeric filename prefix, **bound as a query parameter** (never string-interpolated).
  - A missing/empty `migrate:up` section **fails loudly** rather than recording an empty no-op.
  - `-- migrate:up transaction:false` is honored (single-statement, e.g. `CREATE INDEX CONCURRENTLY`).
  - Serialized across processes with a **`pg_advisory_lock`** (mirrors the core sqlx migrator), so a scaled deployment / overlapping rolling deploy applies the set once.
- **api-rs** waits (bounded) for the repo-cache readiness marker before applying, so it reads a complete clone.
- **chart** — opt-in per `overlays.sources` entry via `migrationsSubdir` (conventionally `services/api/db/migrations`). It has **no default** (unlike `toolsSubdir`/`workflowsSubdir`/`skillsSubdir`), so default and overlay-less installs are unaffected. When a source opts in (and `repoCache.enabled` + `apiRs.runMigrations`), the api-rs container gets `OVERLAY_MIGRATION_DIRS` + the readiness-marker env. No init-container.

```yaml
overlays:
  sources:
    - repo: your-org/your-overlay
      ref: main
      migrationsSubdir: services/api/db/migrations  # opt in to overlay migrations
```

## Notes for reviewers

1. **Ordering** — overlay migrations run after the core migrator, in the same startup path, gated on `RUN_MIGRATIONS`. Overlay sets are numbered independently of core (a separate ledger).
2. **`transaction:false`** must be a single statement — Postgres wraps a multi-statement simple-query string in an implicit transaction, which would defeat `CREATE INDEX CONCURRENTLY`. Documented in `MIGRATING_TO_RS.md`.
3. **Multiple opted-in sources** share the one `schema_migrations_overlay` ledger keyed by version, so they must use distinct version prefixes (documented).
4. **`CENTAUR_OVERLAY_DIR`** is left as-is (it's the prompt-path var, gated on `overlay.systemPrompt`); flagging that it's now vestigial for migrations — a separate cleanup if you want it.

## Testing

- `cargo test -p centaur-session-sqlx` — 9 unit tests (version parsing, anchored/whole-token markers, indented-marker no-truncation, comment-only → malformed, `transaction:false` detection); `cargo fmt --check` + `clippy` clean.
- `helm lint` passes; chart `version` bumped. `helm template` permutations: default/overlay-less → no overlay env; opt-in source → the api-rs container gets `OVERLAY_MIGRATION_DIRS` (only the opted-in source) + the readiness marker; no `migrationsSubdir` and `applyOverlayMigrations=false` → nothing.

### DIFF
--- a/contrib/MIGRATING_TO_RS.md
+++ b/contrib/MIGRATING_TO_RS.md
@@ -125,7 +125,83 @@ from .database import Database
 If a tool imports Centaur SDK modules from the base image, ensure the local
 runner includes the Centaur root (for example `/opt/centaur`) in `PYTHONPATH`.
 
-## 5. Verify console and per-sandbox proxies
+## 5. Apply overlay database migrations
+
+api-rs applies only its own embedded core migrations. The legacy Python `api`
+also applied an overlay's `services/api/db/migrations/*.sql` (dbmate format,
+tracked in a `schema_migrations_overlay` table) whenever `CENTAUR_OVERLAY_DIR`
+was set; the Rust control plane does not. Without this, overlay-owned tables are
+never created and overlay workflows fail at runtime with
+`relation "..." does not exist`.
+
+api-rs restores this: **after** it runs its core migrations, it applies overlay
+migrations from the repo-cache clone (so overlay schema may depend on core),
+tracked in `schema_migrations_overlay` — a ledger independent of the core
+`schema_migrations`, so an applied version is never re-run (an upgrade from the
+legacy dbmate path skips versions it already recorded).
+
+It is **opt-in per source**: set `migrationsSubdir` (conventionally
+`services/api/db/migrations`) on an `overlays.sources` entry to enable it for that
+overlay. Unlike `toolsSubdir`/`workflowsSubdir`/`skillsSubdir`, it has no default,
+so the common case — sources that carry no DB migrations, including the base repo
+— applies nothing and adds no startup wait.
+
+```yaml
+overlays:
+  sources:
+    - repo: your-org/your-overlay
+      ref: main
+      migrationsSubdir: services/api/db/migrations  # opt in to overlay migrations
+```
+
+When at least one source opts in (and `repoCache.enabled` + `apiRs.runMigrations`
+are true), api-rs waits for the repo-cache readiness marker, then applies each
+pending `-- migrate:up` section in `NNN_` filename order, recording the numeric
+version. A migration body and its ledger insert commit together, so a failure
+rolls back and retries on the next start. A configured source whose migrations
+directory is absent at runtime is skipped, matching the overlay-subdir model.
+
+Because overlay migrations run **after** the core migrations but as a separate,
+independently-numbered set, they should be **self-contained** relative to each
+other (numbered independently of core). Migration files must follow dbmate
+conventions:
+
+- A **zero-padded numeric `NNN_` prefix** (e.g. `001_…`, `010_…`) — the prefix is
+  the version (files without one are skipped), and applies in lexical filename
+  order, so pad consistently.
+- `-- migrate:up` / `-- migrate:down` markers at the **start of a line**; the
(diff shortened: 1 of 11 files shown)


goksu (changes requested):
> Ordering / self-containment. The init-container runs before the api-rs container's core migrations, so overlay migrations must be self-contained (not depend on core tables/extensions). This matches the convention that an overlay's migration set is numbered independently of core; documented in MIGRATING_TO_RS.md. If you'd rather guarantee core-then-overlay, an alternative is applying overlay migrations inside api-rs after run_migrations() — happy to take that direction instead.

I think we would want core migrations to occur first always as overlay is a layer above and core schema should drive decisions made at the overlay.

lyoungblood:
Thanks — agreed, core schema should drive overlay decisions. Reworked to apply overlay migrations **inside api-rs, right after `run_migrations()`**, so core always runs first (dropped the init-container, which structurally ran before the main container's core migrations).

- `centaur-session-sqlx::apply_overlay_migrations()` runs after the core migrator: each opted-in overlay dir's pending `-- migrate:up` section in `NNN_` order, tracked in a separate `schema_migrations_overlay` ledger (applied versions never re-run). Markers anchored at column 0 with whole-token matching; version is the numeric filename prefix, bound as a query param; `transaction:false` honored (single-statement); serialized across replicas with a `pg_advisory_lock` (mirroring the core migrator). Unit-tested.
- The chart now just passes `OVERLAY_MIGRATION_DIRS` + the repo-cache readiness marker to the api-rs container (opt-in per `overlays.sources` entry via `migrationsSubdir`; no-op for default/overlay-less installs).

`MIGRATING_TO_RS.md` updated accordingly.

lyoungblood:
@goksu based on @mslipper's Slack reply, should I close this, or is it still useful? Happy to close and pursue a Console-based approach instead.

goksu:
> @goksu based on @mslipper's Slack reply, should I close this, or is it still useful? Happy to close and pursue a Console-based approach instead.

@lyoungblood — Yes. Let's close this one. We have seen already internally when migration orders mattering creating issues down the line. I'd say if you can workshop something with @mslipper on Slack chat first we can then review the code updates after.  Thank you!

lyoungblood:
Thanks, appreciate you taking a look at it.

PR #687, opened 2026-06-22
fix: replace Linear API key placeholder

ryanli-me:
## Summary
- Change the Linear tool secret wiring from header injection to placeholder replacement.
- The Linear client sends `Authorization: LINEAR_API_KEY`, so iron-proxy must replace that placeholder with the real key instead of trying to inject a missing header.
- Keep both `api.linear.app` and `uploads.linear.app` scoped to the same `Authorization` replacement.

## Test plan
- `python3` TOML parse check for `tools/productivity/linear/pyproject.toml`
- `cargo test -p centaur-api-server tool_discovery::tests::discovers_http_oauth_and_overlay_shadowing`

### DIFF
--- a/services/api-rs/crates/centaur-perms/src/tests.rs
+++ b/services/api-rs/crates/centaur-perms/src/tests.rs
@@ -844,6 +844,37 @@ fn real_slack_tool_parses_and_translates() {
 }
 
 #[test]
+fn real_linear_tool_replaces_authorization_placeholder() {
+    let Some(tools_dir) = repo_tools_dir() else {
+        return;
+    };
+    let manifest = tools::find_tool(&[tools_dir], "linear").unwrap();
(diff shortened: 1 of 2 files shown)


Zygimantass:
hiya, what's the reason for doing this? afaik, linear should be able to function with `replace` mode

ryanli-me:
right here it's defined as `inject` mode. so the downstream Linear client couldn't work out of the box with overlay pattern. i think potentially i can make it work by overriding the mode in overlay config but that breaks the built-in tool boundary.

sorry if this is AI slop

mslipper:
I can't reproduce this on my end - we've been using Linear in `inject` mode for several months without issue. Happy to help debug further if you can provide some repro steps + logs. Closing for now.

PR #698, opened 2026-06-22
Add Amplitude product analytics tool

0xdiid:
Adds an Amplitude tool that gives agents the same read-and-query analytics reach as Amplitude's own assistant integration, built on the public REST surface so it works with a plain project key and secret. It can pull event trends over time, walk users through funnels, look at retention and realtime activity, inspect a single person's event history, and read the tracking plan. It follows the existing product-analytics tool already in the tree, supports both data regions, and leans on the established two-credential auth pattern so secrets are swapped in at the network edge rather than handled in-process.

Read-only by design: the parts of the assistant integration that create charts, dashboards, experiments, or flags need a different auth flow and aren't covered here.

Tested on our deployed instance

### DIFF
--- a/tools/infra/amplitude/pyproject.toml
+++ b/tools/infra/amplitude/pyproject.toml
@@ -1,0 +1,37 @@
+[project]
+name = "amplitude"
+description = "Amplitude product analytics — event segmentation, funnels, retention, taxonomy"
+version = "0.1.0"
+requires-python = ">=3.11"
+dependencies = [
+    "httpx>=0.27.0",
+    "typer>=0.12.0",
+    "rich>=13.0.0",
+    "python-dotenv>=1.0.0",
+]
+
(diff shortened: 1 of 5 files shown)


tools/infra/amplitude/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[project]
+name = "amplitude"
+description = "Amplitude product analytics — event segmentation, funnels, retention, taxonomy"
+version = "0.1.0"
+requires-python = ">=3.11"
+dependencies = [
+    "httpx>=0.27.0",
+    "typer>=0.12.0",
+    "rich>=13.0.0",
+    "python-dotenv>=1.0.0",
```
akshaan, line 10: nit: unused dep python-dotenv

akshaan (approved):

PR #702, opened 2026-06-22
feat(api-rs): Anthropic Messages-compatible /v1/messages ingress

tarrencev:
## Summary

Exposes centaur's durable, thread-keyed session control plane behind an **Anthropic Messages-compatible `POST /v1/messages`**, so any Anthropic SDK client or `claude -p` (via `ANTHROPIC_BASE_URL`) can drive a fully-configured centaur thread (tools, sandbox, durability) through the standard Messages API.

**Now fully implemented and validated end-to-end on a live deployment.** This folds in the model/system support (previously a stacked follow-up) plus the harness-output and client-compatibility fixes found while validating with `claude -p`.

## What it does

- **Endpoint:** `POST /v1/messages` (Anthropic Messages shape), registered alongside `/api/session/*`; streaming (SSE) and non-streaming.
- **Thread continuity:** keyed on the client's Claude Code session id — the `X-Claude-Code-Session-Id` header (the same id it uses for `--resume`/`--continue`) → thread `api:claude:<session-id>`, so a resumed CLI session continues the same durable thread and warm sandbox. No custom header to manage; absent → a fresh `api:<uuid>` thread.
- **Honors `model` and `system`** (first-request-wins): threaded into the sandbox harness env (`CLAUDE_MODEL`; `CENTAUR_EXTRA_SYSTEM_PROMPT` appended after `AGENTS.md`). `tools` are accepted but centaur owns the in-sandbox tools.
- **`AnthropicTranslator`** maps harness output → Anthropic content-block stream events, handling both harness shapes:
  - **claude** (natively Anthropic-shaped `assistant`/`user` events);
  - **codex** — both the dotted `item.agentMessage.delta` shape *and* the Rust harness-server's **slash-method/`params`** events (`item/agentMessage/delta`, `item/reasoning/textDelta`, `item/completed`). Text + reasoning + `tool_use` + `tool_result` → content blocks.
- **Client compat:** drives the turn off the **last `user` message**, so clients like Claude Code that append a trailing `system`-role message in `messages[]` work.

## Validation

End-to-end against a live deployment with `claude -p` (`ANTHROPIC_BASE_URL` → the ingress):
- basic reply (`Reply PONG` → `PONG`);
- **tool use** — the inner agent runs shell commands and returns their output;
- **model selection** — a requested `model` lands as `CLAUDE_MODEL` in the spawned sandbox.

`cargo fmt --check` · `clippy -D warnings` · `cargo test` green (translator unit cases for both harness shapes + streaming/non-streaming integration via the stdout-scripting backend; harness-server system-prompt test).

## Follow-ups (out of scope)

- `x-api-key` auth on `/v1/messages` (currently network-gated like `/api/session/*`).
- Full-history replay (v1 appends only the latest user turn; centaur owns thread history).
- Real usage-token accounting (v1 reports zeros).
- **Client tools** round-trip (let the caller's `tools` execute caller-side via an MCP bridge) — designed, not in this PR.

### DIFF
--- a/crates/harness-server/src/claude.rs
+++ b/crates/harness-server/src/claude.rs
@@ -1,6 +1,6 @@
-use std::env;
 use std::path::PathBuf;
 use std::process::Command as ProcessCommand;
+use std::{env, fs};
 
 use codex_app_server_protocol::UserInput;
 use serde_json::json;
@@ -237,8 +237,13 @@ impl HarnessServer for ClaudeCodeHarness {
         if !state.model.is_empty() {
             command.args(["--model", &state.model]);
         }
-        if PathBuf::from("AGENTS.md").is_file() {
+        if state.cwd.join("AGENTS.md").is_file() {
             command.args(["--append-system-prompt-file", "AGENTS.md"]);
+        }
+        if let Some(extra_system_prompt_file) = extra_system_prompt_file(state) {
+            command
+                .arg("--append-system-prompt-file")
+                .arg(extra_system_prompt_file);
         }
         if let Some(session_id) = &state.harness_session_id {
             command.args(["--resume", session_id]);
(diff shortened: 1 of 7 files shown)


tarrencev:
@gakonst lmk if this is of interest to you. I think it is a cool feature, allowing driving Centaur threads from Codex / Claude + supporting handoff from Slack to native clients. We can also improve it to support local tool calling so the model has a combination of Centaur + local dev env

tarrencev:
Shelving the /v1 ingress for now — moving the local↔sandbox unification to a prompt-native approach (the local harness runs local tool calls itself) rather than the in-sandbox bridge. Closing; can reopen later.

PR #704, opened 2026-06-22
Make the chat agent platform-aware (Slack/Discord/Linear/Github)

0xdiid:
We've been having issues on our Discord + Linear ingress where the agent mistakenly thinks it's in Slack, mostly due to a lot of Slack-specific stuff that was a part of that initial implementation. It figured it out after 2/3 turns but that's just wasted tokens, and looks bad in the chain of thought.

This teaches it where it actually is: every turn now carries a short note about which chat surface it's on and where its reply should go, and the base instructions no longer assume Slack. The session context and the SDK report the platform either way. GitHub threads (which landed on main after this went up) are covered too, so githubbot sessions get the same treatment.

The Discord side also gains the ability to send and pull down files, matching what Slack already had.

### Validated:
Screenshot 2026-06-23 at 12 27 09 PM
Screenshot 2026-06-23 at 9 19 30 PM


0xdiid:
Looks like y'all added some more session context stuff for slack, going to put this back in draft and generalize that instead!

goksu (approved):

PR #724, opened 2026-06-23
experiment: allow ambient Slack thread replies

decofe:
## Summary
- execute subscribed Slack thread replies without requiring a fresh @ mention
- update the sandbox prompt to judge thread flow and stay silent unless a reply is useful
- update Slackbot emulator coverage for ambient follow-up execution

## Tests
- pnpm --filter slackbotv2 test

Prompted by: @goksu


goksu:
Closing for now as it is not super trivial to have this work well at an acceptable UX.

PR #738, opened 2026-06-24
Suppress reasoning cards in Slack streams

decofe:
## Summary
- add a renderer option to omit reasoning-derived Thinking task cards
- default Slackbot rendering and recovery to omit those reasoning cards while preserving commentary, plan, command progress, and final answer rendering
- update Slackbot stream expectations for the quieter Slack surface

Prompted by: @DaniPopes

## Tests
- bun test packages/rendering/src/codex-app-server.test.ts
- bun test services/slackbotv2/test/chat-sdk-emulate.test.ts
- pnpm --filter @centaur/rendering typecheck
- pnpm --filter slackbotv2 check:types

### DIFF
--- a/packages/rendering/src/codex-app-server.test.ts
+++ b/packages/rendering/src/codex-app-server.test.ts
@@ -152,6 +152,23 @@ describe('CodexAppServerRendererEventMapper', () => {
     })
   })
 
+  it('can omit reasoning Thinking tasks for Slack surfaces', () => {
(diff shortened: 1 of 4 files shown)


goksu (changes requested):
Why make this a prop? We should remove if the consensus is that they are not useful.

DaniPopes:
mistake, didn't mean to open pr

PR #755, opened 2026-06-24
fix(harness-server): retry codex turns that fail on transient engine warmup

0xdiid:
AWS Bedrock is a bit weird sometimes and 404s randomly with "engine not found", but this is a transient error
```
error → JSON-RPC error -32602: Job registration failed:
        Engine bad request: Task submission failed with status 404 Not Found: Engine not found
```

That being said, codex vaporizes when this happens, killing the pod and making it unusable until at least the pod is torn down and recreated. This just adds retry logic so it'll retry that request twice instead of killing everything. Default 2, can set `CODEX_ENGINE_RETRY_MAX` to override

### DIFF
--- a/crates/harness-server/src/codex.rs
+++ b/crates/harness-server/src/codex.rs
@@ -302,15 +302,22 @@ fn run_codex_user_turn<W: Write>(
                 HarnessServerError::Protocol("turn/start response missing turn.id".to_string())
             })?
             .to_string();
-        let retry_allowed = retries < max_retries;
         match codex.read_until_turn_terminal(
             stdout,
             thread_id.as_deref().unwrap_or_default(),
             &turn_id,
-            retry_allowed,
         )? {
             TurnTermination::Done => return Ok(()),
-            TurnTermination::RetriableEngineError => {
+            TurnTermination::RetriableEngineError { withheld } => {
+                if retries >= max_retries {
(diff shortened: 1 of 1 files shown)


crates/harness-server/src/codex.rs
```diff
@@ -456,27 +486,44 @@ impl CodexJsonRpcChild {
...
+    /// transient engine-registration error before streaming any output, the
+    /// error (and the `systemError` status that precedes it) are withheld and
+    /// `RetriableEngineError` is returned so the caller can re-submit the turn.
     fn read_until_turn_terminal<W: Write>(
         &mut self,
         stdout: &mut W,
         thread_id: &str,
         turn_id: &str,
-    ) -> Result<()> {
+        retry_allowed: bool,
```
goksu: Let's remove this prop to keep the fn signature unchanged since config value will already allow it to be `CODEX_ENGINE_RETRY_MAX = 0`.

goksu (changes requested):
Have one change request. Otherwise looks good and happy to land! Thank you.

crates/harness-server/src/codex.rs
0xdiid: reworked

goksu (approved):

PR #756, opened 2026-06-24
fix(sandbox): honor TOOL_ALLOWLIST when installing/shimming tools

tarrencev:
## What

`services/sandbox/install_tool_shims.py` now honors `TOOL_ALLOWLIST` in **both** the install path (`_copy_published_tools`) and the shim-generation path (`_discover_scripts`). When `TOOL_ALLOWLIST` is set, only the listed tools are installed and shimmed; unset/empty preserves today's behavior (install all) — fully backward compatible. Test added.

## Why

`TOOL_ALLOWLIST` already scopes the agent's tool catalog elsewhere, but the sandbox shim installer ignored it — every mounted tool was installed and shimmed regardless of the allowlist. Consequences:

- **Phantom, credential-less tools in the catalog.** Tools the deployment never configured still show up to the agent, but they have no secrets provisioned, so they surface as broken options that mislead the model and waste its context.
- **Wasted setup.** Each tool's wheel builds lazily on first use via `uvx --from`; installing/shimming tools nobody enabled is work the deployment didn't ask for.

This makes the sandbox catalog exactly the configured tools.

## Test

`services/sandbox/test_install_tool_shims.py::...test_tool_allowlist_restricts_installed_tools` — allowlisted tool installed, unconfigured tool skipped.

### DIFF
--- a/services/sandbox/install_tool_shims.py
+++ b/services/sandbox/install_tool_shims.py
@@ -21,6 +21,20 @@ def _split_paths(value: str) -> list[Path]:
     return [Path(part) for part in value.split(":") if part]
 
 
+def _tool_allowlist() -> set[str] | None:
+    """Tool package names to install, from ``TOOL_ALLOWLIST``.
+
+    Returns ``None`` when unset/empty -> install every mounted tool (backward
+    compatible). When set, only tools whose package name is listed are installed,
+    so the sandbox catalog is exactly the configured tools and the agent neither
+    sees nor wastes context on unconfigured ones (which also lack credentials).
+    """
+    raw = os.environ.get("TOOL_ALLOWLIST", "").strip()
+    if not raw:
+        return None
+    return {name.strip() for name in raw.split(",") if name.strip()}
+
+
 def _home_dir() -> Path:
     return Path.home()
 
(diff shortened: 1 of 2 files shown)


services/sandbox/install_tool_shims.py
```diff
@@ -333,6 +353,12 @@ def _discover_scripts(tool_dirs: list[Path]) -> dict[str, dict[str, str]]:
                 print(f"warning: failed to read {pyproject}: {exc}", file=sys.stderr)
                 continue
             project = data.get("project") or {}
+            # Only shim allowlisted tools (by package dir or project name) so the
+            # agent's catalog is exactly the configured tools. Unset -> shim all.
+            if allowlist is not None and allowlist.isdisjoint(
```
goksu: `isdisjoint()` is non-idiomatic Python and we should avoid it here. Since we're just checking whether either of two names is in the allowlist.

goksu (changes requested):
Thanks! Overall makes sense. Once updated happy to land.

tarrencev:
Addressed the review feedback in a943e20 by replacing the `isdisjoint()` allowlist check with explicit package/project-name membership checks. Relevant sandbox shim test still passes locally: `python -m pytest services/sandbox/test_install_tool_shims.py`.

goksu (approved):

PR #757, opened 2026-06-24
feat(api-rs): set CENTAUR_OVERLAY_DIR for repo-cache overlay delivery

tarrencev:
## What

Under repo-cache overlay delivery (the `overlays.sources` model, no overlay-image mount), api-rs now injects `CENTAUR_OVERLAY_DIR=/home/agent/github/<owner>/<repo>` into the sandbox env, derived from the overlay repo slug. The chart sets a new `CENTAUR_OVERLAY_REPO` from the highest-priority (last) overlay source; api-rs turns the slug into the clone path under the repos mount.

## Why

With repo-cache delivery the overlay repo is cloned at `/home/agent/github/<owner>/<repo>`, but `CENTAUR_OVERLAY_DIR` was only ever set in **overlay-image** mode (from `overlay.mountPath`). So in repo-cache mode it was **never set**, and the sandbox entrypoint's overlay branches were silently skipped — the org's `services/sandbox/SYSTEM_PROMPT.md` and `.agents/skills/` simply didn't apply. The overlay was delivered to the filesystem but never activated.

This wires `CENTAUR_OVERLAY_DIR` to the actual clone path so repo-cache overlay delivery works end-to-end (prompts + skills applied). Overlay-image mode is unchanged (it still sets `CENTAUR_OVERLAY_DIR` from `overlay.mountPath`, and api-rs no-clobbers when it's already present).

## Test

`centaur-api-server` args tests cover the slug→dir derivation (`/home/agent/github/owner/repo`), the `CENTAUR_OVERLAY_REPO` env override, and the no-clobber-when-already-set behavior.

### DIFF
--- a/contrib/chart/templates/apirs.yaml
+++ b/contrib/chart/templates/apirs.yaml
@@ -175,6 +175,17 @@ spec:
             - name: CENTAUR_OVERLAY_DIR
               value: {{ .Values.overlay.mountPath | quote }}
 {{- end }}
+{{- if $useOverlayRepoCache }}
+            # Point the sandbox CENTAUR_OVERLAY_DIR at the org overlay clone
+            # under the repos mount (/home/agent/github/<repo>). The org overlay
+            # is the highest-priority (last) repo-cache overlay source -- the one
+            # that shadows the base and carries the org's services/sandbox/
+            # SYSTEM_PROMPT.md and .agents/skills. api-rs derives the sandbox env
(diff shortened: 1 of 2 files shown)


mslipper:
Thanks for digging into this. I will be closing this as obsolete. Since the PR was opened, #1161 was merged and sandbox startup now discovers prompts from all mounted overlay repositories while filtering out the baked base prompt.

PR #811, opened 2026-06-29
Add Datadog observability tool

decofe:
## Summary
- add a read-only Datadog infra tool with logs, metrics, monitors, hosts, and dashboard commands
- wire Datadog API/application keys through tool metadata for supported Datadog API sites
- document the tool in tools/README.md

## Verification
- uv run python -m compileall tools/infra/datadog
- uvx --from tools/infra/datadog datadog --help
- uvx ruff check tools/infra/datadog

Prompted by: @Zygimantass


goksu (approved):

cjustice:
Heads up — as-is this tool will 403 at runtime. iron-proxy's `header_allowlist` (`services/iron-proxy/iron-proxy.yaml`) doesn't include Datadog's `DD-API-KEY` / `DD-APPLICATION-KEY` headers, and they don't match the `^x-…(api-key|token|auth|key)$` catch-all — so the per-sandbox proxy strips them before egress and every request to `api.datadoghq.com` reaches Datadog unauthenticated (403).

I opened #944 to add those two headers to the allowlist; it's the enabling primitive for this tool. Once both land, the read-only key auth here should work end-to-end. cc @Zygimantass

PR #824, opened 2026-06-30
fix: make `centaur_sdk` importable without rich

### LINKED ISSUE #823: `centaur_sdk` forces a rich dep on every tool runner
The sandbox tool runner imports `from centaur_sdk.tool_sdk import …`, which executes `centaur_sdk/__init__.py`. 

That  `__init__` eagerly imports cli_tables, which imports rich at module top. 

Tool call envs are built from the tool's own (often `dependencies = []`) pyproject and the sandbox is offline, so `rich` isn't there and can't be fetched, every tool call dies:

```
File "<string>", line 11, in <module>
File "/opt/centaur/centaur_sdk/__init__.py", line 11, in <module>
    from centaur_sdk.cli_tables import Table, render_text_table
File "/opt/centaur/centaur_sdk/cli_tables.py", line 5, in <module>
    from rich.table import Table as RichTable
ModuleNotFoundError: No module named 'rich'
```

Repro
```
PYTHONPATH=/opt/centaur python -c "from centaur_sdk.tool_sdk import secret"
# ModuleNotFoundError: No module named 'rich'
```

Possible fix 
- `i_tables.py and __init__.py`; keep `tool_sdk (stdlib)` eager. rich only imported when a CLI actually renders a table (#824)
-  make the runner in `services/sandbox/install_tool_shims.py` (CALL_RUNNER) not import `centaur_sdk` at all — pass `ToolContext` via env/JSON so tool envs never have to satisfy SDK deps. (#825)

lean-apple:
Fixes #823.

- Lazy-load `Table`/`render_text_table` so importing `centaur_sdk` / `centaur_sdk.tool_sdk` no longer pulls in `rich`, which the offline, stdlib-only tool runner does not have.
- Stop the sandbox tool runner (`install_tool_shims.py` `CALL_RUNNER`) from importing `centaur_sdk` at all — derive `ToolContext` from `CENTAUR_TOOL_NAME`/`CENTAUR_THREAD_KEY` env in the SDK so tool envs never have to satisfy SDK deps.

### DIFF
--- a/centaur_sdk/__init__.py
+++ b/centaur_sdk/__init__.py
@@ -8,7 +8,8 @@ Public API:
 
 from __future__ import annotations
 
-from centaur_sdk.cli_tables import Table, render_text_table
+from typing import Any
+
 from centaur_sdk.tool_sdk import (
     ToolContext,
     current_session_context,
@@ -36,3 +37,11 @@ __all__ = [
     "secret",
     "set_tool_context",
 ]
+
+
+def __getattr__(name: str) -> Any:
+    if name in ("Table", "render_text_table"):
+        from centaur_sdk import cli_tables
+
+        return getattr(cli_tables, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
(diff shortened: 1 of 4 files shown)


lean-apple:
I guess we can close this now @mslipper

PR #826, opened 2026-06-30
feat: package agent-sandbox CRDs as chart-managed templates

lean-apple:
Render the agents.x-k8s.io CRDs through templates behind an installCRDs flag (default true) so helm upgrade keeps them current, instead of Helm's install-only crds/ dir. CRDs carry helm.sh/resource-policy: keep so uninstall leaves them in place. Existing installs need a one-time Helm adoption of the pre-existing CRDs before the first upgrade.

cc @penumbra23


mslipper:
I think this is convenient, but it's a tall order to ask existing installs (of which there are many) to run a manual Helm adoption command. Closing for now; can revisit once there's a larger release we can use to ask users to adopt this.

PR #829, opened 2026-06-30
Fix vlogs nested VictoriaLogs fields

decofe:
## Summary
- query thread, event, and execution identifiers across top-level, `fields.*`, and `span.*` VictoriaLogs field layouts
- normalize nested VictoriaLogs fields back to stable top-level aliases in `vlogs` helper results
- add regression tests for nested thread and tool-call records

## Context
We recently upgraded VictoriaLogs for v1.50.0's multi-field search support. The relevant changelog entry is: "add an ability to search across multiple log fields with the `field_name_prefix*:filter` syntax" (https://docs.victoriametrics.com/victorialogs/changelog/#v1500).

The current breakage has two parts. The infra-side fix restores `_msg` promotion by setting `VL-Msg-Field`; this PR fixes the client-side assumption that important identifiers always live at top level. Live records now commonly expose identifiers as `fields.thread_key` / `span.thread_key` and events as `fields.event` / `span.event`, so helper methods like thread logs, thread traces, tool analytics, and execution timelines must query and normalize both old and new shapes. VictoriaLogs documents that `VL-Msg-Field` controls which structured field becomes `_msg`; without it, VictoriaLogs reads `_msg` only and falls back to the default message value (https://docs.victoriametrics.com/victorialogs/data-ingestion/#http-parameters).

## Validation
- `PYTHONPATH=/home/agent/branches/paradigmxyz/centaur/tools/infra uv run --no-project --with pytest --with httpx pytest tests/test_client.py`
- `uv run --no-project --with pytest --with httpx pytest test_client.py`
- `git diff --check`

Prompted by: @sds

### DIFF
--- a/tools/infra/vlogs/client.py
+++ b/tools/infra/vlogs/client.py
@@ -22,6 +22,18 @@ def _field_expr(field: str, value: str) -> str:
     return f"{field}:{_quote_logsql_value(value)}"
 
 
+def _field_any_expr(fields: list[str], value: str) -> str:
+    return "(" + " OR ".join(_field_expr(field, value) for field in fields) + ")"
+
+
+def _field_exists_any_expr(fields: list[str]) -> str:
+    return "(" + " OR ".join(f"{field}:*" for field in fields) + ")"
+
+
+def _field_term_any_expr(fields: list[str], value: str) -> str:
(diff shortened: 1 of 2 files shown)


decofe:
Closing per follow-up investigation: we no longer think these VictoriaLogs message-field/client changes are the root cause. Prompted by: @sds

PR #838, opened 2026-07-01
ccentaur: add node selector, tolerations and resources

penumbra23:
## Summary

Adds operator control over where the pods api-rs spawns at runtime get scheduled and what resources they request/limit:

- Node placement - `sandbox.nodeSelector` and `sandbox.tolerations` are applied to every spawned agent sandbox and its paired per-sandbox iron-proxy pod
- iron-proxy resources - per-sandbox iron-proxy containers now accept CPU/memory requests + limits (`ironProxy.resources`)
- sandbox resources - deployment-wide CPU/memory requests + limits for the agent container (`sandbox.resources`), applied as a default in the spawner

## Reasoning / Purpose

Agent sandboxes are bursty, untrusted compute; the control plane is long-lived and trusted. Co-locating both on the same nodes couples their scaling and lets agent load contend withthe control plane. This lets operators put sandboxes (and their per-sandbox proxies, which scale 1:1 with them) on a dedicated, independently-autoscaled, optionally-tainted node pool, while keeping baseline services on stable nodes. Resource requests/limits make that pool schedulable/bin packable and stop a runaway sandbox or proxy from starving neighbors - the iron-proxy in particular ran completely unbounded before.

Design notes:
- Placement lives on AgentSandboxConfig (deployment-wide) rather than per-spawn, since the target pool is a cluster-level decision
- The api-self/baseline pods are intentionally excluded - only runtime-spawned sandbox + per-sandbox proxy pods are placed
- JSON config is validated at startup (fail-fast) so a malformed selector/toleration can't silently schedule agents cluster-wide
- Empty values are omitted end-to-end, so existing deployments keep default scheduling/resources (caveat: `sandbox.resources.requests` ships with cpu: 100m, memory: 512Mi defaults, so sandboxes now get those requests on upgrade - set requests: {} to opt out)

## Testing

- `cargo fmt --all --check, cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test` (affected crates)
- `build_agent_sandbox` applies nodeSelector/tolerations/resources and the typed CRD from_value round-trips them; omitted when unset
- `build_iron_proxy_pod` inherits placement + resources; omitted when unset.
- clap arg parsing: valid JSON flows into config; invalid nodeSelector/tolerations are rejected; empty leaves defaults.
- `helm lint` + `helm template` - env vars render with values set; omitted at defaults; rendered JSON round-trips through the Rust parsers.


josescuderoh:
@penumbra23 looking to get this change merged any time soon?

penumbra23:
@josescuderoh Resolved the conflicts, should be rdy to merge

mslipper:
Superseded by #1273 and #1207.

PR #865, opened 2026-07-02
feat: always show the model on Console session links and thread header

goksu:
## Summary
- The Slack "Open session in Console · Harness · Model" context line dropped the model segment unless the thread had an explicit `--model`/`--opus`/... override (`--claude` alone showed only "Claude Code"). slackbotv2 now falls back to the harness's baked-in default model — `claude-opus-4-8` for claudecode (harness/claude/settings.json) and `gpt-5.5` for codex (harness/codex/config.toml) — so the line always names a model. Amp has no fixed default model and keeps omitting the segment.
- Explicit `--model` overrides are now recorded in execute metadata (`metadata.model`) so the session DB self-describes the model a thread ran on. The default fallback is display-only and is not sent to the harness.
- The Console thread detail header now shows the model between the harness and the timestamp ("Slack · Claude Code · claude-opus-4-8 · 3m ago"), via a new `thread_model_label` helper: latest execution `metadata.model` → session `metadata.model` → per-harness default map.

## Caveats
- The harness default models are mirrored as small maps in slackbotv2 and the Console (with pointer comments both ways), since the pinned values live in harness images those services can't read at runtime. If harness/claude/settings.json or harness/codex/config.toml change models, update both maps.
- For historical threads without a recorded override, the Console shows the *current* default, which may differ from what an old thread actually ran on.

## Tests
- `bun test` in services/slackbotv2 (68 pass), including a new emulate case for `--claude` with no `--model` and metadata-recording assertions; `bun run check:types` clean.
- Added Console unit tests for `thread_model_label` (execution metadata, session metadata, default fallback, amp nil). Not run here — console Rails tests only run inside the codex-centaur-console-web container, which doesn't mount worktree edits.

### DIFF
--- a/contrib/chart/templates/console.yaml
+++ b/contrib/chart/templates/console.yaml
@@ -121,6 +121,15 @@ spec:
                   key: {{ printf "%sDATABASE_URL" $prefix }}
             - name: CENTAUR_CONSOLE_SLACKBOTV2_USER_NAME
               value: {{ .Values.slackbotv2.userName | quote }}
+{{- range $name := tuple "CLAUDE_MODEL" "CODEX_MODEL" }}
+{{- if hasKey $.Values.sandbox.extraEnv $name }}
+            # Mirror the deployer's harness default-model override
+            # (sandbox.extraEnv) so the Threads view names the model threads
+            # without a recorded override actually ran on.
+            - name: {{ $name }}
+              value: {{ index $.Values.sandbox.extraEnv $name | toString | quote }}
+{{- end }}
+{{- end }}
             - name: IRON_CONTROL_INITIAL_USER_EMAIL
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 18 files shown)


goksu:
Follow-up commit removes the hardcoded default-model maps:

- **slackbotv2** now imports the models pinned in `harness/claude/settings.json` / `harness/codex/config.toml` directly (the Dockerfile copies `harness/` into the image), and prefers the `CLAUDE_MODEL` / `CODEX_MODEL` env vars — the same documented knobs deployers set in `sandbox.extraEnv`.
- **Chart** mirrors `sandbox.extraEnv.CLAUDE_MODEL/CODEX_MODEL` into the slackbotv2 and Console deployments automatically (service-level `extraEnv` still wins), so the model is configured in one place.
- **Execute metadata** now records the effective model (override, else configured/baked default), so the session DB self-describes and historical threads keep showing the model they actually ran on rather than the current default.
- **Console** resolves: recorded execution/session metadata → mirrored env vars → parsing the harness config files when present (`CENTAUR_HARNESS_CONFIG_DIR` or the repo checkout, covering local mirrored-snapshot browsing). No Ruby copy of the model strings remains.

Verified: 137 slackbotv2 tests + type check pass; `helm template` renders the mirrored env in both deployments, omits it when unset, and doesn't duplicate when overridden per-service.

PR #871, opened 2026-07-02
feat: periodically re-scan for orphaned executions

Zygimantass:
## Problem

`adopt_orphaned_executions` runs **once at process startup**, and its lease-deferral path retries only "on next startup" — i.e. at the next deploy. That misses the standard rollout race:

1. New pod boots and runs its startup adoption scan.
2. Old pod is still alive mid-turn (stdout-owner lease active), so there is nothing to adopt — or adoption is deferred.
3. Old pod hits its termination grace period and dies mid-turn.
4. The execution is now orphaned, seconds *after* the only scan that could have rescued it. The turn's finished answer sits unread in the sandbox pod logs (exactly what `read_output_since` recovery exists for) until the next deploy.

This exact sequence silently ate a completed turn in stg on 2026-07-02: the deploy of #870 rolled api-rs at 12:37, the new pod scanned at 12:37:06, the old pod died at ~12:37:31 mid-turn, and codex's completed answer (visible in the sandbox pod logs at 12:38:58) was never delivered — the user saw only activity summaries and no reply.

## Change

- **Periodic re-scan**: the adoption scan now re-runs on an interval — `--session-execution-adoption-interval-secs` / `SESSION_EXECUTION_ADOPTION_INTERVAL_SECS`, default 15, `0` restores startup-only. The interval bounds how long a turn handed off by a terminating pod (#872) stays frozen before a peer adopts it; a steady-state tick is a single SELECT because lease-held executions are skipped before any session or sandbox reads. The first scan still runs immediately at startup. Safety against live owners is already provided by the stdout-owner lease (45s TTL, renewed every 10s by a dedicated task): scans defer while a lease is active and claim it atomically (`claim_expired_stdout_owner`) once the owner dies.
- **Queued-row grace window**: queued executions younger than 120s are now skipped instead of failed. Executions are only queued for the instant between `create_execution` and the running transition, but a scan racing a live `execute_session` (periodic tick, or a startup scan overlapping the old pod's in-flight request during a rollout) could observe that window and kill a healthy row.
- **Deferral dedup**: `session.execution_adoption_deferred` (event + info log) is recorded only on the first deferral per execution, so a long turn owned by a live peer doesn't accumulate one event per tick; repeats log at debug.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` clean.
- New DB-gated tests (run with `SESSION_RUNTIME_TEST_DATABASE_URL` against ParadeDB): young queued rows are left alone then failed once aged; a lease held by another control plane defers adoption and the recorded terminal is recovered after the lease expires; repeated scans record the deferral event exactly once and adopt immediately after the lease is released.
- Full `centaur-session-runtime` suite: 69/72 pass locally; the 3 failing `stdout_eof_*` tests fail identically on pristine `main` in the same environment (pre-existing, and not exercised in CI since `SESSION_RUNTIME_TEST_DATABASE_URL` is unset there).

Companion PR (separate): graceful SIGTERM drain that releases stdout-owner leases at shutdown, so a replacement pod's re-scan picks orphans up immediately instead of waiting out the 45s lease TTL.

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/args.rs
+++ b/services/api-rs/crates/centaur-api-server/src/args.rs
@@ -110,6 +110,11 @@ impl Args {
     pub(crate) fn shutdown_execution_drain_timeout(&self) -> Duration {
         Duration::from_secs(self.server.shutdown_execution_drain_timeout_secs)
     }
+
+    pub(crate) fn execution_adoption_interval(&self) -> Option<Duration> {
+        (self.server.execution_adoption_interval_secs > 0)
+            .then(|| Duration::from_secs(self.server.execution_adoption_interval_secs))
+    }
 }
 
 pub(crate) struct IronControlRuntime {
@@ -463,6 +468,17 @@ pub(crate) struct ServerArgs {
         value_parser = clap::value_parser!(u64).range(0..=600)
     )]
     shutdown_execution_drain_timeout_secs: u64,
+    /// How often to re-run the orphaned-execution adoption scan after the
+    /// startup pass. Executions orphaned while the process is already
+    /// running (e.g. a rolling deploy terminating the previous pod mid-turn
+    /// after this pod's startup scan) are only recovered by these re-scans.
+    /// 0 disables re-scans and keeps the startup-only behavior.
+    #[arg(
+        long = "session-execution-adoption-interval-secs",
+        env = "SESSION_EXECUTION_ADOPTION_INTERVAL_SECS",
+        default_value_t = 60
+    )]
+    execution_adoption_interval_secs: u64,
 }
 
 #[derive(Debug, ClapArgs)]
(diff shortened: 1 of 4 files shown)


Zygimantass:
Follow-up commit 03a797f3 addresses findings from an adversarial review pass (5 reviewers, 24 verified findings across this PR and #872): the scan now checks the stdout-owner lease before touching the session row or sandbox backend (no spurious adoption-deferred events on this process's own healthy turns, no per-tick kube status calls for lease-held executions), the queued-row grace window only applies to periodic scans (a one-shot startup scan still fails queued orphans immediately, so interval=0 cannot wedge a thread), deferral dedup survives transient error ticks, and the lease-expiry test is now deterministic (SQL-forced expiry instead of racing a 200ms TTL). Also adds coverage for the self-owned-skip and the spawned periodic loop.

PR #883, opened 2026-07-02
feat(mcp): scope tool visibility to the principal's granted secrets

Zygimantass:
## What

The MCP `tools/list` showed **every** discovered tool to every authenticated principal, regardless of whether that principal has access to the tool's secrets. A principal without a tool's secret still saw it, and the call then failed later at the proxy with a confusing upstream 401 — exactly what happens today for a freshly OAuth'd, roleless console principal (e.g. `prn_MXythW3X`).

This scopes the MCP catalog to the tools a principal can actually use.

## How

A tool declares the secrets it needs in `[tool.centaur].secrets`; those secrets are registered under a **per-tool iron-control role** `tool-<slug of the tool dir name>` (`RoleSpec::tool`, the same role `centaur-perms` assigns). So the predicate is:

> principal **P** may use tool **T** ⟺ T declares no secrets, **or** P holds role `tool-<slug(T)>`.

- **Discovery** (`tool_discovery.rs`): `DiscoveredTool` now carries `required_tool_role: Option<String>` — `Some("tool-<slug>")` when the tool declares secrets, else `None`.
- **iron-control** (`session.rs`): `SessionRegistrar::principal_role_foreign_ids` returns a principal's assigned role `foreign_id`s; `SessionRuntime::mcp_principal_role_foreign_ids` exposes it (or `None` when iron-control isn't configured).
- **MCP** (`mcp.rs`): `tools/list` is filtered and `tools/call` is gated by `McpToolVisibility`. Tools with no declared secrets are always visible. The whoami tool is unaffected.

## This is a discoverability fix, not a new security boundary

Secret access is **already enforced at runtime** by the per-sandbox iron-proxy: it only injects secrets the principal is granted; an ungranted tool sends the literal placeholder upstream and gets a 401. So nothing leaks today — the problem is phantom tools and confusing failures. Because enforcement is independent, filtering is deliberately **fail-open**: if grants can't be resolved (iron-control unreachable or unconfigured) we show all tools and let the proxy enforce, rather than hide tools on a transient control-plane blip.

## Rollout / safety

Gated behind **`CENTAUR_MCP_FILTER_TOOLS_BY_GRANTS` (default off)**. It's inert until per-tool roles (`tool-<slug>`) are assigned to the principals that need them — enabling it before then would hide every secret-backed tool from everyone. Recommended order: assign roles (via `centaur-perms`) → enable in staging → prod.

## Follow-up (separate PR)

Apply the same predicate to **sandbox** tool installation by computing a per-principal `TOOL_ALLOWLIST` at boot (the plumbing already exists in `install_tool_shims.py`); the sandbox already runs as an `iron_control_principal`.

## Tests

- `tool_discovery`: `catalog_records_per_tool_role_only_for_secret_backed_tools` (secret-backed tool → `tool-<slug>`, secretless → `None`).
- `mcp`: `McpToolVisibility::allows` — unfiltered shows all; `Roles` gates secret-backed tools by membership; empty role set hides all secret-backed tools; secretless tools always visible.
- Full `cargo test` (api-server / iron-control / session-runtime), clippy, and fmt all clean.

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/mcp.rs
+++ b/services/api-rs/crates/centaur-api-server/src/mcp.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     env, fs,
     path::PathBuf,
     sync::{Mutex, OnceLock},
@@ -117,8 +117,9 @@ pub(crate) async fn mcp_post(
         "ping" => json!({}),
         "tools/list" => {
             ensure_mcp_scope(&principal.scopes, "mcp:tools")?;
+            let visibility = mcp_tool_visibility(&state, &principal).await;
             let mut tools = vec![mcp_whoami_tool()];
-            tools.extend(mcp_centaur_tool_entries()?);
+            tools.extend(mcp_centaur_tool_entries(&visibility)?);
             json!({
                 "tools": tools,
             })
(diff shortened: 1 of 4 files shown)


Osraka:
I traced the current boundary while reviewing this work:

- the MCP access token identifies a principal; it does not copy live grants into the token,
- tools/call reuses the per-principal mcp:<principal> tool-host session, and
- actual credential enforcement remains at the session-specific per-principal iron-proxy.

The RFC says a role or grant change should be visible on the next runner/proxy sync without reissuing the access token. For a still-valid MCP token, is the intended contract that revoking a principal role or grant takes effect on the next tools/call? If so, what sync boundary should a regression test assert?

I am not proposing an authorization behavior change. If this is the intended contract, I would first add a narrow allowed -> revoke -> denied test and make the proxy-sync and feature-flag expectations explicit.

mslipper:
Closing this out - MCP has shifted considerably. I will tackle reintroducing this feature myself shortly.

PR #887, opened 2026-07-02
fix: capture Slack app message content and unfreeze busy-channel ETL sync

lyoungblood:
## Summary

Two coupled defects in the Slack ETL (`workflows/slack/`), both verified on a live deployment where six busy channels had been silently frozen for 2–5 weeks:

### 1. Bot/app message content is never stored

`SlackEtlClient._serialize_message` allowlists message keys and drops `attachments`/`blocks`. Bot integrations (the GitHub app, alerting apps) post with an empty top-level `text` and put all content in legacy attachments — so their messages are stored content-free. On our deployment, **915 of 916 rows** in a GitHub-bot channel had empty `text` and no content in `raw_payload`, making them invisible to the FTS index over `text` and to `company_context_documents`.

**Fix:** pass `attachments` and `blocks` through into `raw_payload`, and synthesize the `text` column from attachment `fallback` (else `pretext`/`title`/`text`) only when top-level text is empty. Human messages are untouched. (The interactive tool client `tools/productivity/slack/client.py` has the same allowlist gap; left as a follow-up since nothing persists its output.)

### 2. Busy channels deadlock: frozen watermark + clobbered continuations

`conversations.history` **anchors at `oldest` when only `oldest` is passed** — the page holds the *oldest* slice of the window, paging forward in time. (Verified empirically against the live API; the code labels the page `"order": "desc"` and assumes the opposite.) Consequences on any channel whose backlog exceeds one page (`limit`, default 100, per tick):

- The incremental tick re-reads the same oldest page forever; `latest_seen = max(page ts)` stabilizes at the channel's *density fixed point* (where ~`limit` messages span `thread_lookback_days`) and can even **regress** the stored watermark. New head messages are never ingested. Six of our channels froze exactly this way, each at its own density fixed point.
- Each tick re-enqueues the channel-continuation job under the same job_key, and `enqueue_backfill_job`'s unconditional `ON CONFLICT` overwrite resets `status` to `pending`, zeroes `attempt_count`, wipes `last_error`, and restores the **stale cursor** — erasing the backfill worker's drain progress seconds after it saves. Live evidence: a continuation job pending for 2 weeks with `attempt_count=0`, `last_started_at` set hourly, `last_completed_at` never, its cursor pointing at the freeze date; the backfill run "completed" hourly upserting exactly the same 656 messages every run.

**Fixes (three small diffs):**

- **Head probe + monotonic watermark** (`sync.py`): when the window page has `has_more` and the run isn't `latest`-bounded, issue one extra `_sync_etl_channel_history` call with no `oldest`/watermark and `lookback_days=0` — a default, unambiguously newest-first fetch — and upsert both pages (idempotent). The checkpoint watermark becomes the numeric max of window page, head page, and previous checkpoint, so it can never regress. Correct under either anchoring behavior; the middle of the backlog is still drained by the continuation job.
- **`refresh_pending=False`** on `enqueue_backfill_job` for periodic enqueuers (sync's continuation + thread refreshes): the conflict update then only reopens `completed`/`failed` jobs (or just `failed` when combined with `refresh_completed=False`) and never rewrites the payload/status/attempts of a `pending`/`running` row. The backfill worker's own progress-save keeps the unconditional default. Continuations now drain to the live head and complete (`latest=None` drains terminate on their own).
- **Stale-`running` rescue** in `claim_backfill_jobs`: the sync clobber accidentally rescued jobs orphaned by a crashed worker; now that in-flight rows are protected, reclaim `running` jobs whose `last_started_at` is older than 2 hours.

No schema changes. Existing frozen checkpoints self-heal on the first post-deploy tick (the head probe jumps the watermark to the live head), and re-ingesting historical pages repairs `raw_payload`/`text` through the existing upsert (`raw_payload = EXCLUDED.raw_payload`).

## Tests

- Serializer: attachments/blocks pass-through + fallback text synthesis; user text never overridden; files-vs-attachments non-conflation regression (`test_shared_attachments.py`).
- Enqueue guard matrix for `refresh_pending` × `refresh_completed`; stale-`running` claim clause; continuation drain advances the cursor chain and completes without requeue (`test_shared_attachments.py`).
- Sync handler: head probe fires only on overflow and never on `latest`-bounded runs; both pages upserted; watermark takes the head value and never regresses below the checkpoint; continuation/thread-refresh enqueues carry `refresh_pending=False` (`test_sync_head_probe.py`).

All 34 tests in `workflows/slack/tests/` pass.

### DIFF
--- a/workflows/slack/backfill.py
+++ b/workflows/slack/backfill.py
@@ -55,6 +55,7 @@ from workflows.slack.shared import (
     record_run_finish,
     record_run_start,
     replace_thread_replies,
+    touch_backfill_job_started,
     upsert_messages,
     workflow_run_id_to_sync_run_id,
 )
@@ -295,6 +296,7 @@ async def handler(inp: Input, ctx: WorkflowContext) -> dict[str, Any]:
         job_type = str(job.get("job_type") or "backfill")
         record_slack_retention_backfill_job(job_type, "claimed")
         try:
+            await touch_backfill_job_started(ctx._pool, job_id)
             if job_type == BACKFILL_JOB_THREAD_REFRESH:
                 payload = _thread_refresh_payload(job)
                 thread_ts = str(payload["thread_ts"])
--- a/workflows/slack/backfill.py
+++ b/workflows/slack/backfill.py
@@ -444,6 +444,7 @@ async def handler(inp: Input, ctx: WorkflowContext) -> dict[str, Any]:
                         run_id=run_id,
                         priority=200,
                         refresh_completed=False,
+                        refresh_pending=False,
                     )
                     record_etl_items_enqueued(
                         "slack", "channel", "thread_refresh_job", 1
--- a/workflows/slack/shared.py
+++ b/workflows/slack/shared.py
@@ -29,6 +29,9 @@ BACKFILL_JOB_CHANNEL_CONTINUATION = "channel_continuation"
 BACKFILL_JOB_CHANNEL_BOOTSTRAP = "channel_bootstrap"
 BACKFILL_JOB_THREAD_REFRESH = "thread_refresh"
 BACKFILL_JOB_PAYLOAD_VERSION = 1
+# Reclaim jobs orphaned in `running` by a dead worker after this long. Must
+# exceed the longest plausible backfill workflow run.
+BACKFILL_JOB_STALE_RUNNING_HOURS = 2
 BACKFILL_JOB_TYPES = (
     BACKFILL_JOB_CHANNEL_BOOTSTRAP,
     BACKFILL_JOB_CHANNEL_CONTINUATION,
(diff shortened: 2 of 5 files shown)


lyoungblood:
Note on CI: the `build (…)` failures are the fork-PR token limitation (`permission_denied: Invalid token` in the GITHUB_TOKEN permissions step, across all images including ones this PR doesn't touch); the path-filtered test jobs skipped since only `workflows/slack/*.py` changed. All 34 tests in `workflows/slack/tests/` pass locally on this branch.

lyoungblood:
**Live validation + corrections to the description above.** The fix has been running on our production deployment (via a fork pin) and both defects are confirmed fixed — but a couple of the impact claims in the description are overstated, so correcting the record here.

**Bug 2 (busy-channel deadlock) was much narrower than "six channels frozen for 2–5 weeks."** Re-checking against live data, only **one** channel — our GitHub-bot channel (hundreds of messages per 3-day window) — was actually deadlocked. The other five I'd originally flagged were simply *quiet*: their watermark equals their last real message, and stale watermark age is not a freeze signal (that was my diagnostic error). The deadlock is real, but it only triggers once a channel's density crosses the fixed point (> `limit` messages per `thread_lookback_days` window), so it's channel-specific rather than the broad freeze the description implies. On deploy, that one channel's watermark jumped from May 28 to the live head (head probe) and its continuation drained the ~5-week backlog over a few worker runs — cursor advancing with `attempt_count` incrementing normally, i.e. the enqueue-clobber guard and stale-`running` reclaim verified end-to-end. 985/985 messages since May 25 now carry content.

**Bug 1 (content loss) is the genuinely broad one — and it's independent of the deadlock.** It affects every bot/app channel that posts attachment-only messages, whether or not that channel is busy enough to deadlock. Confirmed on two so far: the GitHub-bot channel (~985 rows) and a separate, low-volume Mamo GitHub channel (118 of 146 rows empty). Re-ingesting the Mamo channel refilled all 146 rows with content and produced **zero new rows** — confirming those were content-stripped in place by the serializer, not a sync gap. So the two bugs have very different blast radii: content loss is widespread; the deadlock is rare but total when density crosses the threshold.

Net: the fixes are validated in production, but the description's "six busy channels silently frozen for 2–5 weeks" should read as **one deadlocked channel + broad attachment content-loss across bot channels**. I've left the description body as-is to preserve history — happy to edit it too if you'd prefer the top of the PR corrected directly.

akshaan (approved):

PR #895, opened 2026-07-03
feat: opt-in pond session archive sync + pond-recall tool

tenequm:
## Summary

Sandboxes run Claude Code / Codex / pi, and the entrypoint already persists their session state dirs (`~/.claude`, `~/.codex`) into `$STATE_DIR` - but nothing reads them: the harness-native session logs (full reasoning, tool bodies, resumable transcripts) are unsearchable and die with their thread. The Postgres event trail keeps Centaur's rendering of each turn, not the harness's own record.

This adds an opt-in path that archives those logs into a shared [pond](https://github.com/tenequm/pond) store and gives every agent cross-thread recall over them:

- **Dockerfile**: install the pond binary (single static Rust binary, checksummed GitHub release, same install pattern as kubectl/nushell/amp).
- **entrypoint.sh**: a `POND_SYNC_ENABLED`-gated background sync loop (same shape as `repo-cache-watch`). Enables the claude-code/codex-cli/pi adapters and runs `pond sync -q` every `POND_SYNC_INTERVAL_SECONDS` (default 300). **Default off - zero behavior change when unset.**
- **tools/infra/pond**: `pond-recall` CLI shim (search / session / message / status), a thin wrapper over the binary, following the grafana tool structure. With it an agent answers "how did we solve this before?" / "what did we decide about X last month?" from the whole team's session history.
- **.agents/skills/pond-recall**: recall guidance, including an absence-honesty rule (a weak result is not "this was never discussed").

Configuration is pure sandbox env (`sandbox.extraEnv`), documented in `tools/infra/pond/.env.example` - no chart changes, no new in-cluster services. The store is any S3-compatible bucket (`s3://` or `s3+https://host/bucket/prefix` for MinIO/R2/etc.). Concurrent sandboxes writing one store is pond's supported optimistic-concurrency multi-writer mode (Lance manifest OCC, no coordinator).

Disclosure: I'm the pond author. Pond is Apache-2.0, and this integration only shells out to the binary - no library dependency.

## Testing

- Extracted the entrypoint pond block verbatim and ran it in a sandboxed `$HOME` with real fixture session files: adapters enabled, first background sync ingested 12 sessions / 1,574 messages into a fresh store; enable is idempotent (warm-pod restart) and an empty-source sync is an instant no-op.
- `pond-recall` client exercised against that store: search, whole-session transcript (incl. `--from-end` paging), single-message detail, status.
- `uv run pytest tools/infra/pond/tests/` - 6 passed.
- `ruff check` / `ruff format --check` clean under `tools/ruff.toml`.
- `bash -n services/sandbox/entrypoint.sh` clean.
- Not yet run: e2e on a live k8s deployment (I don't operate one) - happy to iterate if someone can point this at a staging cluster.

## Open questions for maintainers

- [ ] **Egress**: pond signs S3 requests itself (SigV4), so bucket creds can't be placeholder-swapped by iron-proxy - they arrive as plain env (`POND_CREDS_*`), same class as `GITHUB_TOKEN` today. Bucket-host egress needs granting at the deployment level; is a per-deployment proxy allowlist entry the right home, or should this ride a capability label like #884/#858?
- [ ] **Semantic search**: no embedding model runs in-sandbox (search is full-text). A central `pond optimize` (e.g. a CronJob) can embed the backlog and upgrade search to hybrid later - worth wiring in this PR or a follow-up?
- [ ] **Project attribution**: all sandboxes share the same workspace path, so pond's per-project filter collapses to one project for now. Fixable by stamping the thread key at sync time if recall filtered by thread/channel proves wanted.

## Notes

- pi/amp caveat: pi session dirs aren't currently persisted to `$STATE_DIR` (only `~/.claude`/`~/.codex` are symlinked), so pi sessions archive only within a pod's lifetime; amp has no pond adapter yet.
- The sync loop and the tool degrade cleanly: unset env means the loop never starts, and `pond-recall` returns a clear "not configured" error.

### DIFF
--- a/.agents/skills/pond-recall/SKILL.md
+++ b/.agents/skills/pond-recall/SKILL.md
@@ -1,0 +1,45 @@
+---
+name: pond-recall
+description: "Recall past agent work from the shared pond session archive. Use when a question references earlier threads or past work ('how did we solve X before', 'what did we decide about Y', 'where did we leave off'), before re-deriving a solution another session likely already found, or to verify a claim about what was previously done."
+---
+
+# pond-recall
+
+Every sandbox continuously archives its harness-native session logs (Claude
+Code, Codex, pi) into one shared pond store. `pond-recall` searches that
+archive and retrieves full transcripts, so past work is looked up instead of
+re-derived. Availability depends on the deployment enabling pond sync; if
+`pond-recall status` errors, the archive is not configured and this skill does
+not apply.
+
+## Workflow
+
+1. **Search first.** Keep the query semantic (concepts, error text, feature
+   names). Scope with filters, not the query:
+
+   ```
+   pond-recall search "lance manifest retention window" --limit 5
+   pond-recall search "billing tests failing" --from-date 2026-06-01 --sort-by recency
+   ```
+
+2. **Read what a hit points at.** Hits carry `message_id` and `session` ids:
+
+   ```
+   pond-recall message <message_id>          # full detail incl. tool bodies
+   pond-recall session <session_id> --from-end   # how that thread ended
+   ```
+
+3. **Page long sessions** with `--after-message-id <last id shown>`.
+
+## Judgment rules
+
+- A zero or weak result is not proof of absence: the index covers
+  conversational text only (tool output and reasoning are excluded by design),
+  and the archive only reaches back to when the deployment enabled sync.
+  Say "the archive has nothing on this", never "this was never discussed".
+- Scores are relative within one response; do not compare across queries.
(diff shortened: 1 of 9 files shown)


mslipper:
Hello! Thanks for the contribution.

I'm going to close this for now. Even behind a flag, this puts a third party binary in the Dockerfile and a custom sync loop in `entrypoint.sh` which is the boot path for every sandbox. That introduces a maintenance burden we can't absorb right now. Happy to revisit in the future if there's community demand.

PR #899, opened 2026-07-05
docs: update Codex and Claude auth token setup

richardwu:
## Summary
- replace the old dedicated-account guidance for Codex and Claude access-token setup with fresh-token instructions
- document Codex device-code login setup and `codex login --device-auth`, including the OpenAI docs link
- document `claude auth-token` for Claude and warn not to reuse tokens from local `~/.codex` or `~/.claude` credentials because refreshes can invalidate each other
- keep the public markdown mirror in sync with the MDX source

## Testing
- `git diff --check`
- `npm ci --ignore-scripts` in `docs/`
- `npm_config_ignore_scripts=true npm run build` in `docs/` (passes; emits existing dead-link warning for `/centaur-brand-assets.zip`)

### DIFF
--- a/docs/pages/deploying-in-production.mdx
+++ b/docs/pages/deploying-in-production.mdx
@@ -106,14 +106,14 @@ accordingly. See [Security](/security) for the full threat model.
 
 ### Codex Auth Modes
 
-:::warning[Dedicate the account to Centaur]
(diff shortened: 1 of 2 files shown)


richardwu:
Just realized Centaur doesn't support OAuth tokens atm (just refresh token blob), will fix this later.

PR #906, opened 2026-07-06
Fix slackbotv2 handoff wedges

kamsz:
## Summary

- Add a foreground Slack webhook handoff timeout so captured handoff promises cannot keep HTTP requests pending indefinitely.
- Track `activeExecutionStartedAt` for Slackbot thread state and treat old or timestamp-less active-execution markers as stale.
- Clear or refresh the active-execution timestamp consistently across live render, retry, recovery, stale recovery, and late-file idle checks.
- Add emulator coverage for stale active-execution recovery and bounded webhook handoff retries.

## Root Cause

`slackbotv2` waited on captured Chat SDK handoff promises with an unbounded `Promise.all`. If a handoff task wedged, the Slack webhook request stayed open until the process restarted. Separately, Slackbot persisted `activeExecution: true` without a timestamp/TTL, so a crash or stalled render path could make a thread look active forever.

## Validation

- `pnpm --filter slackbotv2 exec bun test test/chat-sdk-emulate.test.ts`
- `pnpm --filter slackbotv2 run check:types`
- `pnpm --filter slackbotv2 test`

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -108,6 +108,8 @@ type SlackbotV2RequestContext = {
 }
 
 const requestContext = new AsyncLocalStorage<SlackbotV2RequestContext>()
+const ACTIVE_EXECUTION_TTL_MS = 30 * 60 * 1000
+const WEBHOOK_HANDOFF_TIMEOUT_MS = 30_000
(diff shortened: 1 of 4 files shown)


mslipper:
I believe this has been fixed. Please reopen if not.

PR #910, opened 2026-07-06
fix(harness-server): don't hold the stderr lock across the harness child's lifetime (deadlocks claudecode turn completion)

aadityakbh:
## Symptom

On a self-hosted deployment at head with the **claudecode** harness, every turn streams its full final answer and then hangs: no `turn/completed` is emitted, the api-rs execution stays `running` indefinitely, and the answer only reaches Slack after the sandbox dies and render recovery fails the execution over. #903's diagnostic line (`event=harness_turn_completed_on_assistant_end_turn`) never appears even though the fallback's condition is met.

## Root cause

`run_harness_turn`'s child-stderr copy thread holds the global stderr lock for the **lifetime of the harness process**:

```rust
std::thread::spawn(move || {
    let mut parent_stderr = io::stderr().lock();   // held until child stderr EOF
    let _ = io::copy(&mut stderr, &mut parent_stderr);
});
```

The claude CLI process outlives each turn — the server keeps it (and its stdin) in `ThreadState` for the next turn — so its stderr never EOFs and the lock is never released. The turn-completion diagnostic `eprintln!` added in #903 then blocks on that lock, deadlocking the turn at the exact moment it completes: terminal detection has already fired (`stop_reason=end_turn`), but `TurnCompleted` is never written.

Why no existing test catches it: every fake harness in `app_server_stdio.rs` is a `printf`/`cat` that **exits** after printing its stream — the EOF ends the copy, releases the lock, and the diagnostic proceeds. The failure needs a harness process that outlives the turn, which is exactly how the real CLI behaves.

## Repro / regression test

`claude_turn_completes_while_harness_process_outlives_the_turn` replays a real `claude --print --input-format stream-json --output-format stream-json --verbose --include-partial-messages` stream (CLI 2.1.154, the version in the current `centaur-agent` image; captured live, host-specific lines sanitized) with the fake kept alive via `sleep 60`:

- on current `main`: times out after 15s (the deadlock)
- with this fix: completes in milliseconds, alongside the existing claude tests

## Fix

Copy through the unlocked `Stderr` handle, which locks per write. One line; interleaving granularity of child-stderr lines vs. the server's own is unchanged in practice (both are line-buffered writes).

## Impact

Any deployment running the claudecode harness: all turns hang at completion and deliver only via sandbox-death failover. Codex deployments are exposed to the same hazard through any future `eprintln!` on a hot path while a harness child is alive.


mslipper:
Thanks for this! Merged it in as part of #921.

PR #926, opened 2026-07-07
fix(slackbotv2): release the events connection when a live render is abandoned mid-turn

aadityakbh:
## Problem

#920 releases the `GET /api/session/{key}/events` connection when its consumer stops — but only through async-iterator teardown, and teardown cannot fire mid-execution. Every layer of the live render chain is parked on an internal await (the conflate pump on `iterator.next()`, the SSE parser on `reader.read()`), and a queued `.return()` waits for the pending operation to settle — which only the stream's next event does. Once the render is abandoned, that event never comes.

So when `adapter.stream` fails mid-render (Slack stream expiry, `msg_too_long`, rate limit, network error), the abandoned render's connection stays open until the execution's terminal event:

- one slot of Bun's global fetch pool (`BUN_CONFIG_MAX_HTTP_REQUESTS`, default 256) held for the rest of the turn,
- the conflate pump keeps pulling and buffering every remaining chunk with no consumer — the tens-of-thousands-of-chunks volume conflation exists to absorb,
- the retry and recovery paths open fresh streams while the old ones are still parked, so repeated Slack failures accumulate held connections — the same accumulation shape as the 2026-07-06 incident.

There is also a propagation break above all this: `streamAfterFirstChunk` drives the conflated stream with a manual `iterator.next()` loop and no try/finally, so a consumer close never even reaches the conflate layer. (Contrast `renderFallbackFinalAnswer`, which drains the identical chain with plain `for await` and closes cleanly.)

## Change

- **session-api**: the events stream now carries an out-of-band `close()` that cancels its reader, settling the parked read immediately instead of waiting for the next event. A read settled by `close()` is recorded as `reason="cancelled"`, not `done` — the execution didn't finish, the consumer let go.
- **index**: a failed render attempt releases its events connection first thing in the catch — the fallback that follows blocks on a fresh stream until the terminal event, which is exactly how long the dead connection would otherwise stay parked. The recovery render closes its stream the same way, and `streamAfterFirstChunk` now forwards protocol closes to the layers below instead of silently dropping them.

## Testing

- New always-on regression test: a mid-render Slack stream failure while the execution is still emitting output must close the abandoned connection with `reason="cancelled"` *before* the terminal event, and the open-streams gauge must settle at baseline afterward. Fails on `main` (times out waiting for the cancelled closure), passes with this change.
- `bun test test`: 147 pass / 0 fail.
- `bun run check:types` clean.

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -38,7 +38,8 @@ import {
   serializeMessageLinks,
   serializeMessage,
   sessionStreamError,
-  withSlackApiTimeout
+  withSlackApiTimeout,
+  type SessionEventStream
 } from './session-api'
 import {
   buildConsoleSessionContextBlock,
@@ -1057,10 +1058,11 @@ async function renderExecutionAttempt(
   let rendered = false
   let retry = false
   let fallbackLastEventId = 0
+  const eventsConnection: { close?: () => void } = {}
   try {
     const streamResult = await renderExecutionStream(
       thread,
-      streamSessionAfterHandoff(options, input),
+      streamSessionAfterHandoff(options, input, eventsConnection),
       message,
       options,
       trace,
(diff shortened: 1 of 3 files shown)


aadityakbh:
Closing this to stop it sitting as open-but-dead. Five weeks with no review or comment isn't a judgement on the fix — it's a throughput reality, and leaving it open misrepresents it as under consideration.

The branch stays on our fork, so this is one click to reopen if it ever becomes useful to you. Happy to rebase it then.

PR #927, opened 2026-07-07
fix(harness-server): keep the settle fallback armed across rejected mid-turn requests

aadityakbh:
Hardens the settle machinery from #921. A `turn/steer` or `turn/interrupt` that arrives with a stale turn id — a client racing the turn boundary — is rejected with a JSON-RPC error and never reaches the harness, but the drain loop cleared `settle_deadline` unconditionally after every request. If the turn had already stopped at a terminal assistant stop with no native `result` (the case the settle fallback exists to complete), the rejection disarmed the only remaining completion path and the turn hung with the stream quiet.

`handle_active_turn_request` now reports what it did — `Interrupted`, `Steered`, or `Rejected` — and only a delivered steer disarms the pending fallback, since only then does the harness owe a response. Rejections (stale steer, stale interrupt, unknown method) leave the turn's state untouched.

To be direct about reachability: this path only fires in jsonrpc mode, and production sandboxes run blocks mode, whose only mid-turn request kills the process. So this is a latent defect in the settle state machine plus a regression test pinning the invariant, not a live production hang.

## Testing

- New regression test: a wrong-thread steer and a stale-turn-id steer land inside the armed settle window; both must be rejected with `-32600` and the turn must still complete through the fallback. Fails on `main` (the turn never completes), passes with this change.
- `cargo test -p harness-server`: 21 pass / 0 fail (4 ignored: real-CLI tests).
- `cargo +nightly fmt --check` clean; `cargo clippy --all-targets` introduces no new warnings.

### DIFF
--- a/crates/harness-server/src/server.rs
+++ b/crates/harness-server/src/server.rs
@@ -1008,16 +1008,27 @@ fn apply_resume_overrides(state: &mut ThreadState, params: &ThreadResumeParams) 
     Ok(())
 }
 
+/// What handling a mid-turn JSON-RPC request did to the harness.
+enum ActiveTurnRequestOutcome {
+    /// The harness process was killed; the turn is over.
+    Interrupted,
+    /// Steer input was written to the harness, which now owes a response.
+    Steered,
+    /// The request was rejected without reaching the harness: the turn's
+    /// state is untouched.
+    Rejected,
+}
+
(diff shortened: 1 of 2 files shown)


aadityakbh:
Closing — this is a latent defect in a jsonrpc-mode path production doesn't exercise; not worth the review cycle over #921's live fixes.

PR #928, opened 2026-07-07
fix(workflows): keep the Drive sync watermark behind failed document fetches

aadityakbh:
## Problem

`google_drive_sync` catches a per-file `docs_get_text`/upsert failure, records it (`last_error`, metrics, `partial_failed` status), and moves on — but the scope checkpoint still advances to the max `modifiedTime` of the files that *succeeded*. The next run's query is `modifiedTime > (watermark - overlap)`, so a failed file whose `modifiedTime` sits below the new watermark is excluded from every future incremental sync.

Nothing else retries it: the incremental query is the only re-fetch path, and rows with `last_error != ''` are excluded from the `company_context_documents` projection. So one transient 429 during a sync and that document is permanently missing from company context until someone happens to edit it (bumping its `modifiedTime` back above the watermark). The `partial_failed` run status reports the failure, but the advanced checkpoint means the failure is not recoverable.

Concretely: a run sees doc B (modified 10:00, fetch fails on a transient 429) and doc A (modified 12:00, succeeds). The checkpoint advances to 12:00; the next run queries `> 11:59` and B is gone for good.

## Change

The watermark no longer advances past an unresolved failure:

- Clamp the checkpoint to the earliest failed file's `modifiedTime`, so the next run's query re-includes it. The succeeded files it re-reads upsert idempotently (and clear `last_error` on success).
- A failure with no usable `modifiedTime` cannot be re-included by clamping, so the checkpoint writes `NULL` — the upsert's existing `COALESCE` keeps the stored watermark, and the next run rescans the same window.

## Trade-off

Clamping means a *permanently* failing file (for example a 403 that outlives retries — sharing changes don't bump `modifiedTime`, so nothing else would ever surface it) pins the watermark and makes each run re-read the window above it. That is deliberate: bounded re-work over silent data loss. If you'd rather bound the re-work too, the complement is reconciling `google_drive_sync_files` rows with `last_error != ''` explicitly, independent of the watermark — happy to follow up with that on top of this.

## Testing

New `workflows/tests/test_drive_sync_watermark.py` (same self-bootstrapping harness as the existing workflow test), driving `handler` end-to-end with a fake Drive client and pool:

- earliest-failure clamp: mixed success/failure run must checkpoint at the failed file's `modifiedTime`, not the succeeded max — fails on `main` (checkpoints at 12:00), passes here;
- no-`modifiedTime` failure must write `NULL` (keep the stored watermark) — fails on `main`, passes here;
- all-success run still advances to the max `modifiedTime` (passes on both).

`workflows/tests`: 6 passed.


akshaan (approved):

aadityakbh:
Withdrawing — we don't run the Drive sync path ourselves; the patch and its failing-on-HEAD tests stand if they're useful.

PR #929, opened 2026-07-07
fix(session-runtime): release the stdout-owner lease when the pump fails

aadityakbh:
## Problem

When `run_stdout_pump` returns an error — a transient Postgres failure on any per-line write (`append_output_line`, `active_execution_for_thread`, `append_event`) is enough — the pump loop logs, appends `session.stdout_pump_failed`, and breaks. But the fire-and-forget lease renewer keeps renewing this process's stdout-owner claim every 10s for as long as the row stays `running` and the process lives, which after the pump's death is indefinitely.

With the lease held, every recovery path is blind to the execution: this process's own adoption scan takes the *own lease* fast path and skips it, and peers see an active lease and defer. So a still-running execution records no further output and reaches no terminal event until its max-duration budget burns down (`DEFAULT_AGENT_MAX_DURATION_MS`, 30 minutes on the agent turn path) — a half-hour ghost in the thread, ending in a max-duration failure, for an answer that was recoverable from the sandbox backend's recorded output the whole time.

The asymmetry that makes this reachable: the pump does a DB write per output line while the renewer writes once per 10s, so a brief DB blip is far more likely to kill the pump than the renewer.

## Change

Release the stdout-owner lease on the pump-failure path, mirroring what the shutdown handoff (#872) does deliberately: the renewer stops on its next tick (`renew_stdout_owner` no longer matches), and the next adoption scan (#871) reclaims the execution immediately instead of waiting out the lease — recovering its output rather than losing it. A `session.stdout_owner_released` event with `reason: "stdout_pump_failed"` announces the handoff; event consumers whitelist known types (verified for #872), so it is inert downstream.

If this process no longer owns the lease (the execution was already adopted elsewhere), the release is a no-op and nothing is announced.

## Testing

DB-gated (ParadeDB), alongside the existing adoption tests:

- pump-failure release: a running execution owned by this process → after the release, the renewer's next tick returns false and a peer's adoption claim succeeds immediately (no 45s TTL wait); the `session.stdout_owner_released` event carries `reason: "stdout_pump_failed"`.
- non-owner: a pump failure in a process that lost the lease must not touch the new owner's lease or announce a release.

`cargo test -p centaur-session-runtime adoption_tests`: 25 passed; the 3–4 failures are the pre-existing timing-sensitive `stdout_eof_*`/`workflow_cleanup_*` ones that reproduce identically on unmodified `main` against a local ParadeDB (the same set noted in #920). `cargo +nightly fmt --check` clean; `cargo clippy -p centaur-session-runtime --all-targets` clean.

### DIFF
--- a/services/api-rs/crates/centaur-session-runtime/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-runtime/src/lib.rs
@@ -3836,6 +3836,80 @@ fn remove_pipe_if_current(sandbox_pipes: &SessionPipeMap, sandbox_id: &str, pipe
     });
 }
 
+/// Hands a still-active execution to the adoption scan after its pump died.
+///
+/// The pump is gone, but the fire-and-forget lease renewer keeps this
+/// process's stdout-owner claim alive for as long as the process lives. With
+/// the lease active, this process's own adoption scan skips the execution
+/// (own lease) and peers defer to it — so a running execution whose pump
+/// failed would record no further output and reach no terminal event until
+/// its max-duration budget burns down. Releasing the lease lets the next
+/// adoption scan reclaim the execution and recover its output from the
+/// sandbox backend instead.
+async fn release_stdout_owner_after_pump_failure(
+    store: &PgSessionStore,
+    stdout_owner_id: &str,
+    thread_key: &ThreadKey,
+    sandbox_id: &str,
+) {
+    let execution = match store.active_execution_for_thread(thread_key).await {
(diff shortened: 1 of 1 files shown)


aadityakbh:
Closing this to stop it sitting as open-but-dead. Five weeks with no review or comment isn't a judgement on the fix — it's a throughput reality, and leaving it open misrepresents it as under consideration.

The branch stays on our fork, so this is one click to reopen if it ever becomes useful to you. Happy to rebase it then.

PR #950, opened 2026-07-07
fix(slackbotv2): compose personas with harness flags

svc-paradigm:
## Summary
- parse persona directives independently from harness/model flags
- forward `persona_id` during session creation so runtime persona prompts/env apply across harnesses
- keep persona sticky in Slack threads and cover `--invest --claude --model=fable` with regression tests

## Tests
- `bun test services/slackbotv2/test/overrides.test.ts services/slackbotv2/test/session-api.test.ts services/slackbotv2/test/chat-sdk-emulate.test.ts`
- `bun run --cwd services/slackbotv2 check:types`

Prompted by: arjun

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -142,7 +142,10 @@ type PendingLateSlackFileMention = {
   user: string
 }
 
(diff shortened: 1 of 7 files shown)


mslipper:
We need to rethink personas - closing for the time being as we redesign this feature.

PR #984, opened 2026-07-08
feat(console): chat composer — start and continue chats from the console

goksu:
## What

Adds a chat composer to the Console threads view, so chats can be started and continued directly from the Console instead of only via Slack.

- **New chat**: a sidebar tab (above Chats, standard tab geometry/spacing) opens a full-page composer — heading + input box sitting slightly above vertical center, at the transcript column width. The no-selection / no-chats states render this screen directly; only an unresolvable `?thread=` still shows not-found.
- **Model + effort picker**: a two-level menu (codex-app style) — root rows show Model and Effort with current values; submenus check the active choice. Each model pins its harness (Claude Opus/Sonnet/Haiku/Fable → `claudecode`, GPT-5.6 Sol / GPT-5.5 → `codex`, and Amp as a plain entry with no model, since Amp picks its own per turn). GPT-5.6 Sol is the default. Reasoning efforts are per-model (`max` is 5.6-only), apply to codex picks only, and ride the blocks line as `reasoning` — Default sends none so the harness config decides. The popover reuses the account dropdown's panel/item classes and posts through hidden `model`/`effort` fields.
- **Follow-ups**: open chats and split-view panes get a docked composer (prompt only; meta line shows the chat's harness · model). Enter sends, Shift+Enter breaks a line. While a turn is running the page re-visits every 4s — skipped whenever the composer holds a draft.
- **Optimistic sidebar**: a chat the turbo-permanent sidebar list has never seen gets an inserted row immediately (title from the thread header, "now"), then the frame refetches so the server row/ordering replaces it.

## Backend

`POST /console/threads` (previously a read-only stub) now drives api-rs through the existing `CentaurApiClient` — chats are writable unconditionally; the old read-only mode and its `CENTAUR_CONSOLE_THREADS_READ_ONLY` env are removed:

- New thread: `console:<uuid>` key → `create_session` (metadata `platform`/`source=console` + `user_email`/`actor_email`, matching the existing owner scope) → `append_session_messages` → `execute_session` with a blocks-protocol user line carrying the chosen model (omitted for Amp). The same UUID rides as `client_message_id` / `client_user_message_id` so api-rs dedupes the harness echo.
- Reply: the thread key resolves through `visible_thread_scope`, so nobody can post into another user's chat; the recorded model (execution metadata → session metadata → deploy default) is reused. Split-view panes survive the redirect via an `open_threads` field.
- The client is injectable (`client_factory`), mirroring the workflows controller. Note for deployers: the console needs `CENTAUR_CONSOLE_CENTAUR_API_URL` pointed at a live api-rs; posting into a snapshot-only deployment will surface API errors as flash alerts.

## Testing

```
docker exec codex-centaur-console-web bin/rails test test/controllers/console/threads_controller_test.rb
```

68 runs green (composer rendering, start/reply/Amp/unknown-model/error paths, owner-scope rejection, DB-unavailable fallback to the new-chat screen); full controllers suite 577/577; rubocop clean. Verified end-to-end in a browser against a local api-rs (thread started from the composer, reply ran on the picked model/harness, both themes screenshotted).

### DIFF
--- a/services/console/app/controllers/application_controller.rb
+++ b/services/console/app/controllers/application_controller.rb
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
 
   helper_method :current_user, :acting_admin?, :descoped?
   helper_method :public_base_url, :oauth_callback_redirect_uri
+  helper_method :console_threads_read_only?
 
   # The public origin the console is reached at. Derived from the request by
   # default; CENTAUR_CONSOLE_PUBLIC_URL overrides it for deployments behind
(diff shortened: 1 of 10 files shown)


gakonst:
Super excited for this

PR #988, opened 2026-07-09
fix(slack): populate channel member emails

Osraka:
## Summary
- Enrich Slack channel members with profile data from `users.list` keyed by the IDs returned by `conversations.members`.
- Preserve the existing fallback when profile lookup is unavailable.
- Add regression coverage for `get_channel_member_emails`, which backs the `channel-emails` CLI used by downstream GSuite permission helpers.

## Why
`get_channel_member_emails()` expected `get_channel_members()` entries to include `email`, but `get_channel_members()` only returned `id` and `name` from the user-name cache. That made email export return an empty list even when Slack user profiles included emails.

## Tests
- `python3 -m ruff check tools/productivity/slack/client.py tools/productivity/slack/tests/test_client.py`
- `PYTHONPATH=tools/productivity /tmp/centaur-slack-venv/bin/python -m pytest tools/productivity/slack/tests/test_client.py -q`
- `python3 -m py_compile tools/productivity/slack/client.py tools/productivity/slack/tests/test_client.py`

### DIFF
--- a/tools/productivity/slack/client.py
+++ b/tools/productivity/slack/client.py
@@ -1261,6 +1261,49 @@ class SlackClient:
 
         return sorted(users[:limit], key=lambda x: x["name"])
 
+    def _users_by_id(self, user_ids: list[str]) -> dict[str, dict[str, Any]]:
+        """Fetch Slack user profiles for a known set of user IDs."""
+        wanted = {user_id for user_id in user_ids if user_id}
+        users: dict[str, dict[str, Any]] = {}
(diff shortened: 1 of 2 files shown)


mslipper:
Thanks for looking into this. I don’t think we should merge this implementation as-is. It scans the full workspace user list on every channel lookup, can silently return partial results after an API or rate-limit failure, and includes disabled or bot accounts when exporting emails. Given the age of this PR, I'm going to close it for now.

PR #997, opened 2026-07-09
docs: document access-token broker credential bootstrap

akandic47:
Closes #995.

Documents the executable Codex and Claude access-token bootstrap flow, including operator CLI setup, credential extraction for file and macOS keychain storage, and the required broker commands. Corrects the Helm auth-mode configuration reference and points the Kubernetes secret bootstrap helper to the out-of-band broker step. The branch includes the latest main and leaves api-rs runtime behavior unchanged.

PR #1000, opened 2026-07-09
ci(docs): fail the build when public/md mirrors drift from pages; resync

### LINKED ISSUE #999: docs: committed public/md mirrors silently drift from their mdx sources — no CI verification
`docs/public/md/*.md` are committed copies of `docs/pages/*.mdx`, kept in the repo "so agents can curl them" (`docs/scripts/build-md.sh`). The Docs workflow regenerates them on every build via the `prebuild` hook before deploying, so the published site is always current — but the regenerated files are never committed back, and nothing verifies the committed copies. Staying in sync relies on each author remembering to rerun `build-md.sh` and commit.

Predictably, they drift: on current main a fresh `build-md.sh` run dirties **eight** mirrors (`extend/acme-example`, `extend/apps`, `operate/slack-etl`, `reference/tool-directory`, `reference/configuration`, `secrets/environment`, `secrets/onepassword`, `what-is-centaur`). For example, `secrets/onepassword.md` still tells readers to exec into `deploy/centaur-centaur-api`, which the mdx long since corrected to `deploy/centaur-centaur-api-rs`. Agents reading raw md from the repo get outdated instructions while the website looks fine — which is why nobody notices.

Proposal: add a drift check to the Docs workflow (it already triggers on `docs/**` for PRs and main) — run `build-md.sh`, fail on `git diff --exit-code -- public/md` with a rerun-and-commit message. The script is a plain `cp` loop, so the check costs nothing. Resync the eight stale mirrors together with the check so it starts green.

(#996 mentioned this staleness in passing; this issue tracks the resync + verification specifically.)

akandic47:
Closes #999

The committed `docs/public/md` mirrors only stay in sync with `docs/pages` when authors remember to rerun `build-md.sh`. The Docs workflow regenerates them at build time for the deployed site but never verifies the committed copies, so drift is invisible on the website and accumulates in the repo — eight files on current main.

- Add a step to the Docs workflow (before node setup; the script is a plain `cp` loop needing no deps): run `scripts/build-md.sh`, fail on `git diff --exit-code -- public/md` with a message telling the author to rerun the script and commit.
- Resync the eight currently stale mirrors in the same commit so the check starts green. Pure `mdx → md` copies; no source content changes.

The workflow already triggers on `docs/**` for both PRs and pushes to main, so the check runs exactly where mirror edits can land. Verified locally: the check fails on unmodified main and passes after the resync.

Note: #997 / #998 also touch mdx pages and their mirrors; whichever lands after this needs at most a trivial `build-md.sh` rerun — and this check will catch it if forgotten.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,6 +45,17 @@ jobs:
         with:
           persist-credentials: false
 
+      # The build regenerates public/md on the fly, so the deployed site never
+      # drifts — only the committed copies agents curl from the repo do.
+      - name: Check committed md mirrors are in sync
+        run: |
+          bash scripts/build-md.sh
+          if ! git diff --exit-code --stat -- public/md; then
+            echo "docs/public/md is out of sync with docs/pages;" \
+              "run 'bash docs/scripts/build-md.sh' and commit the result." >&2
+            exit 1
+          fi
+
       - name: Set up Node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -50,6 +50,9 @@ jobs:
       - name: Check committed md mirrors are in sync
         run: |
           bash scripts/build-md.sh
+          # Intent-add so mirrors for newly added pages (untracked files)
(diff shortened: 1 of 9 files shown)


mslipper:
Thanks for the contribution! `public/md` files are now generated dynamically rather than committed to the repo, so this PR is no longer necessary. Closing.

PR #1018, opened 2026-07-10
fix: detect bot mentions in Slack attachments

decofe:
## What changed

- Detect Centaur mentions in legacy Slack `attachments[].text`, not only top-level message text.
- Include attachment text in the message sent to the agent so alert context is preserved.
- Add regression coverage for attachment-only VictoriaMetrics bot alerts.

## Why

Alertmanager/VictoriaMetrics posts alerts as attachment-only bot messages. The alert template already included `<@U0ANX3AM5RR> investigate this alert`, but Slack delivered an empty top-level `event.text`, so Centaur did not recognize or execute the mention.

## Validation

- `pnpm --filter slackbotv2 exec bun test test/chat-sdk-emulate.test.ts`

Prompted by: @Zygimantass


decofe:
Closing per requester — this should be implemented in the alert configuration layer instead.

PR #1019, opened 2026-07-10
fix: restore direct sandbox OTLP egress

Zygimantass:
## Summary

- restore endpoint-derived OTLP egress on observability-enabled sandbox NetworkPolicies
- keep the matching proxy rule while preserving the restricted-sandbox boundary
- update the config contract and regression test to require direct collector access

## Root cause

PR #864 removed the direct sandbox-to-collector rule under the assumption that a chart-level label policy would replace it. Harness OTLP exporters bypass the HTTP proxy via `NO_PROXY`, and production had no matching Laminar destination in the replacement policy, so model spans stopped at the rollout boundary while api-rs spans continued.

## Validation

- `cargo test -p centaur-sandbox-agent-k8s` (43 passed)
- `cargo fmt --all -- --check`
- `git diff --check`
- `just build-one api-rs`
- deployed the built image to the local OrbStack Kubernetes stack
- real Codex session completed with `OTLP_OK` and emitted token usage
- generated sandbox NetworkPolicy contained the configured collector namespace and port rule

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/args.rs
+++ b/services/api-rs/crates/centaur-api-server/src/args.rs
@@ -1374,8 +1374,8 @@ impl TryFrom<&SandboxArgs> for AgentSandboxConfig {
         }
         config.iron_control = args.iron_control.settings();
         config.tools = args.tools_source.to_config();
-        // The chart label policy handles sandbox OTLP egress; keep the
-        // per-sandbox proxy's own in-cluster OTLP egress explicit.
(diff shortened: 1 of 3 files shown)


Zygimantass:
Superseded by https://github.com/tempoxyz/prd-centaur-infra/pull/569.

Centaur #864/#898 intentionally centralized observability egress behind the `centaur.ai/observability-enabled=true` label. Restoring destination parsing in api-rs would violate that design; the actual production gap was that GitOps values never enabled the chart-managed Laminar destination.

PR #1050, opened 2026-07-13
feat(slack): read huddle transcripts — the one Slack record no bot token can reach

aadityakbh:
## The gap

Slack records every huddle and saves a full, speaker-attributed transcript. **A bot token cannot read it, and no public API exposes it.** `files.info` on the transcript's file id returns metadata and no words, whatever scopes the app holds — Slack support confirms this is deliberate.

That's the wrong hole to have. The huddle is where the standup sets the day's priorities, where the incident call finds the cause, where the design argument actually gets settled. Channels get the summary; **the huddle had the reasoning.** An agent that reads every channel and no huddle is missing the part that mattered.

## The mechanism

The Slack *web client* reads them perfectly well: `files.info` with `include_transcription=true` against the **workspace host** (`<workspace>.slack.com`, never `slack.com`), authenticated by a user web session — an `xoxc` token paired with the `d` cookie. This replays exactly that call.

**Not a scope escalation.** A user session sees precisely what that user could already read by opening the huddle in their own Slack client.

```bash
slack huddle-transcript F0123456789
```
```
34 turns, 3 speakers
<@U01ABC> [0:04]: the migration is blocked on the schema review
<@U02DEF> [0:11]: I can review it this afternoon
<@U01ABC> [0:19]: then we ship Friday
```

## Why the token rides in a header — the part I'd most like reviewed

The obvious implementation copies the browser and posts the token as a **form field**. I didn't, and the reason is your own security model: **a request body is the one place iron-proxy cannot look.** The secret schema offers `match_headers`, `match_query`, and `match_path` — there is no `match_body`. Copying the browser would have forced a real user web session to sit inside the sandbox, which is exactly the exposure `mode = "inject"` exists to prevent.

So I checked against the live API before choosing. **Slack accepts this token in the body, in the query string, *and* in the `Authorization` header — all three authenticate.** Since they're equivalent to Slack, take the one the firewall can reach:

```toml
{type = "http", name = "SLACK_WEB_TOKEN",  match_headers = ["Authorization"], hosts = ["*.slack.com"]},
{type = "http", name = "SLACK_WEB_COOKIE", match_headers = ["Cookie"],        hosts = ["*.slack.com"]},
```

The sandbox holds **placeholders only**; the real session never leaves the proxy. Same declaration shape as the existing `SLACK_SEARCH_TOKEN` / `SLACK_ETL_TOKEN`.

## Shape

- **Extends the existing `slack` tool** rather than adding a parallel one, riding the `optional_secrets` pattern already there. Without the session, every other method behaves exactly as before, and only `get_huddle_transcript` reports what's missing.
- **The workspace host is derived from `auth.test`**, not configured — the bot token already knows which workspace it belongs to. **Zero new deployment config.**
- **Speakers stay as Slack ids, never names.** An id round-trips to a real mention and can't drift; a name resolved at read time silently rots when someone changes their display name.
- **Only the rendered text is returned, not the segments as well.** They're the same words twice, and this is the largest payload the tool can produce — returning both writes the whole huddle into the model's context a second time and pays for it on every request of the turn.

## The failure mode is the design

A lapsed session raises `needs_reauth`. It **never returns an empty transcript**.

That distinction is the whole point. A silent empty read looks exactly like *"nobody spoke"* — and an agent that quietly concludes a meeting was silent is far worse than one that says it cannot see. The first fabricates an absence of evidence; the second reports a broken credential.

```json
{"error": "huddle_transcript_failed", "slack_error": "invalid_auth", "needs_reauth": true}
```

Discovery (which huddles exist) runs on the ordinary bot token and never needs the session, so nothing is lost while a session is stale — refresh it and the backlog reads fine.

## Trust boundary, stated plainly

A user web session carries **that user's full Slack visibility, DMs included.** It should be minted from an account whose visibility you're content for agents to inherit. That's called out in `.env.example` and in the docs, not buried.

## Verification

- 13 new tests on the parser and the failure modes — including the one that would silently corrupt the record: the **bold** ` [mm:ss] ` marker vs a **bold word inside the speech** (treating the latter as a marker deletes words from the transcript).
- **90 passing in the slack suite, no regression.** `ruff` clean.
- Transport verified against the live Slack API (above).
- Docs at `operate/slack-huddle-transcripts`, sidebar registered, `tool-directory` updated, `public/md` mirrors synced for the pages I touched only.
- Adds the `.env.example` the tool was missing (61 of 79 tools ship one; `slack` was an outlier, and this change adds two credentials that need documenting).

### DIFF
--- a/docs/pages/operate/slack-huddle-transcripts.mdx
+++ b/docs/pages/operate/slack-huddle-transcripts.mdx
@@ -1,0 +1,102 @@
+---
+title: Slack huddle transcripts
+description: Read the verbatim, speaker-attributed transcript of a recorded Slack huddle — the one Slack record no bot token can reach.
+---
+
+# Slack huddle transcripts
+
+Slack records every huddle and saves a full, speaker-attributed transcript. **A bot token cannot read
+it, and no public API exposes it.** `files.info` on the transcript's file id returns metadata and no
+words, whatever scopes the app holds; Slack support confirms this is deliberate.
+
+That leaves a hole in an agent's view of a workspace, and it is the wrong hole. The huddle is where
+the standup sets the day's priorities, where the incident call finds the cause, where the design
+argument is actually settled. Channels get the summary; the huddle had the reasoning. An agent that
+reads every channel and no huddle is missing the part that mattered.
+
+Centaur's Slack tool can read them, given a **user web session**.
+
+## How it works
+
+The Slack *web client* reads transcripts perfectly well. It calls `files.info` with
+`include_transcription=true` against the **workspace host** (`<workspace>.slack.com` — never
+`slack.com`), authenticated by a user web session: an `xoxc` token paired with the `d` cookie.
+The tool replays exactly that call.
+
+This is not a scope escalation and not a permission bypass. A user session sees precisely what that
+user could already read by opening the huddle in their own Slack client. If they cannot see the
+huddle, neither can this.
+
+```bash
+slack huddle-transcript F0123456789
+```
+
+```
+34 turns, 3 speakers
+<@U01ABC> [0:04]: the migration is blocked on the schema review
+<@U02DEF> [0:11]: I can review it this afternoon
+<@U01ABC> [0:19]: then we ship Friday
+```
+
+The agent-facing method is `get_huddle_transcript(file_id)`. It returns `{file_id, speakers, turns,
+text}`. Speakers stay as **Slack ids**, not names — an id round-trips to a real mention and cannot
+drift, while a name resolved at read time silently rots when someone changes their display name.
+
+Find a `file_id` from the huddle's message in `conversations.history` (the room object references its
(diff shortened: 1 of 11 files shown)


goksu (approved):
Looks directionally good! Can you rebase again with latest so we can get the CI going?

aadityakbh:
@goksu done — rebased onto latest `main`, clean, CI should be good to go. It's pretty annoying Slack doesn't let you hit the transcript directly, so this is a neat workaround.

Separately: we run a platform where institutions come to trade, invest in HFT strategies, and hedge on-chain. Built a bunch of neat tools for on-chain data analysis through the bot, which we use to manage $21M. Included it in my Paradigm fellowship application but it never received a look — would love to connect with the team and demo what we've built if interested!

goksu:
Closing as there are some gaps in the implementation and the feature is not actually needed at this time as most usage we see are on Granola for meeting notes which Centaur natively supports.

PR #1066, opened 2026-07-14
fix(workflows): Linear searchIssues takes `term`, not `query`

aadityakbh:
#832 fixed the Linear `searchIssues` argument in `tools/productivity/linear/readonly.py`. There is a **second, independent copy** of the same client at `workflows/linear/readonly.py`, and it was never touched. It still builds:

```graphql
query SearchIssues($query: String!, ...) { searchIssues(query: $query, ...) }
```

`searchIssues` names its search string `term`, and requires it. `query` was the argument of `issueSearch` — the endpoint Linear deprecated in favour of this one. Passing it to `searchIssues` fails GraphQL validation, which Linear returns as an **HTTP 400**.

### Why it matters even though nothing calls it yet

This copy ships: `services/api-rs/Dockerfile` copies `workflows/` into the image and `WORKFLOW_DIRS` points at it, and `LinearReadonlyClient` is exported from `workflows/linear/__init__.py`. Nothing on the workflows path calls `search_issues` today — so this is a latent 400 on a public class rather than a live break. That is precisely why it went unnoticed.

The two copies now agree character-for-character in this method.

### The test is the actual fix

This also mirrors the regression test #832 added on the tools side (`assert "searchIssues(term: $term" in query` / `assert "query:" not in query`). **The absence of that test here is why the fix didn't propagate** — the `tools/` copy is pinned, this one wasn't, so they were free to drift.

Verified the new test fails against the unfixed client and passes against the fixed one.

### DIFF
--- a/workflows/linear/readonly.py
+++ b/workflows/linear/readonly.py
@@ -374,8 +374,8 @@ class LinearReadonlyClient(LinearGraphQLClient):
     def search_issues(self, query_str: str, limit: int = 25) -> list[dict[str, Any]]:
         """Search issues by text."""
         query = """
-        query SearchIssues($query: String!, $first: Int!, $after: String) {
-            searchIssues(query: $query, first: $first, after: $after) {
+        query SearchIssues($term: String!, $first: Int!, $after: String) {
+            searchIssues(term: $term, first: $first, after: $after) {
                 nodes {
                     id
                     identifier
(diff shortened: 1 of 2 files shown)


aadityakbh:
Withdrawing — already handled on our side; the patch stands if it's useful.

PR #1067, opened 2026-07-14
feat(console): serve the egress allowlist iron-proxy already knows how to enforce

aadityakbh:
`proxy_sync_controller.rb` has said this for a while:

> The top-level `rules` ... intentionally omitted: centaur-console has no models for them yet. Each secret still carries its own per-secret `rules`.

The consequence is larger than the comment suggests. iron-proxy builds its `allowlist` transform **only** when the control plane sends `rules`:

```go
// internal/config/sync.go — TransformsFromSync
if isNonNullJSON(rules) {
    transforms = append(transforms, Transform{Name: "allowlist", ...})
}
```

No rules served → no allowlist transform → nothing rejects. **A managed-mode sandbox can reach any host on the internet.** The per-secret rules don't close this: they scope where a credential may be *injected*, never where the workload may *connect*. And the `Ready` gate doesn't either — it fails closed only until the first config is applied, so an allowlist-free config just makes the proxy ready to pass everything.

This PR gives console the model it says it lacks, and populates it from two sources, because neither is sufficient alone:

- **derived** — every host a granted credential is already scoped to. This is what makes the allowlist self-maintaining: grant a tool its secret and its host is allowed. There's no second list to forget, and no new tool that 403s in production with CI green.
- **base** — the hosts that carry no credential and so appear in no rule: the model provider, package registries, the git host. Derivation is blind to them, and a sandbox that can't reach them can't run.

### Design notes

**Derived from *granted* credentials, not the served subset.** `served_credentials` drops any secret whose source is currently undeliverable. Deriving from it would let a vault blip silently rewrite the egress boundary and 403 a host that is still perfectly well authorized. Losing a credential should cost the request its credential, never its destination.

**Off by default.** An existing deployment upgrades with its egress behavior *and its `config_hash`* byte-for-byte unchanged. No proxy re-applies.

**Omitted is not empty.** `rules: []` is not "no allowlist" — it's an allowlist that matches nothing and 403s *every* request, including the one to the model provider. So the key is absent when off, never null and never empty. Turning the allowlist on with an empty base list is refused by a validation, because that's the same blackout arriving by a different road.

**`warn` mode** serves the rules but tells the proxy to log what it *would* have blocked rather than blocking it — the audit pass before enforcing. It rides on an `allowlist_warn` field that iron-proxy does not read yet; a companion PR to `ironsh/iron-proxy` would carry it through `TransformsFromSync`. Until that lands, `warn` serves rules the proxy will enforce, so it should be left off.

### Tests

11 new tests covering off/warn/enforce, the union, dedup, method+path scoping surviving the `http_methods`→`methods` translation, the undeliverable-source case, and the empty-base-list guard. Full console suite: **1211 runs, 0 failures, 0 errors.**

### DIFF
--- a/services/console/app/controllers/api/v1/proxy_sync_controller.rb
+++ b/services/console/app/controllers/api/v1/proxy_sync_controller.rb
@@ -14,9 +14,17 @@ module Api
     # entry per granted PgDsnSecret, keyed by foreign_id; the proxy's
     # locally-defined listeners bind to these by foreign_id.
     #
-    # The top-level `rules`, `mcp`, and `ingest_token` fields the proxy also
-    # understands are intentionally omitted: centaur-console has no models for them
-    # yet. Each secret still carries its own per-secret `rules`.
+    # `rules` is the egress allowlist -- the destinations the sandbox may reach
+    # at all, as opposed to the per-secret `rules` that scope where a credential
+    # may be injected. iron-proxy compiles it into its `allowlist` transform,
+    # which runs first in the pipeline, so an unmatched host is rejected before
+    # any secret is swapped in. Sent only when the operator has turned the
+    # allowlist on; omitted otherwise, and omitted is not the same as empty --
+    # an empty `rules` array is an allowlist that matches nothing and 403s
+    # everything. See Principal#egress_rules.
+    #
+    # The top-level `mcp` and `ingest_token` fields the proxy also understands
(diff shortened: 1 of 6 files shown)


mslipper:
Permissive egress is a deliberate product decision in Centaur. If you close down the egress, you make agents much less useful. Closing for now; can revisit if we change our minds here later.

PR #1071, opened 2026-07-14
fix(slackbot): upgrade deduped mention events

decofe:
Slack may deliver the same visible message as both `message` and `app_mention`. When the empty `message` callback arrives first, Chat SDK currently claims the dedupe key and drops the later actionable mention.

Patch Chat SDK dedupe state to rank mentions above non-mentions, allowing one atomic upgrade while retaining duplicate suppression. The upgrade path is serialized; normal first deliveries retain the existing set-if-absent fast path.

Adds regression coverage for both observed callback orders and asserts exactly one append/execution with attachment content preserved.

Validation:
- `pnpm --filter slackbotv2 test -- chat-sdk-emulate.test.ts` (175 pass, 1 skipped)

Prompted by: @kuyziss


mslipper:
Closing for the time being. We need to upgrade the Chat SDK; that work will most likely invalidate the patch in this PR.

PR #1075, opened 2026-07-14
feat(slackbotv2): default model and reasoning effort per Slack channel

cjustice:
## Summary

Pin a default **harness / model / provider / reasoning** per Slack channel, so users there don't type `--claude` / `--model` / `-rsn` every message. Today defaults are deployment-wide and display-only; there's no way to make an engineering channel default to Claude Code at high effort while triage stays on codex.

## How it works

`SLACKBOTV2_CHANNEL_DEFAULTS` (chart: `slackbotv2.channelDefaults`) — JSON keyed by Slack conversation id, each value an object of optional `harness`/`model`/`provider`/`reasoning`:

```json
{
  "C0ENG":    {"harness": "claude", "model": "opus", "reasoning": "high"},
  "C0TRIAGE": {"reasoning": "low"}
}
```

- **Shared vocabulary.** Each object is normalized by `normalizeHarnessOverrides`, which resolves through the *same* tables the inline flag parser uses (harness aliases, provider shortcuts, Claude model aliases, reasoning efforts) — so a channel field and a `--flag` validate identically.
- **Coherent harness+model.** A model is only valid within a harness, so `harness` is first-class; a Claude model implies claude-code (like `--opus`). Other models are validated at startup against the harness the channel will use (its own, else the deployment default `SLACKBOTV2_DEFAULT_HARNESS`); a real mismatch (e.g. a codex model on a claude-code channel) is warned, not dropped. A `provider` (e.g. bedrock) legitimately runs Claude on codex, so the check is skipped then.
- **Precedence** (all four knobs): explicit/sticky per-thread flag → channel default → deployment/baked default. Unlike the deployment default, the channel default rides the harness input line so it takes effect this turn (harness/model/provider sticky, reasoning per-turn) and is recorded in execution metadata. Setting the harness restarts a thread onto it like `--claude`/`--codex`.
- Malformed config is logged and skipped; the bot never crashes on it.

Reads the channel id from the existing `slack:CHANNEL[:TS]` thread key. Conceptually mirrors #1073 (`ctx.agent_turn` / `AGENT_DEFAULTS`).

## Testing

- `bun test test`, `bun run check:types` — pass (adds parser/validation, `normalizeHarnessOverrides`/`harnessForModel`, and e2e cases that a channel default's harness reaches session-create and its model/reasoning reach the input line, and that explicit flags win).
- `helm template … --set-json slackbotv2.channelDefaults=…` renders the env var; unset renders nothing. Docs: `reference/configuration`.


goksu (changes requested):
This part of the flow is a bit confusing so it is easy to miss but default model is not sufficient enough as it depends on the harness & model pair. This would effectively result in possible configs where Codex is the default harness as usual and a model of Claude.

What you want here instead channel overrides to be both be model and harness values. Once both are set in combination: rest makes sense.

cjustice:
@goksu addressed your review concerns.

This uncovered a larger chunk of work to rationalize the model / reasoning / harness configuration; the configuration precedence hierarchy should be made explicit, and model + harness should be validated as a pair. But hopefully you feel comfortable approving this PR as-is.

goksu (approved):

PR #1115, opened 2026-07-17
feat(console): make login session lifetime configurable

cjustice:
## Summary

The console login cookie had no configured expiry, so Rails emitted a browser-session cookie. Many browsers drop that on close, signing operators out far sooner than expected.

This adds a `config/initializers/session_store.rb` that sets an absolute cookie lifetime, configurable via `CENTAUR_CONSOLE_SESSION_MAX_AGE` (seconds, with the usual `IRON_CONTROL_` legacy fallback):

- unset → 14 days (persistent, the new default)
- positive integer → that many seconds
- `0` or `session` → browser-session cookie (previous behavior)
- invalid → falls back to the default (never silently shortens the session)

The cookie key is pinned to the historical `_iron_control_session` (the value Rails derived from the `IronControl` app module), so changing the lifetime does not invalidate operators' existing sessions.

## Why

Operators were being signed out of the console too quickly. This makes the session persistent by default and lets deployments tune the lifetime.

Parsing lives in a dependency-free `ConsoleSession` module (mirrors `ConsoleEnv`), unit-tested; an integration test asserts the login response sets a persistent cookie with the pinned key. New env var documented in `services/console/README.md`.


cjustice:
Parked for now. The lifetime change itself is sound and tested, but extending the login cookie widens the offboarding window: the console session is a self-contained cookie-store cookie with no server-side revocation, and nothing auto-disables a user when they lose IdP/Slack access. Revisit alongside a decision on IdP revalidation — viable for Google via `prompt=none`+`id_token_hint`, but **not** for Slack (its OIDC discovery advertises no `prompt_values_supported`, `request_parameter_supported: false`, and no `offline_access` scope) — or simply documenting that offboarding must disable the console account. Branch retained for reopen.

PR #1122, opened 2026-07-18
fix(chart): avoid duplicate agent-sandbox archive entries

alphastorm:
## Summary

- package the release chart from a staging copy containing only the packaged `agent-sandbox` dependency;
- fail the release job when a chart archive contains any duplicate tar member path; and
- publish the corrected chart under the next chart version.

## Problem

The `centaur-0.1.105` release asset `centaur-0.1.105.tgz` (SHA-256 `fd810b08a2d13811bfed25e1a3832f57da5dd6312747785773aedb566c9f69da`) contains 17 paths under `centaur/charts/agent-sandbox/` twice. Every pair inspected is identical in bytes, mode, and size.

Duplicate archive member names make the chart non-canonical, can cause strict archive or supply-chain verification to reject it, and create extractor-order semantics in the general case. No divergent duplicate content, installation compromise, or malicious ambiguity was observed in this artifact.

<details>
<summary>Duplicated paths</summary>

1. `charts/agent-sandbox/Chart.yaml`
2. `charts/agent-sandbox/values.yaml`
3. `charts/agent-sandbox/templates/_controller-args.tpl`
4. `charts/agent-sandbox/templates/_helpers.tpl`
5. `charts/agent-sandbox/templates/clusterrolebinding-extensions.yaml`
6. `charts/agent-sandbox/templates/clusterrolebinding.yaml`
7. `charts/agent-sandbox/templates/deployment.yaml`
8. `charts/agent-sandbox/templates/extensions-rbac.generated.yaml`
9. `charts/agent-sandbox/templates/namespace.yaml`
10. `charts/agent-sandbox/templates/rbac.generated.yaml`
11. `charts/agent-sandbox/templates/service.yaml`
12. `charts/agent-sandbox/templates/serviceaccount.yaml`
13. `charts/agent-sandbox/README.md`
14. `charts/agent-sandbox/crds/agents.x-k8s.io_sandboxes.yaml`
15. `charts/agent-sandbox/crds/extensions.agents.x-k8s.io_sandboxclaims.yaml`
16. `charts/agent-sandbox/crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml`
17. `charts/agent-sandbox/crds/extensions.agents.x-k8s.io_sandboxwarmpools.yaml`

</details>

The source-correlated cause is the release sequence:

1. `helm dependency build contrib/chart` packages the local `file://charts/agent-sandbox` dependency as a `.tgz` while the checked-in `charts/agent-sandbox/` directory remains present.
2. The chart-releaser packaging path then loads both representations.
3. Helm serializes dependencies beneath `centaur/charts/<chart-name>/`; both representations are named `agent-sandbox`, so their files receive the same tar member paths.

## Fix

The workflow now:

1. copies the dependency-built chart into a temporary staging directory;
2. verifies that exactly one packaged `agent-sandbox-*.tgz` exists;
3. removes only the unpacked `charts/agent-sandbox/` copy from staging;
4. packages the staged chart with the installed Helm version;
5. rejects any duplicate member path with a standard-library Python check; and
6. tells chart-releaser to upload the already verified package with `skip_packaging: true`.

The checked-in dependency source remains available for development and linting; only the temporary release staging copy is normalized.

## Verification

- Exact-release verifier: published asset SHA-256 matches `fd810b08a2d13811bfed25e1a3832f57da5dd6312747785773aedb566c9f69da` and reproduces exactly 17 identical duplicate pairs.
- `python3 .github/scripts/check-chart-archive.py /tmp/centaur-0.1.105.tgz`: exits non-zero and lists exactly 17 duplicate paths.
- Helm v4.1.1: `helm dependency build contrib/chart`
- Helm v4.1.1: `helm lint contrib/chart` and lint of the corrected packaged archive
- staged `helm package` command from the workflow
- `python3 .github/scripts/check-chart-archive.py .cr-release-packages/centaur-*.tgz`: passes on the corrected package.
- `git diff --check`

No Kubernetes deployment is needed for this release-packaging-only change.


mslipper:
I'd rather fix the root cause of the duplication than add a bunch of machinery to check it after the fact. If this is still causing you issues, please reopen with a root cause fix.

PR #1127, opened 2026-07-19
fix: reload overlay skills between turns

decofe:
Automatically refresh long-running Codex sessions when a repo-cache-backed overlay changes.

- publish the applied repo-cache generation after a successful skills refresh
- restart only the internal Codex app-server/thread before the next turn
- ignore the stale continuation ID for that one restart, while Slack re-supplies the same thread context
- keep the user in the existing Slack thread

Validation:
- `uv run python -m unittest discover -s services/sandbox -p 'test_repo_cache_watch.py'` (4 passed)\n- `cargo +nightly fmt --manifest-path crates/harness-server/Cargo.toml --all --check`\n- Rust tests could not fetch the pinned `openai/codex` dependency because the environment returned HTTP 401.\n\nPrompted by: vijith

### DIFF
--- a/crates/harness-server/src/codex.rs
+++ b/crates/harness-server/src/codex.rs
@@ -1,5 +1,7 @@
 use std::env;
+use std::fs;
 use std::io::{self, BufRead, Write};
+use std::path::{Path, PathBuf};
 use std::process::{Child, ChildStdin, Command as ProcessCommand, Stdio};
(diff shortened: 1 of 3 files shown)


decofe:
Closing at the requester’s direction. Automatically resetting internal sessions on overlay updates is too disruptive; no runtime change should ship from this PR.

PR #1138, opened 2026-07-21
Make Python workflow tool calls durable

odyslam:
## Summary

- Checkpoints `WorkflowContext.call_tool` and `ctx.tools` calls across the Python shim and raw api-rs protocol boundary, with replay-stable identities derived from canonical tool/method arguments and occurrence indexes.
- Exposes the authoritative positive Absurd task attempt as `WorkflowContext.attempt` through local and sandboxed Python host paths.
- Requires every workflow-principal agent turn to declare explicit `allowed_tools`, rejects omission, and derives an API-disabled child principal holding only the exact parent-role intersection.
- Revalidates child identity, direct grants, parent roles, and exact tool scope on creation and every execution/recovery; parent-role revocation blocks future reuse.
- Persists the tool allowlist in sandbox capabilities so a changed scope replaces the assigned sandbox.
- Physically removes disallowed tool source packages before an enforced sandbox starts; scoped bootstrap failures are terminal rather than warning-and-continuing.
- Rejects empty or malformed successful tool responses, rejects workflow-principal slug collisions before registration, and keeps workflow-owned sandbox cleanup explicit.
- Publishes final api-rs and agent manifests with exact revision labels, strict platform/attestation descriptors, signed SLSA provenance, and per-platform SPDX SBOM attestations.

## Why

Python tool calls could repeat after a task retry or worker restart, and workflows lacked a truthful task-attempt identity. The earlier workflow principal boundary also trusted a broad parent assignment, hid unapproved shims without removing source, and could reuse a sandbox after scope or parent-role drift. This head makes those boundaries explicit and fail-closed.

## Evidence

- Exact head: `3b11281793ba63f7978c73972be79c59b01fda82`
- `cargo fmt --all --check`: passed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- `cargo test --workspace`: 512 passed, 18 Kind infrastructure tests ignored.
- SQLx against `paradedb/paradedb:0.23.0-pg16`: all 15 tests passed, including migration 0047 and empty/scoped allowlist round trips.
- Shared-database migrations are serialized inside the SQLx unit binary; the fresh-database suite passes with eight test threads.
- Sandbox and workflow-host Python suites: 58 passed.
- Actionlint, shell syntax, migration ordering, generated-doc scope, and `git diff --check`: passed.

## Operational limits

Checkpointing gives completed-call replay safety. A process failure after an external mutation succeeds but before its result checkpoint persists can still retry, so mutating tools need their own stable idempotency boundary.

`ctx.attempt` is the workflow task attempt, not an outbound delivery-attempt counter. Callers must number and checkpoint their callback attempts separately.

This draft does not make the downstream PAS release deployable. It must merge, api-rs and sandbox images from the merged descendant must be promoted, and the downstream exact provenance/capability gates plus the replacement Argo transaction must be implemented before deployment.


odyslam:
Final operational-readiness follow-up on `8776c0aa75cfdd0b4e6c3f5ca309a4d1e3c2770c`:

- auto-registered workflow principals explicitly set `sandbox_api_server_enabled=false` and reconciliation fails if Console returns effective access enabled;
- principal-backed workflow sandboxes independently disable the api-server capability and remove direct api-rs URLs/credentials;
- durable `ctx.call_tool` remains available over the workflow host protocol;
- docs require route-scoped api-rs authorization before any principal may enable direct API access.

Verification: workspace fmt/clippy passed; 229 affected-crate tests and 490 workspace tests passed (18 Kind e2e ignored); diff check passed. The latest local run had no `ABSURD_TEST_DATABASE_URL`, so its database-only replay branch was not rerun; non-database restricted-sandbox/checkpoint/tool-dispatch coverage did run.

PR #1147, opened 2026-07-22
fix(slack): preserve upload context for DMs

ribeirojose:
## What changed

- recognize legacy Slack DM thread keys shaped as `<channel>:<thread_ts>`
- use the root message timestamp for root-DM keys shaped as `slack:<D-channel>:`
- include the API-owned Slack upload destination in both cases
- add regressions for both session-key formats

## Why

DM sessions can reach the session API in either format. Previously neither
format produced the Slack session context that tells the sandbox where uploads
belong, so file delivery from those sessions could lose its channel/thread
destination.

## Validation

- Slackbot v2 tests: 211 passed, 1 skipped
- Slackbot v2 typecheck
- `git diff --check`

### DIFF
--- a/services/slackbotv2/src/session-api.ts
+++ b/services/slackbotv2/src/session-api.ts
@@ -1647,7 +1647,7 @@ function codexInputContent(
   contextPreamble?: string
 ): JsonValue[] {
   const content: JsonValue[] = []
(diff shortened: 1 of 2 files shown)


mslipper:
I believe this behavior is fixed with the new Slack agent experience and the upgraded chat SDK. Please reopen if that's not the case.

PR #1148, opened 2026-07-22
fix(ci): skip docs deploy without Cloudflare credentials

ribeirojose:
## What changed

- require both Cloudflare credentials before running the non-PR docs deploy
- preserve the existing build and explicit skip message when credentials are
  unavailable

## Why

The workflow already detects missing credentials, but its deploy step only
checks the event type. On a main-branch push in a fork or repository without
Cloudflare secrets, the build succeeds, the skip message runs, and Wrangler is
still invoked with empty credentials.

## Validation

- RED/GREEN static assertion for the deploy condition
- `actionlint .github/workflows/docs.yml`
- YAML parse validation
- `git diff --check`


mslipper:
Superseded by #1428.

PR #1151, opened 2026-07-22
fix(deps): upgrade Hono past security advisories

ribeirojose:
## Summary

- raise all four Hono-based bot services to `^4.12.27`
- resolve the workspace to Hono 4.12.31
- clear the three moderate Hono advisories reported on the previous lockfile

This is intentionally limited to the direct dependency ranges and lockfile resolution.

## Validation

- `pnpm audit --json` reports zero Hono advisories
- Slackbot: 209 passed, 1 skipped
- Discordbot: 140 passed
- GitHubbot: 77 passed
- Linearbot: 103 passed
- Discordbot, GitHubbot, and Linearbot typechecks pass
- `git diff --check`

## Current-main baseline

Slackbot's typecheck currently fails in unchanged code from #1141 at `src/session-api.ts:872` because an indexed message may be `undefined`. This branch does not modify that file; the Slackbot test suite passes.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,8 +90,8 @@ importers:
         specifier: ^14.25.1
         version: 14.26.4
       hono:
-        specifier: ^4.12.18
-        version: 4.12.25
+        specifier: ^4.12.27
+        version: 4.12.31
       pg:
         specifier: ^8.21.0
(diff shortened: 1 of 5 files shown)


mslipper:
Will handle in broader dep upgrade.

PR #1153, opened 2026-07-22
fix(credentials): expose safe inject placeholders

ribeirojose:
## Summary

- deserialize the non-secret source metadata for effective inject secrets
- expose an env-backed inject secret as `NAME=NAME` in the sandbox, matching existing placeholder discovery
- keep the actual credential exclusively in iron-proxy
- ignore non-env, missing, empty, or `=`-containing source names

This lets tool clients detect that an inject-mode credential is configured without materializing its value in the sandbox.

## Validation

- `cargo fmt --all --check`
- `cargo test -p centaur-iron-control -p centaur-sandbox-agent-k8s` (63 + 47 passed)
- `cargo clippy -p centaur-iron-control -p centaur-sandbox-agent-k8s --all-targets -- -D warnings`
- `git diff --check`

The regression tests cover replace placeholders, env-backed inject placeholders, and rejection of unsafe or unsupported inject sources.

### DIFF
--- a/services/api-rs/crates/centaur-iron-control/src/lib.rs
+++ b/services/api-rs/crates/centaur-iron-control/src/lib.rs
@@ -18,11 +18,11 @@ pub use client::IronControlClient;
 pub use error::{IronControlError, Result};
 pub use models::{
     AwsAuthSecretInput, BrokerCredentialInput, BrokerCredentialRecord, EffectiveConfig,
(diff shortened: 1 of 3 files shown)


mslipper:
Closing this because SDK-based tools already receive secret-name placeholders through StubBackend, while this implementation does not work with the default OnePassword secret source and can use the wrong name when the logical secret and source reference differ.

PR #1154, opened 2026-07-22
fix(runtime): cold-start principal-scoped sandboxes

ribeirojose:
## Summary

- skip warm-pool claims for sessions carrying an iron-control principal
- cold-start those sessions with the principal-resolved sandbox spec
- leave the unclaimed warm sandbox ready for an eligible session
- record `principal_scoped` as the warm-pool rejection reason

A running warm container cannot gain principal-specific environment placeholders when its proxy is rebound. Cold-starting keeps the sandbox environment and proxy principal consistent.

## Validation

- `cargo fmt --all --check`
- `cargo test -p centaur-session-runtime` (102 passed)
- `cargo clippy -p centaur-session-runtime --all-targets -- -D warnings`
- `git diff --check`

The always-running unit regression covers the eligibility policy. The DB-backed adoption regression also compiles and follows the crate's existing `SESSION_RUNTIME_TEST_DATABASE_URL` gate.

### DIFF
--- a/services/api-rs/crates/centaur-session-runtime/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-runtime/src/lib.rs
@@ -2566,6 +2566,12 @@ impl SessionRuntime {
                 .as_ref()
                 .is_none_or(|warm| warm == harness_type);
             let warm_persona_matches = persona_context.is_none();
+            // A warm agent container has already started, so its environment
+            // cannot gain the placeholders resolved for a session principal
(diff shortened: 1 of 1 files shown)


mslipper:
As written, this will functionally disable the warm pool since every sandbox eventually receives a principal. Tools should generally not need env vars. Instead, tools should know that they're in an iron-proxy environment and allow credentials to be injected at the network layer.

PR #1155, opened 2026-07-22
feat(slack): join public channels on mention

ribeirojose:
## Summary

- call `conversations.join` for public-channel `app_mention` events before the normal handoff
- cache successful or already-joined results for six hours per process
- keep join failures non-fatal and visible in structured warning logs
- document the required `channels:join` bot scope

The behavior is limited to public channel IDs (`C…`). Private channels and rich-text mentions delivered as ordinary message events are unchanged.

## Validation

- regression proved the join call was absent before implementation
- focused signed-webhook/emulated-Slack regression passes and proves repeated mentions issue one join
- `pnpm --filter slackbotv2 test` (210 passed, 1 skipped)
- `git diff --check`

## Current-main baseline

`pnpm --filter slackbotv2 run check:types` reaches only the existing current-main error at `src/session-api.ts:872`, introduced by #1141. This PR does not modify that file and adds no typecheck errors.

### DIFF
--- a/docs/pages/quickstart.mdx
+++ b/docs/pages/quickstart.mdx
@@ -167,7 +167,9 @@ value and sanitized Slack user, team, channel, thread, and message metadata.
 
 Subscribe to the `app_mention` bot event. For a minimal channel-mention test,
 the app also needs Bot Token Scopes that let it read mentions and write replies,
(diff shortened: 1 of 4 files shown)


mslipper:
This is superseded. `main` already supports automatically joining new public channels.

PR #1156, opened 2026-07-22
feat(slack): include bounded root file context

ribeirojose:
## Summary

- query bounded channel history when a user starts a root channel mention
- include only that user's preceding messages that contain files
- limit history to ten results, a ten-minute window, and four prior files plus the current mention
- preserve chronological ordering and exclude stale, other-user, text-only, future, and bot-authored messages
- document the required public/private channel history scopes

Thread replies keep the existing full thread-context path. If history collection fails, the existing degraded-context fallback sends only the current mention.

## Validation

- regression proved the preceding file was absent before implementation
- focused signed-webhook/emulated-Slack regressions cover inclusion, exclusions, ordering, and caps
- `pnpm --filter slackbotv2 test` (212 passed, 1 skipped)
- `git diff --check`

## Current-main baseline

`pnpm --filter slackbotv2 run check:types` reaches only the existing current-main error at `src/session-api.ts:872`, introduced by #1141. This PR does not modify that file and adds no typecheck errors.

### DIFF
--- a/docs/pages/deploying-in-production.mdx
+++ b/docs/pages/deploying-in-production.mdx
@@ -205,7 +205,9 @@ Create the Slackbot app at [api.slack.com/apps](https://api.slack.com/apps).
 Use the app page to install the bot, copy the Bot User OAuth Token for
 `SLACK_BOT_TOKEN`, and copy the Signing Secret for `SLACK_SIGNING_SECRET`.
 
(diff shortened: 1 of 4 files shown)


mslipper:
I tried out this issue internally, and Centaur correctly detects root channel files. Closing for now, please reopen if the issue persists.

PR #1159, opened 2026-07-22
security: harden sandbox and ingress boundaries

sehyunc:
## Summary

- enforce the restricted Kubernetes security context on agent containers
- replace wildcard iron-proxy egress with an explicit production domain allowlist
- reject Slack webhook bodies larger than 1 MiB before signature processing
- support least-privilege proxy Secrets and digest-pinned agent-sandbox controller images

## Evidence

- live usw-dev audit detected all three boundaries
- 2 MiB signed Slack payload currently returns 200 before this patch
- fresh agent pod currently lacks a container securityContext
- unlisted `https://example.com` currently succeeds through iron-proxy
- Helm manifests were rendered and the chart changes are deployed on usw-dev

## Validation

- `git diff --check`
- chart rendered with the production values
- runtime TypeScript/Rust test execution is delegated to CI because the deployment VPS is explicitly prohibited from running Cargo and does not have the Bun workspace dependencies installed


sehyunc:
Closing at the author's request; this deployment audit is not intended to submit upstream code changes.

PR #1175, opened 2026-07-24
feat: datasearch tool (Tako) with keyless free-tier fallback

25eliu:
## Summary

New `tools/research/tako/` plugin: **datasearch** — cited, chart-backed structured data across companies & financial markets, macroeconomics & government data, digital & industry intelligence, sports, polling & live events, and weather & climate, plus web results, via [Tako](https://tako.com). Coverage is stated by domain rather than by the licensed sources behind it: sources change without the coverage changing, each card carries its own attribution, and naming three made the tool read as narrower than it is. The list has one definition, `DOMAINS` in `_coverage.py`, which the CLI help and the tool-directory row render from. Installed as the `datasearch` CLI; the package/directory keep the `tako` name.

Commands: `available-data` (free coverage discovery), `search` / `answer` (priced structured+cited results), `contents` (dataset export), `slack-card` (render a card as a Slack message), `health`.

## Two backends, one shape

Selected by whether `TAKO_API_KEY` is configured (never by key value — sandboxes hold only placeholders):

- **With a key** → the official `tako-sdk` against `tako.com`, full features.
- **Without a key** → Tako's free rate-limited hosted MCP (`mcp.tako.com`), same additive-credential pattern as websearch's Parallel fallback. `contents` stays key-only.

Both backends return the same shape: `search` → `answer_markdown` (readable `## Tako Data` document) + structured `cards` (`image_url`, `webpage_url`, `node_ids`, `exportable`, row `content`) + `meta`; `answer` → `answer` prose + `cards`; `available-data` → `found`/`summary`/`matches`. The SDK path renders `answer_markdown` from its cards; the MCP path reads the hosted tool's two-channel response (readable text + `structuredContent`). Robust across [tako-mcp#187](https://github.com/TakoData/tako-mcp/pull/187) (bulk payload moving into `structuredContent`): works before and after that deploys. Every response carries `meta.backend` (`tako:sdk`/`tako:mcp`) and `meta.partial_failures`.

## Sandbox integration

- Optional replace-mode `TAKO_API_KEY` bound to `X-API-Key` on `tako.com`; read via the `secret()` placeholder mechanism. `mcp.tako.com` added to the host allowlist for the keyless path.
- Firewall-aware routing: inside a sandbox, tool secrets have no env signal, so with `HTTPS_PROXY` set the client takes the SDK path optimistically and falls back to the free MCP once on a 401 (remembered for the client's lifetime).
- SDK workarounds (documented in the README): explicit iron-proxy/CA wiring (tako-sdk's urllib3 ignores proxy env vars); per-request timeouts via the generated API (the facade drops them); `X-API-Key` default header (the SDK's OpenAPI spec omits auth on the beta graph endpoints).

## Rendering a card into Slack

`slack-card` turns a card into a Block Kit message and prints it; `search` / `answer`
gain `--slack-card`, which attaches the same rendered message under
`slack_card.message`. When a result carries a card the payload also gains
`slack_card.hint` naming the pipeline that posts it, emitted as the **first** key
because a search result runs past a hundred lines of JSON and callers routinely
pipe it through `head -N`.

It renders an `image` block pointed at the card's own `image_url`. Slack scales an
image block to the message width and preserves aspect ratio without cropping, so
there is no geometry to fit, and the headline (the card's own `description`) and
the "Open in Tako" button sit outside the image where they cost no image height.
Nothing here reads, scales, or reformats a value: Tako already chose the axes,
units, and currency, so its render and its description pass through untouched, and
tests pin that. Verified against a live workspace: both the `container` and flat
`image`/`context`/`actions` layouts post, and Slack keeps the blocks as sent.

**Posting deliberately lives in the `slack` tool, not here.** This PR therefore
touches one file outside `tools/research/tako/`: `slack send` gains
`--blocks-json` (a file path, or `-` for stdin; accepts a bare blocks array or a
whole `chat.postMessage` body), about ten lines.

    datasearch slack-card <card> | slack send <channel> "<fallback>" --thread <ts> --blocks-json -

Posting from this tool was implemented first and did not work. Traced from a real
sandbox: `secret("SLACK_BOT_TOKEN")` returns the placeholder, iron-proxy allows
egress to `slack.com` and forwards `POST /api/chat.postMessage`, but no credential
is substituted, so Slack answers `invalid_auth`. The `slack` tool's own `health`
fails identically in the same sandbox, so this is per-tool secret scoping
(`tools/README.md`: "tool secrets are scoped") working as intended rather than a
misconfiguration. Moving the send to the tool that already owns the Slack
credential removed ~120 lines here, plus the `SLACK_BOT_TOKEN` binding and the
`slack.com` host entry from this tool's manifest, so its permissions match its
purpose again.

Two known limitations, neither introduced here:

- Rendering is Slack-specific. The sandbox is not told which chat surface a turn
  belongs to, so the hint is phrased conditionally ("if this turn is in a Slack
  thread") rather than suppressed, which would mean never showing it. Discord,
  Teams, Linear and GitHub thread keys all resolve to no Slack target and degrade
  without error; the retrieved data is never affected.
- Card images render dark only. `output_settings.image_dark_mode` is accepted and
  returns a distinct card id but does not change the rendered pixels, so a light
  variant is not reachable today. Filed upstream as TAKO-3967.

## Notes

- The keyless MCP backend rebuilds its response dict, which dropped the top-level `pub_id`/`image_url`/`embed_url` pointer the SDK path passes through. It is copied over now, so a response whose `cards` come back empty still names its chart and both tiers render identically (pinned by a test asserting byte-identical blocks).
- Tests live in `tests/` (no `__init__.py`) so pytest's module naming doesn't shadow the `tako` SDK package.
- Reviewed against `AGENTS.md` + an adversarial pass; validated live against both `tako.com` (keyed) and `mcp.tako.com` (keyless free tier).
- `docs/pages/reference/tool-directory.mdx` gains a `datasearch` row, matching every other tool under `tools/research/`. The generated `docs/public/md/` copy is hand-edited to match rather than regenerated, so unrelated stale generated pages stay untouched.

## Test plan

- [x] `uv run --no-project --with pytest --with "tako-sdk>=2.2.6" --with httpx --with python-dotenv --with typer --with rich python -m pytest tools/research/tako/tests/test_client.py -v` — 154 tests: coverage helpers, summary builder, available-data pipeline, `_sources` contract, priced-path shaping, input validation, both-channel MCP parsing, sandbox auth-fallback. No network.
- [x] `python scripts/validate_cli_packaging.py` — passes (`datasearch → centaur_tool_tako.cli:app`).
- [x] `ruff check tools/research/tako` — clean, zero findings (CLI imports reordered to the `websearch` pattern; `load_dotenv()` now runs after the imports, and the client is imported lazily inside each command so behavior is unchanged).
- [x] `uv build --wheel tools/research/tako` — wheel contains only the nine source files; no `.env`, `.venv`, or caches leak in.
- [x] Live keyed (`TAKO_API_KEY`): `datasearch health`, `available-data`, `search`, `answer`.
- [x] Live keyless free tier (`mcp.tako.com`): `available-data`, `search`, `answer` return real readable content.
- [x] `datasearch --help` from an installed wheel — Typer app resolves and lists `health`, `search`, `answer`, `contents`, `available-data`.
- [x] Catalog discovery + one real command from a local Kubernetes sandbox (`kind-centaur`), per `AGENTS.md`, run against this PR's head (`f292f359`) — the local stack's repo-cache tracks this branch, so the sandbox ran the pushed source, not a local copy:
  - `centaur-tools list` → `datasearch	/app/tools/research/tako`, listed alongside its `tools/research/` siblings. This also confirms the catalog name is `datasearch`, which is the name used in the Tool Directory row.
  - `datasearch --help` → Typer app resolves through the shim; all six commands listed, including `slack-card`.
  - `datasearch health` and `datasearch available-data "Nvidia"` → real cited data returned through iron-proxy. With no `TAKO_API_KEY` in this deployment's vault, the client hit a 401 on the SDK path and emitted one line to stderr — `tako: API rejected the placeholder credential (no TAKO_API_KEY in the deployment vault); falling back to the free MCP tier` — then succeeded on `"backend": "tako:mcp"`. That is the firewall-aware routing and one-shot 401 fallback described above, exercised end to end in-cluster.
  - `datasearch search "NVIDIA Corporation Revenue"` → documented shape confirmed on the wire: `answer_markdown` (`## Tako Data (1 card)`) + `cards` carrying `image_url`, `webpage_url`, `exportable`, `content` + `meta.backend` / `meta.partial_failures`.
- [x] `ruff check tools/research/tako tools/productivity/slack/cli.py` — clean for the tako tool; the slack CLI's 17 pre-existing findings are unchanged before and after this PR's edit (verified by checking out the HEAD version at the same path and re-running).
- [x] Render-to-post handoff: `datasearch slack-card` output fed straight into `slack send`'s `--blocks-json` reader, which accepts it and returns the container block. The reader is unit-tested against a whole `chat.postMessage` body, a bare blocks array, empty stdin, non-JSON, a missing `blocks` key, and an empty `blocks` array.
- [x] Live Slack workspace: both layouts posted via `chat.postMessage`; `response_metadata` shows the blocks were kept, not silently dropped. Slack's `ignored_extra_attributes_for_image_block` warning is benign and unavoidable — the minimal legal image block produces it, and omitting `alt_text` to silence it fails outright with `missing required field: alt_text`.
- [x] Non-Slack surfaces: the real thread-key shapes for Discord, Linear, GitHub and Teams (taken from each ingress's source) all resolve to no Slack target across both code paths — no exceptions, and `cards` intact in every case.
- [ ] Posting from a sandbox is currently blocked in the local stack: credential substitution is not applied to any tool there, including the `slack` tool's own `health`. Not introduced by this PR and not fixable within it.


25eliu:
Addressed all six findings from @robertabbott's review in 698c16c3 (replying here because the review threads aren't visible to the API yet — will follow up inline once they are):

1. **`found` only reflects the top-2 drill / `other_matches` lack node_id** — did both suggested fixes: `other_matches` now carry `node_id`, and the summary marks them "Also matched (not checked)". When no drilled match has coverage, the summary states outright that the unchecked hits are not proof Tako lacks data and points at `--types`/`--label` narrowing or `search --node-id`. Verified live: rank-3+ hits now come back pinnable.

2. **Frozen `NER_LABELS`/`NODE_TYPES` copies** — switched to deriving from the SDK enums (`tuple(l.value for l in NerLabel)` etc.); the frozen copies weren't deliberate. Same treatment applied to the new option validators (`SearchEffortLevel`, `ContentsDeliveryMode`, `ContentsFormat`), with tests asserting the derived tuples track the enums.

3. **`--effort medium` / `--data-count 50` uncaught tracebacks** — all enum and range options are now validated in the client before any network call (`ValueError`), and the CLI translates those into one-line `BadParameter` messages. Confirmed no tracebacks for effort/mode/content_format/count violations.

4. **No request timeout** — went through `_api` as suggested; it's the same escape hatch the auth-header workaround already uses. Every operation now passes `_request_timeout`: 120s for search/answer/contents, 30s for graph calls, both constructor-overridable. README maintainer note updated.

5. **`Sources(data=None)` serializing `"data": null`** — fixed with conditional kwargs exactly as proposed; a skipped source is now genuinely absent from the wire body. Test asserts `"data" not in to_dict()` and not in `model_fields_set`.

6. **`data_count=0` silently dropping `node_ids`/`strict`; negatives folded into skip** — both now raise: `data_count=0` with pinning is rejected as contradictory, counts are range-checked 0–20 (negatives rejected), and skipping both sources is rejected too.

69 unit tests pass (up from 54); live free smoke (`health`, `available-data "tesla"`) green.

@claude-address

25eliu:
## Follow-up: the card renders, but nothing tells the agent to post it

Flagging a deliberate gap so it is a decision on the record rather than a surprise.

**What works.** `datasearch search`/`answer` return the cards, and when a result carries one the payload gains a `slack_card` hint naming the exact pipeline, on both the keyed and keyless backends:

```
if this turn is in a Slack thread, post this chart with:
datasearch slack-card 2DCtQtFLAXH7nxEe_50X | slack send <session_context.slack.channel_id> "<fallback>" --thread <session_context.slack.thread_ts> --blocks-json -
```

**What does not.** The agent usually ignores it. Traced on a real Slack turn (keyless sandbox, `meta.backend: tako:mcp`) — the complete list of commands it ran was:

```
datasearch --help 2>&1 | head -30
datasearch answer "Microsoft annual revenue 2021 2022 2023 2024 2025" 2>&1
```

It read the hint and stopped. No `slack-card`, no `slack send`. No rate limit, no auth error, cards present in the output. Keyed behaves the same, so this is not a free-tier gap.

**Why.** Nothing in the agent's instructions asks for the chart to be rendered into the thread. `services/sandbox/SYSTEM_PROMPT.md` has a routing line telling the agent to *use* `datasearch` for real-world metrics, but no render obligation, so the only signal is a conditional hint inside a large JSON payload — a weak one competing with "answer the question." That is why the behaviour is inconsistent rather than absent.

**Deliberately not fixed here.** Making it reliable needs roughly one line of agent instruction ("when a datasearch result carries a `slack_card` hint and the turn is in Slack, run the command it names before answering"). Two candidate homes:

1. **`services/sandbox/SYSTEM_PROMPT.md`** — reliable, but 30081e02 in this branch removed a `datasearch` line from that file precisely to keep this PR tool-scoped per AGENTS.md, and its commit message says the routing preference "belongs in a separate prompt PR or a deployment persona overlay." Re-adding it here would reverse that. It also needs an agent-image rebuild to take effect: the prompt is `COPY`'d to `/home/agent/AGENTS.md` at build time, unlike tools, which come from the repo cache.
2. **A persona overlay** — what 30081e02 itself nominates, and it needs no image rebuild.

Recommendation: keep this PR tool-scoped and do (1) or (2) separately. The tool side needs no further change either way — the hint is already emitted with the right card id and the right pipeline.

mslipper:
closing in favor of https://github.com/paradigmxyz/centaur/pull/1600

PR #1182, opened 2026-07-26
feat: add the official Vercel CLI with iron-proxy token injection

droplet-rl:
> **Note:** this PR was prompted by Hart Lambur at Across Protocol / Risk Labs, and authored by Droplet, Across's Centaur-like multiplayer harness. The Across team is trying to build feature parity between Droplet and Centaur before porting over. Please tag @hlambur in GitHub or `@hart` in the Centaur community Slack channel with comments on the quality or style of this PR.

## What

- Pins `vercel@57.0.0` in the sandbox image next to the other global npm CLIs, with the same version-verification pattern.
- Adds `tools/infra/vercel`: the `vercel` / `vc` CLI shims are thin auth-injecting passthroughs to the npm binary — they append `--token VERCEL_TOKEN` (the placeholder) when the caller didn't pass one, then `execv` the real binary so streaming output and interactive prompts are preserved.
- Declares the replace-mode secret (`match_headers = ["Authorization"]`, `hosts = ["api.vercel.com"]`) so iron-proxy swaps the placeholder on the wire; the sandbox never holds the credential.
- `client.py` exposes a small `_client()` surface (`run` / `whoami` / `list_projects` / `list_deployments` / `deployment_logs`) for workflow `ctx.call_tool(...)` use.
- Adds the tool-directory docs row.

## Why this shape

- Vercel maintains a first-class CLI with weekly releases; delegating to it beats hand-maintaining a REST-wrapper subset that goes stale. Agents get the full familiar surface (`vercel ls`, `vercel inspect <url> --logs`, `vercel env pull`).
- Credential handling stays in iron-proxy where it belongs — same philosophy as the `alchemy` tool's placeholder pattern, just applied to a vendor binary instead of an HTTP client.
- The shim intentionally shadows the npm binary on PATH (`~/.local/bin` precedes `/usr/bin`). It resolves the real binary by skipping generated shims via the `install-tool-shims` marker, so both `vercel` and the npm `vc` alias route through token injection and cannot recurse into themselves.

## Notes

- `deployment_logs` uses `inspect --logs` (prints and exits). `vercel logs` streams until interrupted, so it is left to the interactive shim rather than the non-interactive client method.
- Deployments with a locked-down egress allowlist need `api.vercel.com` added.
- A caller-supplied `--token` / `-t` / `--token=` is respected; the wrapper only fills the gap.

## Testing

- `python scripts/validate_cli_packaging.py` passes (both script entries).
- `ruff check` and `ruff format --check` are clean.
- `pytest tools/infra/vercel/test_client.py` — 5 tests covering token injection, explicit-token passthrough, shim-skipping binary resolution, the missing-binary error, and subprocess arg assembly.
- Not yet exercised by the author: a sandbox image build and a live `vercel whoami` through iron-proxy (this PR was prepared without access to a running Centaur deployment). Verification steps below.

## Verify in a deployment

```bash
just build-one sandbox && just deploy
kubectl exec -n centaur <agent-sandbox-pod> -- centaur-tools list | grep vercel
kubectl exec -n centaur <agent-sandbox-pod> -- vercel whoami
# env in the sandbox shows only the VERCEL_TOKEN placeholder; iron-proxy logs
# show the Authorization substitution on api.vercel.com
```


mslipper:
Thanks for this! We have a feature coming that allows official CLIs to become tools natively rather than requiring bespoke wrappers. Closing this in advance of landing that feature.

PR #1187, opened 2026-07-27
fix(sandbox): persist GitHub credentials to disk

cjustice:
## Summary

The sandbox entrypoint leaves GitHub credentials reachable *only* through the environment, for two compounding reasons:

1. `gh auth login --with-token` refuses to persist while `GITHUB_TOKEN` is set in the environment — it prints "To have GitHub CLI store credentials instead, first clear the value from the environment", exits non-zero, and writes no `hosts.yml`. The entrypoint swallows this with `2>/dev/null || true`, so it has always been a silent no-op.
2. `gh auth setup-git` then prepends an *empty* credential helper for `github.com`, which resets the helper list. That orphans the `store` helper and the `~/.git-credentials` file written two lines above, leaving `gh` as the only credential source for git.

Net effect: both `git` and `gh` depend on `GITHUB_TOKEN` being live in the environment of whatever process invokes them.

This change writes `hosts.yml` directly, and sets the `github.com` helper chain explicitly (`store`, then `gh auth git-credential` as a fallback) so both paths work independently. Neither step touches the network, so the block now completes before readiness instead of racing a pod claim from a background job.

## Why

A harness that scrubs sensitive-looking variables out of its tool subprocesses leaves the agent unable to authenticate to GitHub at all. This is not hypothetical: nanocodex's shell tool spawns children with `env_clear()` and re-adds only variables whose names don't match `SENSITIVE_ENV_PARTS` (`nanocodex-tools/src/shell/process.rs`), which drops `GITHUB_TOKEN`. On that harness `git push` gets a 401 challenge it cannot answer, and `gh` reports no credential — so an agent that has finished its work can only report a patch.

It surfaces first as a *signing* failure rather than a push failure, because a server-side signed-commit helper is pure `gh api` (`createCommitOnBranch`) and its first call is `gh api user`. That makes the symptom read as "the token is invalid" rather than "the token is absent from this process".

More generally this is a latent single point of failure for every harness: signed commits, PR creation, and CI polling all hang on one environment variable surviving into every subprocess, and nothing enforces that it does.

## Validated

In an isolated `HOME` with both `GITHUB_TOKEN` and `GH_TOKEN` unset, to reproduce the scrubbed-child condition: `git credential fill` returns a username and password, `gh api user` authenticates, and `git ls-remote` against a private repository succeeds. Before the change, the same conditions give `fatal: could not read Username for 'https://github.com': No such device or address`.

## On credential-at-rest

This writes no credential that wasn't already on disk. `~/.git-credentials` already contains the same value unconditionally; this adds `~/.config/gh/hosts.yml` with that value at `0600`, and adds a `chmod 600` to `.git-credentials`, which currently inherits the process umask.

Worth stating plainly: if `env_clear()`-style scrubbing is intended as a security boundary rather than hygiene, moving the credential to disk routes around it. That boundary is not currently effective — the value is also readable from the embedding process's `/proc/<pid>/environ` at the same uid, and from `.git-credentials` — and a sandbox whose purpose is to open pull requests needs a working credential. But if you'd rather harnesses opt in explicitly (e.g. nanocodex forwarding chosen variables via `Tools::process_environment`), that's a reasonable alternative and I'm happy to take it that way instead.

### DIFF
--- a/services/sandbox/entrypoint.sh
+++ b/services/sandbox/entrypoint.sh
@@ -465,17 +465,31 @@ if [ -n "${CENTAUR_TOOLS_URL:-}" ]; then
     done
 fi
 
+# GitHub credentials have to be reachable from disk, not only from the
+# environment: a harness that scrubs sensitive-looking variables out of tool
+# subprocesses leaves the child with no credential at all otherwise. Nothing
+# here touches the network, so it completes before readiness rather than racing
+# a claim from a background job.
+if [ -n "${GITHUB_TOKEN:-}" ]; then
+    git config --global credential.helper store
+    printf 'https://oauth2:%s@github.com\n' "$GITHUB_TOKEN" > "$HOME_DIR/.git-credentials"
+    chmod 600 "$HOME_DIR/.git-credentials"
+
+    # `gh auth login --with-token` cannot persist this: it refuses while
+    # GITHUB_TOKEN is set in the environment, exits non-zero, and writes nothing.
+    mkdir -p "$HOME_DIR/.config/gh"
+    printf 'github.com:\n    oauth_token: %s\n    git_protocol: https\n' "$GITHUB_TOKEN" \
+        > "$HOME_DIR/.config/gh/hosts.yml"
+    chmod 600 "$HOME_DIR/.config/gh/hosts.yml"
+
+    # Set the helper explicitly rather than via `gh auth setup-git`, which
+    # prepends an empty helper that resets the list, leaving the `store` helper
+    # above unreachable for github.com and gh as the only credential source.
+    git config --global --replace-all credential."https://github.com".helper store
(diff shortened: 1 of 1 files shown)


cjustice:
Superseded by #1192, which fixes this the intended way — forwarding the sandbox credentials to tool subprocesses via `Tools::process_environment` rather than persisting them to disk in the entrypoint.

Two harness-independent defects found while diagnosing are left unfixed and worth noting for whoever picks them up: `gh auth login --with-token` refuses to persist while `GITHUB_TOKEN` is set in the environment, so it exits non-zero and writes no `hosts.yml` (silenced by `2>/dev/null || true`); and `gh auth setup-git` then prepends an empty credential helper that resets the helper list, leaving the `.git-credentials` written two lines earlier unreachable for github.com.

Issue #59, opened 2026-05-19
Layout needs to be centered for smaller sized screens for last two links in landing


decofe:
The landing page layout needs to be centered on smaller screen sizes for the last two links.

Screenshot provided in the Slack thread: `image.png`.

PR #1192, opened 2026-07-27
fix(harness-server): forward sandbox credentials to nanocodex tools

cjustice:
Per Slack with @gakonst. Companion: gakonst/nanocodex#59.

## Problem

The shell tool spawns children with `env_clear()` and re-adds only what survives `sanitized_environment`, dropping any name whose `_`-split parts match `SENSITIVE_ENV_PARTS`. In a Centaur sandbox that removes the credentials tools authenticate with — and most are not secrets: iron-proxy substitutes the real credential at the network boundary, so the sandbox holds a marker whose value is its own variable name.

`GITHUB_TOKEN` matches on `TOKEN`, so `git push` gets a 401 challenge it can never answer and `gh` reports no credential. The agent finishes the job and can then only report a patch. Sandboxes on the codex harness are unaffected, which is what made it read as a bad token rather than a missing variable.

## Fix

Forward them via `Tools::process_environment` — the documented opt-in, and what the CLI already does for the mpp-egress proxy password (`bin/nanocodex/src/config.rs`). Overrides are applied after the filter but still join the redaction list, so forwarded values stay masked in tool output.

Two details:

- **Only sensitive-named variables, not the whole environment.** `NORMALIZED_ENVIRONMENT` (`TERM=dumb`, `PAGER=cat`, `NO_COLOR=1`, …) is applied *before* overrides, so a blanket forward would clobber it and give tools a pager and colour output. The two sets are disjoint; a test pins that.
- **`build_agent` builds tools on two paths** (`tools_factory` when subagents are on, plain `build()` otherwise), so the selection is built once and threaded into both — setting it on one silently misses the other.

## The copied predicate

`SENSITIVE_ENV_PARTS` and `is_sensitive_name` are private in nanocodex-tools, so this copies them, and that copy drifts in the dangerous direction — a part added upstream and missed here silently stops forwarding that variable, which is this bug again. gakonst/nanocodex#59 exposes the set as `ambient_sensitive_environment()`; once it lands this becomes a `rev` bump and one call, deleting the constant, the local predicate, and two of the three tests. Drafted for that reason — please don't merge this version.

## Validated

`crates/harness-server` is not covered by CI (the Rust job in `ci.yml` is scoped to `services/api-rs`), so these were run locally: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test` — 67 unit, 20 integration (4 pre-existing ignored), including two new tests covering that the sandbox credentials match the predicate and that the predicate is disjoint from the normalized shell variables.

Not exercised on a live nanocodex session — our rollout is pinned to 0 while this is open, so end-to-end is a canary once it deploys. The mechanism itself is verified separately: a probe at the pinned rev confirms a `process_environment` override reaches the child intact, string-compared shell-side so output redaction cannot fake a pass.

### DIFF
--- a/crates/harness-server/src/nanocodex.rs
+++ b/crates/harness-server/src/nanocodex.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::env;
+use std::ffi::{OsStr, OsString};
 use std::fs::OpenOptions;
 use std::io::{self, BufRead, Write};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


cjustice:
On the mirrored `SENSITIVE_ENV_PARTS` copy — opened gakonst/nanocodex#59 to expose the predicate's result as `ambient_sensitive_environment()`, beside the predicate itself. Once that lands, this PR collapses to a `rev` bump plus:

```rust
let tools = Tools::default()
    .into_builder()
    .process_environment(ambient_sensitive_environment())
    .build()
    .map_err(|error| nanocodex_error(error.into()))?;
```

deleting the copied constant, the local `is_sensitive_name`, and two of the three tests here. Happy to hold this until then so it merges DRY, or land it now and follow up — whichever you prefer.

cjustice:
Correction to my comment above: the snippet should be `Tools::builder()`, not `Tools::default().into_builder()` — `ToolsBuilder` derives `Default` over `Tools`, so the two are identical and the shorter form is now what the PR uses.

cjustice:
Moved to draft — please don't merge this version.

It carries a copy of `SENSITIVE_ENV_PARTS` and `is_sensitive_name` only because they're private in nanocodex-tools, and gakonst/nanocodex#59 removes that need. Once #59 lands this becomes a `rev` bump plus one call, deleting 40-odd lines: the constant, the local predicate, and both remaining tests. Asking you to merge the duplicated version first would put a known drift hazard in the tree for no reason.

Two things from a self-review in the meantime, both applied:

- Dropped `TURBO_TOKEN` / `API_SECRET` / `FUNDER_TOKEN` from the test — those are our deployment's variables, not centaur's, and don't belong in this repo's test suite. It now asserts only the credentials api-rs puts in every sandbox.
- Dropped a test that asserted every forwarded name satisfies the predicate the forwarding filter is built from. It couldn't fail.

Also worth flagging: `crates/harness-server` isn't covered by CI — the Rust job in `ci.yml` is scoped to `services/api-rs` — so nothing here gets verified automatically. Ran the equivalent gates locally: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test` (67 unit, 20 integration, 4 pre-existing ignored) all clean.

PR #1193, opened 2026-07-27
feat(console): add role Slack channel permissions

mslipper:
Adds role-owned Slack channel permissions and merges them with direct principal permissions for effective API, JWT, snapshot, and sandbox access. Extends the role API and console UI, invalidates assigned principals when role permissions change, and adds database constraints and regression coverage.


mslipper (commented):
.

PR #1196, opened 2026-07-27
fix(api-rs): disable Nanocodex A/B when Codex auth is not api_key

Ayush7614:
## Summary
- Clamp Nanocodex A/B rollout to 0 when `CODEX_AUTH_MODE` is not `api_key`.
- Remap explicit Nanocodex harness requests to Codex under the same auth modes so sandboxes are not started without `OPENAI_API_KEY`.

## Problem
Nanocodex requires `OPENAI_API_KEY` at harness startup. The shared sandbox env template only injects that placeholder when `CODEX_AUTH_MODE=api_key`. With ChatGPT/`access_token` auth and a non-zero Nanocodex rollout, assigned Nanocodex sessions fail with MissingEnvironment. Injecting the placeholder in access_token mode would incorrectly push Codex onto the API-key path.

## Fix
Treat Nanocodex as incompatible with non-`api_key` Codex auth: disable A/B assignment and fall back to Codex for explicit Nanocodex requests when auth mode cannot support Nanocodex.

## Verification
- `rustup run 1.95 cargo test -p centaur-api-server harness_rollout` (passed)
- `rustup run 1.95 cargo test -p centaur-api-server nanocodex_rollout` (passed)

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/args.rs
+++ b/services/api-rs/crates/centaur-api-server/src/args.rs
@@ -114,7 +114,10 @@ impl Args {
     }
 
     pub(crate) fn codex_nanocodex_rollout_percent(&self) -> u8 {
-        self.server.codex_nanocodex_rollout_percent
+        effective_codex_nanocodex_rollout_percent(
+            self.server.codex_nanocodex_rollout_percent,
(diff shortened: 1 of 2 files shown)


mslipper:
Superseded by #1218.

PR #1207, opened 2026-07-28
feat(chart,api-rs): allow resource requests/limits for api-rs-managed pods

ivpusic:
## Summary

- wire Kubernetes resource requests/limits for every pod api-rs creates at runtime, each class sized independently:
  - session sandbox pods (per-session + warm): `sandbox.resources` → `SESSION_SANDBOX_{CPU,MEMORY}_{REQUEST,LIMIT}`
  - workflow-host sandbox pod: `apiRs.workflowHostResources` → `WORKFLOW_HOST_{CPU,MEMORY}_{REQUEST,LIMIT}`
  - per-sandbox iron-proxy pods: `ironProxy.resources` → `KUBERNETES_IRON_PROXY_{CPU,MEMORY}_{REQUEST,LIMIT}`
- `sandbox.resources` has been dead config since the Rust migration: no template read it and nothing reached api-rs, while `SandboxSpec.resources` was only ever set by a unit test. Its inert default (`requests: cpu=100m/memory=512Mi`) is replaced with an empty map so defaults stay behavior-preserving.
- replace the test-only, limits-only `ResourceLimits` on `SandboxSpec` with `ResourceRequirements` (independent requests/limits as k8s quantity strings); partial specs (e.g. memory request==limit, no cpu) work and unset keys render nothing
- fix the stale `KUBERNETES_SANDBOX_*` row in the configuration docs

Re-opens the concern from #420 (closed by #725, which didn't add resource wiring) and replaces #423, which targeted the removed Python API.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` (DB-backed tests skipped: no local test Postgres)
- `helm lint contrib/chart`
- `helm template` with values unset renders no resource env; with partial values it renders exactly the set keys


mslipper:
Thanks for this! I updated the API to match how we've implemented other configuration values that pass through YAML to k8s.

PR #1208, opened 2026-07-28
feat: add Loki log queries to grafana tool

akandic47:
## Why

The grafana tool queries VictoriaMetrics (`query`) and VictoriaLogs (`vlogs`) through the Grafana datasource proxy, but has no path for Grafana Loki: `query` builds Prometheus-style paths (`/api/v1/query[_range]`) and `vlogs` builds VictoriaLogs paths (`/select/logsql/...`), so neither works against a Loki datasource UID. Loki serves logs at `/loki/api/v1/query_range` with LogQL.

## What

- `GrafanaClient.query_loki(query, datasource_uid="loki", start, end, limit)` runs LogQL range queries via `/api/datasources/proxy/uid/<uid>/loki/api/v1/query_range` and flattens `streams` responses into time-sorted entries (RFC3339 `time`, `stream` labels, `line`), tolerating Loki 3.x structured-metadata value tuples.
- Metric LogQL responses (`resultType: matrix`) are rejected with a clear error instead of being misparsed as logs (seconds-vs-nanoseconds timestamps, dropped labels).
- `GrafanaClient.loki_labels()` / `loki_label_values(label)` mirror the metric label helpers.
- `loki` CLI command mirroring `vlogs` (`--ds/--start/--end/--limit/--json`), same output shape (time, labels, line).
- Docs: grafana rows in the tool directory mention Loki; md mirror regenerated.

## Testing

- `test_client.py`: 7 `httpx.MockTransport` tests covering request construction (path, params, start/end handling, datasource UID), stream flattening and time sorting, matrix rejection, label endpoints, and HTTP error propagation.
- `ruff check` clean; `ruff format` clean on the new/changed files.
- Validated live against a self-hosted Grafana + Loki through the datasource proxy (host-scoped API key, real LogQL queries).

### DIFF
--- a/docs/pages/reference/tool-directory.mdx
+++ b/docs/pages/reference/tool-directory.mdx
@@ -35,7 +35,7 @@ These are broadly useful across most deployments and are good candidates to conf
 | `gsuite` | Use Gmail, Calendar, Drive, Docs, Sheets, Slides, and Google Analytics | `GOOGLE_TOKEN_JSON` |
 | `websearch` | Free web search via Parallel and deep research | None; `PARALLEL_API_KEY` for `deep_research`; `ANTHROPIC_API_KEY` for search synthesis |
 | `company_context` | Search indexed company history, private Slack conversations, and Google Docs | None |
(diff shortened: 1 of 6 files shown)


mslipper:
Thanks for looking into this. We'll be replacing some of these tools with native CLIs like LogCLI shortly, so I'll be closing this for the time being.

PR #1220, opened 2026-07-29
fix: deliver Slack root DM replies

dcbuild3r:
## What changed

- route root Slack DMs through the normal top-level message delivery path
- preserve the stable `slack:<channel>:` session key for continuous DM context
- use the same DM-safe delivery behavior during render recovery
- add an emulated root-DM regression test

## Root cause

The Slack adapter intentionally represents a root DM with an empty thread timestamp so every message in that DM shares one durable Centaur session. Slack's native streaming API rejects that empty timestamp because streaming requires a thread. The execution completed, but the first render attempt failed before a message became visible and had to fall through the generic recovery path.

Root DMs now collect the durable renderer output and post the final answer as a normal top-level message. Channel threads continue to use native Slack streaming.

## Validation

- `bun run --cwd services/slackbotv2 check:types`
- `bun test services/slackbotv2/test`
  - 221 passed, 1 skipped
- built `services/slackbotv2/Dockerfile` for `linux/amd64`
- imported the image into the home-lab k3s node
- rolled out `centaur-centaur-slackbotv2` successfully

Live DM delivery verification is pending a post-rollout user message.

### DIFF
--- a/services/slackbotv2/src/index.ts
+++ b/services/slackbotv2/src/index.ts
@@ -2317,7 +2317,7 @@ async function renderExecutionStream(
   consoleSessionBlock?: SlackContextBlock
 ): Promise<{ diverged: boolean; messageId?: string }> {
   const promptText = slackMessagePromptText(message)
-  if (isPlainTextOnlyRequest(promptText)) {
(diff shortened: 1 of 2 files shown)


dcbuild3r:
please ignore, agent was misled and placed an internal fork PR on upstream

PR #1224, opened 2026-07-29
feat: thread node steering into sandbox pods

tobernguyen:
## Problem

Sandbox pods are created at runtime by `api-rs`, not by the chart. So nothing the chart renders can reach them, and neither a Helm nor a kustomize patch can touch them either. That leaves an operator with no way to:

- keep sandboxes on a **dedicated node pool** — commonly a tainted, sandbox-only pool, so untrusted agent workloads never share a node with the control plane, and so a spot reclaim cannot take out a live session; or
- select a **stronger sandbox runtime** via `runtimeClassName` (e.g. gVisor) for untrusted agent code.

Related: `sandbox.runtimeClassName` already exists in `values.yaml` and `values.schema.json`, but is referenced by no template — setting it silently does nothing today.

## Approach

`Sandbox.spec.podTemplate.spec` already accepts `nodeSelector`, `tolerations` and `runtimeClassName`, so this is plumbing rather than new API — no CRD change, no new abstraction. Three chart values reach `api-rs` as env and are passed straight through to the CR:

| Chart value | Env | Shape |
| -- | -- | -- |
| `sandbox.nodeSelector` | `SESSION_SANDBOX_NODE_SELECTOR` | JSON object of label key/value pairs |
| `sandbox.tolerations` | `SESSION_SANDBOX_TOLERATIONS` | JSON array, Kubernetes toleration shape |
| `sandbox.runtimeClassName` | `SESSION_SANDBOX_RUNTIME_CLASS_NAME` | string (existing value, now wired up) |

```yaml
sandbox:
  nodeSelector:
    workload: centaur-sandbox
  tolerations:
    - key: example.com/sandbox
      operator: Exists
      effect: NoSchedule
  runtimeClassName: gvisor
```

## Two decisions worth flagging for review

**Malformed input fails startup rather than warning-and-ignoring**, which differs from `SESSION_SANDBOX_EXTRA_ENV` deliberately. Silently dropping a `nodeSelector` schedules sandboxes wherever the default scheduler chooses — precisely the outcome an operator setting it is trying to prevent, and a silent one. The asymmetry is commented at both call sites so it does not later read as an inconsistency. Happy to switch to warn-and-ignore for consistency if you would rather.

**Tolerations are carried as raw JSON** (`Vec<serde_json::Value>`) rather than a typed struct, so the toleration shape stays defined in one place — the CRD types. Malformed entries surface as the existing `InvalidSpec` error when the CR is deserialized. If you would prefer a typed value in `AgentSandboxConfig`, that is an easy change.

## Compatibility

All three fields are omitted from the CR when unset, so default behaviour is unchanged.

## Tests

- `node_steering_reaches_the_sandbox_pod_template` — all three land in `podTemplate.spec`
- `node_steering_is_omitted_when_unset` — nothing emitted by default
- `node_steering_parses_from_json`, `node_steering_is_empty_when_unset`, `invalid_node_steering_is_rejected` — arg parsing and the fail-fast behaviour

`cargo test -p centaur-sandbox-agent-k8s -p centaur-api-server` passes (48 + 90 + 50), `cargo fmt --check` clean, and `helm template` verified to emit all three env vars when set and none at defaults.

### DIFF
--- a/contrib/chart/templates/apirs.yaml
+++ b/contrib/chart/templates/apirs.yaml
@@ -330,6 +330,21 @@ spec:
             - name: SESSION_SANDBOX_IMAGE_PULL_SECRETS
               value: "{{- range $index, $secret := .Values.global.imagePullSecrets }}{{- if $index }},{{ end -}}{{ $secret.name }}{{- end }}"
 {{- end }}
+            # Node steering for sandbox pods. Passed to the control plane rather
+            # than templated onto a pod: api-rs creates sandbox pods at runtime,
+            # so nothing this chart renders can reach them.
+{{- with .Values.sandbox.nodeSelector }}
+            - name: SESSION_SANDBOX_NODE_SELECTOR
+              value: {{ toJson . | quote }}
+{{- end }}
+{{- with .Values.sandbox.tolerations }}
+            - name: SESSION_SANDBOX_TOLERATIONS
+              value: {{ toJson . | quote }}
+{{- end }}
+{{- with .Values.sandbox.runtimeClassName }}
+            - name: SESSION_SANDBOX_RUNTIME_CLASS_NAME
+              value: {{ . | quote }}
+{{- end }}
             # Sandboxes call back into this control plane via the in-cluster Service.
             - name: SESSION_SANDBOX_CENTAUR_API_URL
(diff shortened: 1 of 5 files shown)


mslipper:
Superseded by #1273.

PR #1234, opened 2026-07-31
feat: add registry-backed MPP charges

brendanjryan:
## Motivation

Give Centaur a safe fallback for paid APIs when no native tool exists, without exposing wallet credentials to sandboxes or teaching iron-proxy payment semantics.

## Summary

- Replace the market-specific `mpp` tool with registry-backed `list`, `search`, `show`, `request`, and `health` commands plus a bounded last-known-good cache and streamed response limits.
- Add one cluster-internal Tempo charge signer using the global private key, active execution leases, durable idempotency and audit state, conservative completion accounting, optional single-currency atomic budgets, chain pinning, metrics, traces, and health checks.
- Configure iron-proxy v0.50.0-rc.2 for one exact 402 replay through authenticated authorize and complete callbacks, with ironsh/iron-proxy#221 adding the original upstream scheme required by the signer.
- Wire the signer, dedicated Secret, restart checksums, NetworkPolicies, budget currency, and Prometheus scraping through Helm.
- Add a deterministic paid origin for local end-to-end validation.

### Product contract

- v1 supports only Tempo `charge`; there is no wallet onboarding or per-user opt-in.
- Only registered GET routes execute by default. Operator allow rules enable mutating routes, and deny rules always win.
- Requests accept only a registry service ID, registered method and path, path parameters, query parameters, and an optional JSON body. Arbitrary URLs, plaintext HTTP, redirects, host substitution, sensitive headers, overlapping routes, and path traversal are rejected.
- Live 402 terms are authoritative over registry pricing. Missing budgets mean no software spending cap and emit a warning and metric. Configured caps require one currency so atomic units are never combined across assets.
- The wallet key is mounted only into the signer. Paid response bodies are never cached, response reads are bounded while streaming, and audit records contain bounded metadata and hashes only.
- Registry entries are fresh for 15 minutes, discovery may use clearly marked stale data for 24 hours, and paid execution fails closed after 24 hours. Signer refreshes are time-bounded, backed off after failure, and never block readiness probes.
- The signer is pinned to the configured Tempo chain, and Secret rotation restarts it for both token and wallet-key changes.
- Only an explicit replay 402 releases a reservation. Transport failures and other ambiguous replay outcomes remain charged against the daily cap as `unknown`; stale in-flight reservations are reconciled conservatively.

### Service-selection waterfall

1. Prefer a purpose-built native Centaur tool when one is installed.
2. Otherwise use `mpp list`, `mpp search`, or `mpp show` to identify registry candidates and expose why routes are unavailable.
3. `mpp request` resolves an exact service ID and registered route locally, then sends ordinary HTTPS through sandbox iron-proxy.
4. On a 402 response, iron-proxy forwards the exact scheme, authority, method, path/query, challenge, sandbox identity, replayability, and trace metadata to signer `/authorize`.
5. The signer requires a live execution lease, re-resolves the route from its authoritative last-known-good registry, and applies scheme, chain, intent, policy, budget, and idempotency checks before returning payment headers.
6. iron-proxy performs one exact replay and reports its status and receipt metadata to signer `/complete`.
7. The signer settles, explicitly releases, or conservatively retains the reservation and records bounded audit, metric, and trace data. If any step declines, the original 402 is returned.


brendanjryan:
Follow-up hardening is pushed in `0cf3a859`:

- accepts registry provider realms that differ from HTTP authorities and preserves registered base paths/ports
- validates central/client cache schemas and keeps invalid refreshes from replacing last-known-good data
- isolates `MPP_PRIVATE_KEY` to the signer Secret; proxy/signer auth now uses the normal infra Secret
- removes in-memory payment credential retention, binds duplicate challenges to the original sandbox/execution, and adds atomic Postgres budget concurrency coverage
- bounds metric labels, redacts execution identities, reconciles reservation gauges from Postgres, and links charges to Centaur thread traces

A fresh local Kind run used a real managed sandbox and shared signer: registry discovery cached on the second call, the controlled origin observed exactly `402 -> 200`, the audit row settled with a receipt, POST/arbitrary-service calls were denied, an expired execution lease returned the original 402 without creating an attempt, and only the signer pod referenced the private key. The disposable namespace and key Secret were removed afterward. Live Moderato remains pending because `MODERATO_PROXY_PASSWORD` is not available in the local environment or accessible 1Password session.

mslipper:
Closing for the time being - @brendanjryan please reopen once you're ready for review here.

PR #1246, opened 2026-08-01
feat(githubbot): emit durable workflow events from CI and review webhooks

cjustice:
## Summary

Durable workflows can `ctx.wait_for_event(...)` (the Python host + Rust bridge + `POST /api/workflows/events` emission route all shipped), but nothing produces lifecycle events today — waiters still have to poll. This makes githubbot the producer:

- **`ci-completed`** — emitted once **every check for a head sha has settled** (aggregate `evaluateCi` over `listForRef` + combined status, the same gate the owned-PR path trusts; the webhook's completed/terminal action is only a cheap pre-filter). Correlation: `<owner>/<repo>:<head_sha>`. Payload `{failed, failing}`.
- **`review-submitted`** — emitted when a configured reviewer bot (default `chatgpt-codex-connector`, override via `GITHUBBOT_WORKFLOW_REVIEW_BOTS`) submits a review. Correlation: `<owner>/<repo>:pr-<n>:<head_sha>`, with the head sha from `fetchPr` (the webhook payload doesn't carry it). Payload `{review_id, state}`.

Both correlation ids are **lowercased** so case drift between a PR URL slug and `repository.full_name` can never miss a waiter (waiters lowercase their side too).

Two engine semantics shape the design:

- **Durable events are immutable — first write wins per correlation** (`absurd.emit_event` only applies the `NULL → payload` transition). So `ci-completed` must never fire on a single check's completion: that would lock the event row while the rest of the suite is still running, and waiters could never receive the real one. The settled aggregate is the only safe emission point.
- Keying both on **head sha** means a stale event for an old push can never wake a waiter that already moved on.

Placement is deliberate:

- Emission sits **before the owned-PR gate** in `handleCiEvent` / `handleReviewEvent` — durable workflow waiters are not bot-owned PRs, so gating on ownership would silently drop their wake-ups.
- The settled evaluation is computed once and shared with the owned-PR path (`processCi`), so an owned PR doesn't pay a second `listForRef` + combined-status round-trip per CI event.
- The aggregate payload can't see *required* checks, so the event is the wake-up, not the verdict — waiters still verify with one live read.
- Emission is best-effort and never fails the webhook handler; a missed event degenerates to the waiter's timeout path.
- Off by default (`githubbot.workflowEvents` / `GITHUBBOT_WORKFLOW_EVENTS`); api-rs URL and service token are already wired for the session API.

Also closes a coverage gap the design validation for this found: `absurd-sdk`'s `await_event` / `emit_event` had no database-backed test. The new test exercises early-emit-resolves-later-wait, wait→emit→wake with payload, checkpointed re-await, and timeout-as-error against a disposable Postgres (`ABSURD_TEST_DATABASE_URL`).

## Validation

- `pnpm --filter githubbot run check:types` ✓, `pnpm --filter githubbot test` — 86 pass (77 baseline + 9 new emission tests) ✓
- `helm lint contrib/chart` ✓ (chart version bumped)
- `ABSURD_TEST_DATABASE_URL=… cargo test -p absurd-sdk` — 8/8 incl. the new DB test ✓; `cargo fmt --check` + `cargo clippy -p absurd-sdk --all-targets -- -D warnings` ✓

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.108
+version: 0.1.109
 appVersion: "0.1.0"
 dependencies:
   - name: connect
--- a/contrib/chart/templates/githubbot.yaml
+++ b/contrib/chart/templates/githubbot.yaml
@@ -98,8 +98,7 @@ spec:
               value: {{ .Values.githubbot.autoMerge | quote }}
             - name: GITHUBBOT_MERGE_METHOD
               value: {{ .Values.githubbot.mergeMethod | quote }}
-            # Durable-workflow event emission (ci-completed / codex-review ->
-            # POST /api/workflows/events on api-rs).
+            # Durable workflow-event emission (ci-completed / codex-review) to api-rs.
             - name: GITHUBBOT_WORKFLOW_EVENTS
               value: {{ .Values.githubbot.workflowEvents | quote }}
 {{- if .Values.githubbot.workflowReviewBots }}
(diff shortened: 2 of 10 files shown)


cjustice:
Superseded by #1247 — restructured after design feedback: the emission logic is now a standalone workflow-event producer module independent of the owned-PR manager, with the engine's curation rules (semantically-complete events, emission-time filtering, correlation convention) documented as the contract for future events. Same feature, clean history.

PR #1266, opened 2026-08-03
feat(slackbotv2): decouple response metadata from Console

snario:
## Summary

- add `slackbotv2.responseMetadataEnabled`, independent of `CENTAUR_CONSOLE_PUBLIC_URL`
- append model, harness, reasoning effort, and service tier metadata to every live streamed Slack response when enabled
- preserve the existing first-message Console link behavior
- report Codex's configured service tier only when it is authoritative; omit it for alternate providers and Nanocodex

## Why

The Slack response footer was coupled to the optional public Console URL, so self-hosted installations without a Console URL could not show model or reasoning metadata at all. Service tier was also missing.

## Validation

- `pnpm --filter slackbotv2 run check:types`
- `pnpm --filter slackbotv2 test` (227 passed, 1 skipped)
- signed emulated Slack webhook coverage for first and subsequent responses without a Console URL
- Bun production bundle
- Helm dependency build, lint, and template render with `slackbotv2.responseMetadataEnabled=true`
- `git diff --check` and chart schema/YAML parsing

### DIFF
--- a/contrib/chart/templates/slackbotv2.yaml
+++ b/contrib/chart/templates/slackbotv2.yaml
@@ -79,6 +79,8 @@ spec:
               value: {{ .Values.apiRs.activitySummary.enabled | quote }}
             - name: SLACKBOTV2_AUTO_JOIN_CREATED_CHANNELS
               value: {{ .Values.slackbotv2.autoJoinCreatedChannels | quote }}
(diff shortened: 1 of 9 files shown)


mslipper:
Superseded by #1275, which preserves this commit and adds independent controls for footer frequency and service-tier display.

PR #1270, opened 2026-08-04
feat(api-rs): bind the requesting user's principal per turn

ivpusic:
Step 4 of 4 towards RFC 0005 (#1256). Builds on #1258 and #1269, so their commits show here until they merge; review the last commits. Requester eligibility also depends on `slack_home_team_id` from #1360.

On every execute, api-rs now resolves who is asking and binds that user principal to the sandbox proxy for that turn, next to the unchanged conversation principal. The console (#1269) then serves the union.

Only Slack channel turns from users in the app home workspace bind a requester. api-rs requires nonblank, matching `slack_team_id` and `slack_home_team_id`, so Slack Connect users and ambiguous requests retain channel credentials only. In a DM the conversation principal already is the user, and other surfaces have no requester.

The binding is sent on every assignment, null included, so a turn with no requester clears the previous one. If the lookup fails the turn runs with no requester rather than failing. A user gets the same principal whether first seen in a DM or channel.

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/lib.rs
+++ b/services/api-rs/crates/centaur-api-server/src/lib.rs
@@ -60,6 +60,15 @@ mod tests {
             Ok(test_principal("prn_test"))
         }
 
+        async fn register_requester(
+            &self,
+            _thread_key: &str,
+            _metadata: Option<&Value>,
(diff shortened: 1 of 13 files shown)


mslipper:
Looks mostly ready, but I found some edge cases around Slack Connect that I'll be closing. Specifically, the requester credentials misattribute the user's organization when a Slack Connect user responds in thread.

PR #1273, opened 2026-08-04
fix: honor sandbox nodeSelector, tolerations, and runtimeClassName on agent and proxy pods

buffalu:
## Summary

Sandbox and iron-proxy pods are created at runtime by `api-rs`, not by Helm. Chart values `sandbox.nodeSelector`, `sandbox.tolerations`, and `sandbox.runtimeClassName` were therefore inert: nothing passed them into the Sandbox CR or the per-session proxy pod. Operators could set gVisor and a tainted agents pool in values and still schedule untrusted workloads on the default path under ordinary `runc`.

This wires those three values through api-rs env (`SESSION_SANDBOX_NODE_SELECTOR`, `SESSION_SANDBOX_TOLERATIONS`, `SESSION_SANDBOX_RUNTIME_CLASS_NAME`) into **both** the agent Sandbox pod template and the paired iron-proxy pod. Malformed selector/toleration JSON fails startup instead of silently dropping placement. Empty values keep default scheduling.

Chart version bumped to `0.1.113` (rebased onto current main at `0.1.112`).


buffalu:
**Ops note from jito-labs carry (verified on GKE):** once this lands, setting `sandbox.runtimeClassName: gvisor` together with the usual repo-cache **hostPath** mount (`/var/lib/centaur/repos`) makes Sandbox CRs fail Ready with:

`pods \"asbx-…\" is forbidden: HostPath is not allowed: \"/var/lib/centaur/repos\"`

Placement (`nodeSelector` / `tolerations`) works; RuntimeClass=gvisor is what trips GKE Sandbox. We are pinning sandboxes to our agents pool **without** RuntimeClass until repo-cache is non-hostPath (CSI/NFS). Worth a docs callout in the production deploy section for operators using hostPath + gVisor.

buffalu:
**Follow-up ops note:** with hostPath repo-cache + a tainted sandbox node pool, `repoCache.tolerations` must match the sandbox taints. Otherwise the DS stays off those nodes, hostPath is empty, and sandbox `tools-bootstrap` fails with `tools repo-cache entry unavailable`. Chart already supports `repoCache.tolerations` — worth mentioning next to the placement docs.

PR #1279, opened 2026-08-06
fix(websearch): enable Parallel REST/deep_research under iron-proxy

buffalu:
## Summary
- Switch `PARALLEL_API_KEY` from **inject** to **replace** (`Authorization` on `api.parallel.ai` / `search.parallel.ai`) so sandboxes get the `PARALLEL_API_KEY=PARALLEL_API_KEY` placeholder and iron-proxy swaps the real value.
- Fix `_is_configured` for CLI/shim invokes: Centaur tool shims leave `ToolContext.secrets` empty; treat non-empty env presence as configured (replace placeholders are `KEY=KEY`). StubBackend’s value-equals-key fallback still means “not configured” when the env var is absent.

## Why
With inject-mode, the real key can sit on iron-proxy while `websearch deep-research` still errors with “PARALLEL_API_KEY is not available” and `search` stays on free `parallel:mcp`. Replace-mode matches how harness API keys (OPENAI/ANTHROPIC) already work.

## Test plan
- [x]  Test with local overlay

### DIFF
--- a/tools/research/websearch/README.md
+++ b/tools/research/websearch/README.md
@@ -27,10 +27,10 @@ That's the minimum — works with zero credentials via the free MCP. Add a `PARA
 
 ## Secrets
 
-Set in root `.env` (preferred) or `tools/research/websearch/.env`.
+Set in root `.env` (preferred) or `tools/research/websearch/.env` for local CLI use.
 
-- `PARALLEL_API_KEY` — optional. Get one at <https://platform.parallel.ai>. Unlocks the paid Search REST path (with source filters) and enables `deep_research` via the Task API.
-- `ANTHROPIC_API_KEY` — optional. Enables the Claude-backed synthesis pipeline on `search(synthesize=True)`.
+- `PARALLEL_API_KEY` — optional. Get one at <https://platform.parallel.ai>. Unlocks the paid Search REST path (with source filters) and enables `deep_research` via the Task API. In Centaur sandboxes this is a **replace**-mode secret (`Authorization`): grant it (e.g. `centaur-perms roles grant infra --tool websearch --secret-name PARALLEL_API_KEY`) so agents see the `PARALLEL_API_KEY=PARALLEL_API_KEY` placeholder and iron-proxy swaps the real value.
+- `ANTHROPIC_API_KEY` — optional. Enables the Claude-backed synthesis pipeline on `search(synthesize=True)` (inject-mode toward `api.anthropic.com`).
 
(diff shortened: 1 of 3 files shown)


buffalu:
verified fixed locally

mslipper:
Why can't this be injected?

buffalu:
> Why can't this be injected?

This can't be injected because Parallel needs the tool to know the key exists before it talks to anything. [_is_configured](https://github.com/paradigmxyz/centaur/blob/main/tools/research/websearch/client.py#L39-L63) decides whether we have a key. Inject never puts a placeholder in the env, so it always says no, even if the tool is sitting on the proxy.

[init](https://github.com/paradigmxyz/centaur/blob/main/tools/research/websearch/client.py#L84-L104) builds ParallelBackend with api_key=None, [use_rest](https://github.com/paradigmxyz/centaur/blob/main/tools/research/websearch/_parallel.py#L190) stays false (free Search MCP), and [deep_research](https://github.com/paradigmxyz/centaur/blob/main/tools/research/websearch/_parallel.py#L317-L331) just hard-fails — MCP can't do deep research, that needs the Task API.

So with inject the key is on the proxy but invisible to the client, and the client picks the wrong path. Replace is what gives the sandbox PARALLEL_API_KEY=PARALLEL_API_KEY so routing actually works.

mslipper:
This isn't always true. Warm sandboxes, for example, receive only "infrastructure" env vars like OPENAI_API_KEY. They won't get the PARALLEL_API_KEY defined here, so this fix is incomplete for that case.

Broadly, we'd like to move away from replace-type secrets for exactly this reason. Warm sandboxes can't change their env vars once they're running, so only the things that are absolutely necessary and common between each sandbox principal should be put on the env.

I've opened https://github.com/paradigmxyz/centaur/pull/1303 to resolve this issue without requiring replace keys. Please take a look and try it. Closing this PR in the meantime; happy to revisit if 1303 doesn't work for you.

buffalu:
> This isn't always true. Warm sandboxes, for example, receive only "infrastructure" env vars like OPENAI_API_KEY. They won't get the PARALLEL_API_KEY defined here, so this fix is incomplete for that case.
> 
> Broadly, we'd like to move away from replace-type secrets for exactly this reason. Warm sandboxes can't change their env vars once they're running, so only the things that are absolutely necessary and common between each sandbox principal should be put on the env.
> 
> I've opened #1303 to resolve this issue without requiring replace keys. Please take a look and try it. Closing this PR in the meantime; happy to revisit if 1303 doesn't work for you.

switching to injected only + making tools act the same based on if an env var is passed in or not makes a lot of sense to me. ill try 1303 out shortly + report back.

PR #1288, opened 2026-08-06
fix(codex): disable subagents via agents.enabled=false

### LINKED ISSUE #1287: Codex 0.144+: features.multi_agent=false does not disable spawn_agent; child turn/completed closes parent
## Summary

Centaur already tries to disable Codex multi-agent via:

- baked `harness/codex/config.toml`: `features.multi_agent = false`, `features.multi_agent_v2 = false`
- sandbox `entrypoint.sh` force-rewriting those same feature flags to `false`

On **Codex 0.144+**, that is not a hard off. Collaboration tools (`spawn_agent`, `wait_agent`, etc.) can still be available, and a collab child's `turn/completed` can be treated as the parent Centaur execution finishing — dropping later parent tool calls from `session_events` / the console UI even though they still ran in the sandbox.

## Codex 0.144+ behavior

Effective multi-agent version resolves roughly as:

1. If `features.multi_agent_v2` → force V2
2. Else if `agents.enabled = false` → force **Disabled**
3. Else → use the **model catalog** `multi_agent_version`, then fall back to features

With Centaur's current defaults:

- `features.multi_agent{,_v2} = false`
- `agents.enabled` unset → defaults to **`true`**
- `gpt-5.6-sol` (and similar) catalog: `multi_agent_version = v2`

…the session still gets `multi_agent_version: v2` / `explicitRequestOnly`. Skills that ask to spawn an isolated agent (e.g. pre-push review) call `spawn_agent` successfully.

`codex features list` can show `multi_agent` / `multi_agent_v2` as `false` while the live turn context still reports V2 and `spawn_agent` works. The hard kill switch is:

```toml
[agents]
enabled = false
```

## Impact we hit

Console thread opened a PR successfully, but the console never showed the push/`gh pr create` tool call.

What happened:

1. Parent Codex turn committed, then spawned collab subagent `/root/review_llm_pr` (`spawn_agent` + `wait_agent`).
2. Child finished with final answer `"No material issues."` and emitted `turn/completed` on the **child** thread id.
3. Centaur marked the **parent** execution completed at that moment (`completion_reason: turn_completed`, `result_text: "No material issues."`).
4. ~2s later the parent continued in Codex (`git push --set-upstream origin HEAD && gh pr create …`), which succeeded on GitHub.
5. Those post-completion tool calls and the parent's real final answer never landed in Centaur `session_events`, so the UI looked like the turn was already done with no push tool call.

So there are two related bugs:

1. **Config:** intended multi-agent disable via feature flags alone is ineffective on Codex 0.144+ for models that advertise `multi_agent_version`.
2. **Runtime:** child/collab `turn/completed` should not complete the parent Centaur execution / close the event stream.

## Suggested fix

- Force `agents.enabled = false` in the baked Codex config and the same entrypoint path that already clears `features.multi_agent{,_v2}` (operators can still opt back in via `CODEX_CONFIG_OVERLAY` if desired).
- Fix session completion so only the **parent** turn's terminal `turn/completed` (or equivalent) closes the Centaur execution; ignore / scope child thread completions.

buffalu:
## Summary
- Set `[agents] enabled = false` in the baked Codex harness config and force the same key in the sandbox entrypoint (alongside the existing `features.multi_agent{,_v2}=false` rewrite).
- **Codex 0.144+:** clearing the multi-agent feature flags is insufficient. When `agents.enabled` stays at its default (`true`) and the model catalog advertises `multi_agent_version` (e.g. `gpt-5.6-sol` → `v2`), collaboration tools remain available in `explicitRequestOnly` mode, so skills can still call `spawn_agent`.
- **Impact:** a parent turn that spawned a collab review subagent was closed when the **child** emitted `turn/completed` (`completion_reason: turn_completed`, result text from the child). The parent continued afterward (`git push` / `gh pr create`), but those tool calls and the final answer never made it into Centaur `session_events` / the console UI — so the turn looked finished with no push tool call even though GitHub was updated.

`CODEX_CONFIG_OVERLAY` can still re-enable after the entrypoint force if a deployment wants subagents back.

Fixes #1287

### DIFF
--- a/harness/codex/config.toml
+++ b/harness/codex/config.toml
@@ -19,6 +19,10 @@ enable_fanout = false
 runtime_metrics = true
 
 [agents]
+# Codex 0.144+: features.multi_agent{,_v2}=false is not enough — model catalog
+# multi_agent_version (e.g. gpt-5.6-sol → v2) still enables spawn_agent while
+# agents.enabled defaults to true. This is the hard kill switch.
+enabled = false
 max_depth = 2
 job_max_runtime_seconds = 1800
 
--- a/services/sandbox/entrypoint.sh
+++ b/services/sandbox/entrypoint.sh
@@ -194,6 +194,32 @@ else:
         rewritten.append(f"{name} = false")
     lines = lines[: features_start + 1] + rewritten + lines[features_end:]
 
+# Codex 0.144+: features.multi_agent{,_v2}=false alone does not disable
+# collaboration tools when the model catalog advertises multi_agent_version
+# (e.g. gpt-5.6-sol → v2) and agents.enabled defaults to true. Force the hard
+# kill switch here; CODEX_CONFIG_OVERLAY can still re-enable after this.
+agents_start = next((i for i, line in enumerate(lines) if line.strip() == "[agents]"), None)
+if agents_start is None:
+    lines.extend(["", "[agents]", "enabled = false"])
+else:
+    agents_end = next(
+        (i for i in range(agents_start + 1, len(lines)) if lines[i].lstrip().startswith("[")),
+        len(lines),
+    )
+    seen_enabled = False
+    rewritten = []
+    for line in lines[agents_start + 1 : agents_end]:
+        stripped = line.strip()
+        name = stripped.split("=", 1)[0].strip() if "=" in stripped else None
+        if name == "enabled":
+            rewritten.append("enabled = false")
+            seen_enabled = True
+        else:
+            rewritten.append(line)
+    if not seen_enabled:
+        rewritten.insert(0, "enabled = false")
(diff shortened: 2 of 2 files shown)


buffalu:
Hold — this approach is wrong for Codex **0.144** (what Centaur sandboxes ship today).

In 0.144, `AgentsToml` has no `enabled` field; `enabled = false` under `[agents]` is deserialized via the flattened `roles` map as role `"enabled"` and fails with:

```
invalid type: boolean false, expected struct AgentRoleToml
```

That surfaces in Centaur as `terminal harness output reported failure`.

`agents.enabled` exists on newer Codex main, but not in 0.144. Closing this until we have a 0.144-safe disable (or a Codex bump).

buffalu:
Closing: breaks Codex 0.144 config load (see comment).

PR #1329, opened 2026-08-10
feat(console): paginate principal detail sections

gorried:
## Summary

- paginate assigned roles, direct grants, and effective grants independently on the principal detail page
- preserve each section's pagination state and return navigation to the relevant section anchor
- load effective-grant records and source attribution only for the current page

## Why

Principal detail pages rendered every assigned role and grant at once. On high-cardinality installations this produced large responses and unnecessary record loading even when an operator only needed the first page.

Each section now uses bounded 50-row offset pagination, following the existing Console list pagination pattern. Slack catalog behavior is intentionally unchanged and will be addressed separately.

## Validation

- `bundle exec rails test test/controllers/console_controller_test.rb test/controllers/console/workflows_controller_test.rb` (44 tests, 327 assertions)
- `bundle exec rubocop app/controllers/console_controller.rb app/services/principal_effective_grants_page.rb test/controllers/console_controller_test.rb`
- `git diff --check`

### DIFF
--- a/services/console/app/services/principal_effective_grants_page.rb
+++ b/services/console/app/services/principal_effective_grants_page.rb
@@ -38,16 +38,11 @@ class PrincipalEffectiveGrantsPage
 
   def effective_secret_counts
     kinds = relations.keys
-    unique_ids = kinds.to_h { |kind| [ kind, {} ] }
     columns = kinds.map { |kind| GRANT_COLUMN_BY_KIND.fetch(kind) }
+    aggregates = columns.map { |column| Grant.arel_table[column].count(true) }
+    counts = principal.effective_grants.pick(*aggregates)
 
-    principal.effective_grants.pluck(*columns).each do |ids|
-      ids.each_with_index do |id, index|
-        unique_ids.fetch(kinds.fetch(index))[id] = true if id
-      end
-    end
-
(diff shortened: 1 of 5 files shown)


services/console/app/services/principal_effective_grants_page.rb
```diff
@@ -0,0 +1,101 @@
...
+  private
+
+  attr_reader :principal, :relations, :requested_page, :per_page
+
+  def normalized_page
+    page = Integer(requested_page.to_s, 10, exception: false) || 1
+    page < 1 ? 1 : page
+  end
+
+  def effective_secret_counts
```
mslipper, line 39: I think this still pulls all rows into Ruby. You could do something like this and have the DB return them automatically:

```ruby
def effective_secret_counts
  kinds = relations.keys
  columns = kinds.map { |kind| "#{kind}_secret_id" }

  aggregates = columns.map do |column|
    Arel.sql("COUNT(DISTINCT grants.#{column})")
  end

  counts = principal.effective_grants.pick(*aggregates)
  kinds.zip(counts).to_h
end
```

gorried: Good catch, updated

PR #1333, opened 2026-08-11
feat: Hermes Agent as a first-class harness (sessions, learning loop, crons)

teknium1:
## Summary

Adds **Hermes Agent** (github.com/NousResearch/hermes-agent) as a first-class harness alongside Codex, Claude Code, Amp, and Nanocodex — `--hermes` in Slack, `harness: hermes` in channel defaults, `CENTAUR_DEFAULT_HARNESS=hermes` for deployments.

Hermes differs from the spawn-per-turn CLIs: it ships a long-lived JSON-RPC gateway that owns durable sessions, a background skill/memory learning loop, and a cron scheduler. This integration keeps those working inside the sandbox rather than flattening Hermes into a one-shot CLI:

- **Session continuity** — one persistent gateway child per sandbox, one Hermes session per thread; conversation history and provider prompt cache survive across turns. After a sandbox restart, `HERMES_CONTINUE_SESSION_ID` resumes the durable session from Hermes's session store (mirrors `CODEX_CONTINUE_THREAD_ID`).
- **Learning loop** — Hermes's post-turn memory/skill reviews run inside the gateway process and persist to `HERMES_HOME`; mount a volume there and learning compounds across sandboxes.
- **Crons** — a background ticker runs `hermes cron tick` (cross-process file-locked on Hermes's side) so scheduled jobs created in-conversation keep firing while the sandbox lives. `HERMES_CRON_TICK_SECONDS=0` disables.
- **Interrupts** — blocks `interrupt` maps to Hermes's `session.interrupt`: the turn ends Interrupted, the session survives for the next message.

## Changes

- `crates/harness-server/src/hermes.rs` — new blocks-mode runtime (codex shape: persistent child, not spawn-per-turn). Handshake `gateway.ready` → `session.create`; per-turn `prompt.submit`, events normalized into the shared `CodexTurnNormalizer`:
  | Hermes event | NormalizedEvent |
  |---|---|
  | `message.delta` | AgentTextDelta |
  | `reasoning.delta` / `thinking.delta` | ReasoningTextDelta |
  | `tool.start` | AssistantMessage(ToolUse) |
  | `tool.complete` | ToolResults (error + exit-code mapping) |
  | `turn.usage` | TokenUsage |
  | `message.complete` | final AssistantMessage + Result (error path preserved verbatim) |
- `HarnessType::Hermes` through session-core, session-runtime (subcommand map), session-cli, api-server (proxy fragment + auth-mode arms), integration-test wire contract
[...]

PR #1336, opened 2026-08-11
feat: add Hermes ACP harness adapter

mslipper:
Adds Hermes ACP as a first-class harness across the harness server, session runtime, bot overrides, console, sandbox image, and Helm configuration. The adapter normalizes Hermes ACP streams into the shared App Server interface and supports native session restoration, permissions, steering, cancellation, model selection, and token usage. It also hardens stale-session recovery and failed-cancellation process cleanup to keep subsequent turns healthy.

The database constraint migration is split into #1337 and should land first.


mslipper:
Closing in favor of #1333.

PR #1344, opened 2026-08-11
perf(console): refresh Slack channel catalog asynchronously

gorried:
## Summary

- separate the synchronous Slack API client from a request-facing cached catalog provider
- serve fresh or stale cached channel snapshots immediately and refresh them through a deduplicated background job
- render an ID-entry fallback on a cold cache instead of blocking the principal or role detail page on Slack
- keep Slack tokens out of job arguments and preserve successful cached data across refresh failures

## Why

Principal and role detail requests previously fetched the full Slack channel catalog synchronously whenever the five-minute cache expired. A cold production request spent roughly one second fetching 912 channels, so Slack latency and timeouts directly delayed the page shell.

This introduces stale-while-refresh behavior: successful snapshots are fresh for five minutes and may be served stale for up to 24 hours, while cold failures are cached for 30 seconds to avoid retrying Slack on every request. Rendering the complete cached catalog into the channel selector remains a separate follow-up.

## Validation

- `bundle exec rails test` — 1,427 runs, 5,579 assertions, 0 failures
- `bundle exec rubocop` — 349 files, no offenses
- `bundle exec brakeman --quiet --no-pager --exit-on-warn --exit-on-error` — 0 warnings
- focused provider and job tests — 7 runs, 33 assertions, 0 failures

### DIFF
--- a/services/console/app/controllers/concerns/console/slack_channel_permission_management.rb
+++ b/services/console/app/controllers/concerns/console/slack_channel_permission_management.rb
@@ -5,7 +5,7 @@ module Console
     private
 
     def load_slack_channel_permission_form(owner)
-      @slack_channel_catalog = SlackChannelCatalog.fetch
+      @slack_channel_catalog = SlackChannelCatalogProvider.fetch
       @slack_channel_names = @slack_channel_catalog.channels.to_h { |channel| [ channel.id, channel.name ] }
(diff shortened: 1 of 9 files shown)


gorried:
Superseded by #1345, recreated from an upstream-owned branch now that the author has direct write access.

PR #1362, opened 2026-08-13
fix(granola): use replace-mode auth so the API key actually gets injected (#1243)

### LINKED ISSUE #1243: granola tool: API-key backend 401s because inject mode collides with the client's own Authorization header
**Version:** `centaur-0.1.108` (chart + images), api-rs control plane, per-sandbox iron-proxy.

## Summary

`tools/productivity/granola`'s API-key backend always returns 401 against `public-api.granola.ai`, even with a valid workspace key, because the secret is declared as `mode = "inject"` on the `Authorization` header while the tool's own client already sets that header.

## Detail

`tools/productivity/granola/pyproject.toml` declares:

```toml
secrets = [
    {type = "http", name = "GRANOLA_API_KEY", mode = "inject",
     inject_header = "Authorization", inject_formatter = "Bearer {{ .Value }}",
     hosts = ["public-api.granola.ai"]},
]
```

`client.py` independently sets the same header:

```python
self._api_key = api_key or secret("GRANOLA_API_KEY", "")
...
"Authorization": f"Bearer {self._api_key}",
```

In a sandbox the tool has no real value for `secret("GRANOLA_API_KEY", ...)`, so it emits an `Authorization` header with a non-credential value. Inject mode is designed to *add* a header the client doesn't send, not to correct one already present, so the upstream service sees the client's value and rejects it.

The contrast with two secrets that do work in the same deployment:

| secret | mode | header | result |
|---|---|---|---|
| `AFFINITY_API_KEY` | replace (`match_headers: [Authorization]`) | Authorization | works |
| `COINGECKO_API_KEY` | inject | `x-cg-demo-api-key` | works |
| `GRANOLA_API_KEY` | inject | Authorization | 401 |

Inject works when the header is one the client never sets. Replace works when the client sends a placeholder in a header it does set. Granola is the only one using inject on a header its client also sets.

## What we ruled out

- **Key validity.** The same key authenticates fine against `https://public-api.granola.ai/v1/notes` with `Authorization: Bearer <key>` from outside the cluster.
- **Stale credential in the proxy.** The iron-proxy sidecar's `GRANOLA_API_KEY` env hashes identical to the source Secret.
- **Host / header / prefix mismatch.** Tool client, our out-of-cluster reference client, and the injection rule all agree on `public-api.granola.ai`, `Authorization`, and the `Bearer ` prefix.

## Repro

1. Populate `GRANOLA_API_KEY` with a valid Granola workspace key.
2. Start a sandbox so its iron-proxy sidecar picks the value up.
3. Run `granola health`.

Observed:

```json
{"ok": false, "tool": "granola",
 "error": "Client error '401 Unauthorized' for url 'https://public-api.granola.ai/v1/notes?page_size=1'"}
```

## Caveat

This is inferential rather than directly observed: the proxy (correctly) doesn't log secret values, so we could not capture the actual outbound `Authorization` header. Everything else is consistent with the above, but a header-level trace on your side would confirm it.

## Suggested fix

Switch the declaration to replace mode so it matches the client, as `AFFINITY_API_KEY` does:

```toml
secrets = [
    {type = "http", name = "GRANOLA_API_KEY",
     match_headers = ["Authorization"], hosts = ["public-api.granola.ai"]},
]
```

Alternatively, drop the `Authorization` header from `client.py` when running in-sandbox and let inject own it.

## Question

Is the API-key backend still a supported path, or is `mcp.granola.ai` + OAuth now the only intended one? The docstring calls MCP "preferred", so if the API-key path is effectively deprecated it'd help to say so in the tool's docstring, since it's the path that runs when no OAuth grant exists.

sam-shridhar1950f:
## What

The granola REST tool always 401s against `public-api.granola.ai`, even with a valid workspace key. Fixes #1243.

## Why

The client sets its own `Authorization: Bearer <key>` header, where `<key>` comes from `secret("GRANOLA_API_KEY")`. But the secret was declared as `mode = "inject"` on that same header.

Inject mode is for adding a header the client does not send. When the client already sends the header, iron-proxy has nothing to swap, so the placeholder value goes upstream and Granola rejects it.

This is the only secret in the tree using inject on a header its own client also sets. The working ones (`SLACK_BOT_TOKEN`, `AFFINITY_API_KEY`) use replace mode on `Authorization`.

## The fix

Switch `GRANOLA_API_KEY` to replace mode:

```toml
# before
{type = "http", name = "GRANOLA_API_KEY", mode = "inject", inject_header = "Authorization", inject_formatter = "Bearer {{ .Value }}", hosts = ["public-api.granola.ai"]}

# after
{type = "http", name = "GRANOLA_API_KEY", match_headers = ["Authorization"], hosts = ["public-api.granola.ai"]}
```

With `mode` omitted the parser defaults to replace, and `replacer` defaults to the secret name. So iron-proxy scans the `Authorization` header for `GRANOLA_API_KEY` and swaps in the real value, which is exactly the placeholder the client puts there. This is now the same rule shape as the Slack tokens that work in production today.

## Tests

- `centaur-perms`: new `granola_api_key_translates_to_authorization_replace_rule` runs the real parser and translator on the shipped declaration and asserts it produces a replace rule on `Authorization` with no inject config. Full crate green (53 passed).
- granola tool: added a declaration regression test (fails on the old inject config) and a client test asserting it emits the placeholder in the matched header. Suite green (5 passed).

### DIFF
--- a/services/api-rs/crates/centaur-perms/src/tests.rs
+++ b/services/api-rs/crates/centaur-perms/src/tests.rs
@@ -473,6 +473,46 @@ fn translates_http_replace_to_static_input() {
 }
 
 #[test]
+fn granola_api_key_translates_to_authorization_replace_rule() {
+    // Regression for issue #1243: the granola REST client sets
+    // `Authorization: Bearer <placeholder>` itself, so GRANOLA_API_KEY must be a
+    // replace-mode secret whose placeholder iron-proxy swaps inside the header the
+    // client already sends — not inject-mode, which would add a second header and
+    // leave the placeholder in place (upstream 401). This asserts the exact
+    // declaration shipped in tools/productivity/granola/pyproject.toml.
+    let secrets = vec![
+        tools::parse_secret(
+            &entry(
+                r#"{type = "http", name = "GRANOLA_API_KEY", match_headers = ["Authorization"], hosts = ["public-api.granola.ai"]}"#,
+            ),
+            &[],
+        )
+        .unwrap(),
+    ];
+
+    // Parses as replace mode with the placeholder defaulting to the secret name —
+    // the same string the tool client emits as the Bearer token.
+    let ParsedSecret::Http(http) = &secrets[0] else {
+        panic!("expected http")
+    };
+    assert_eq!(http.mode, SecretMode::Replace);
+    assert_eq!(http.replacer, "GRANOLA_API_KEY");
+    assert_eq!(http.match_headers, vec!["Authorization".to_owned()]);
+
+    // Translates to a proxy rule that scans the Authorization header for that
+    // placeholder and swaps it — no inject config.
+    let out = translate::translate("tool-granola", &secrets, &SourcePolicy::env());
+    let SecretInput::Static(input) = &out.inputs[0] else {
+        panic!("expected static")
+    };
+    let replace = input.replace_config.as_ref().unwrap();
+    assert_eq!(replace.proxy_value, "GRANOLA_API_KEY");
+    assert_eq!(replace.match_headers, vec!["Authorization".to_owned()]);
+    assert!(input.inject_config.is_none());
+    assert_eq!(input.rules[0].host.as_deref(), Some("public-api.granola.ai"));
+}
+
+#[test]
 fn translates_gcp_auth_defaults_scopes_when_unset() {
     let secrets = vec![
         tools::parse_secret(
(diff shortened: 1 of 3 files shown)


mslipper:
Hi there,

This is not how how inject mode works. Inject mode _always_ injects the header. It will override whatever the downstream client sends, and will add the header if it is not provided. I will revert back to the original issue with details and to figure out the root cause of their authentication problems. Closing for now.

sam-shridhar1950f:
Thanks for the correction!

Issue #231, opened 2026-05-27
Centaur API stale Consumer Edge pods serving old method surface


gakonst:
Consumer Edge tool calls are intermittently served by stale Centaur API pods exposing the old BigQuery surface instead of the new GCS surface.

Slack thread: slack:TGM4GKHA8:C0AV0P904EP:1779851685.406749
Date: 2026-05-27 UTC

Evidence:
- Stale response: `consumer-edge.latest_version {"table":"brand_detail"}` returned `Method latest_version not found`, with old methods `check_access`, `describe_table`, `list_datasets`, `list_projects`, `list_tables`, `query`.
- Stale pod from vlogs: `centaur-centaur-api-665c996cdd-8c7p4`, pod-template-hash `665c996cdd`.
- Fresh response window: 10/10 `latest_version` calls hit GCS surface and returned `brand_detail` version `2026_05_25`; `check_access`, `list_shards`, and `read_table brand_detail limit=3` succeeded.
- Fresh pod from vlogs: `centaur-centaur-api-6c4d6f4c45-jt5nz`, pod-template-hash `6c4d6f4c45`.
- Follow-up closeout attempts flipped stale again: `find_brand_by_id {"brand_id":93561}` was stale 10/10, and `discover consumer-edge` remained stale 20/20.

Impact:
- Service appears to be load-balancing between old and new Consumer Edge tool implementations, or routing can stick to a stale pod.
- This blocks reliable validation of the GCS proxy workaround and produces contradictory method availability in the same Slack thread.

Requested infra checks/actions:
- Run `kubectl get rs -l app=centaur-api -o wide` and `kubectl rollout status deployment centaur-api`.
- Compare images for ReplicaSets serving pod hashes `665c996cdd` and `6c4d6f4c45`.
- Reap stale ReplicaSet/pods or run `kubectl rollout restart deployment centaur-api` as appropriate.
- Investigate why image updater / Deployment / Argo reconciliation is not reaping prior ReplicaSets after tool implementation changes.

Sandbox note: `kubectl` is not installed in the agent sandbox, so cluster state could not be checked directly. Linear create_issue also returned GraphQL 400, so this was filed in GitHub instead.

rosspeili:
Hey @gakonst, thi is my first time in Centaur, but I'd like to help with this issue.

[...]

PR #1366, opened 2026-08-13
fix: route company context by physical database and selector

PengDeng-Cyber:
## Summary
- keep `company_context` on the physical `ai_v2` database
- send `iron.route=company-context` from the tool so it selects the dedicated reader route
- retain every granted Postgres route in Console effective configs, including same-database routes
- document and test the physical-database plus logical-selector contract
- build the route-selector Iron Proxy binary from a pinned fork commit and checksum until upstream PR #231 is released

## Root cause
The physical database is `ai_v2`; there is no `centaur` database. Iron Proxy 0.49.0 and Console previously keyed Postgres routes only by database, so audit and company-context permissions targeting the same database could not coexist. The fix keeps the physical database unchanged and adds a fail-closed `iron.route` selector for same-database policy routes.

## Test plan
- `cargo fmt --all --check`
- `cargo clippy -p centaur-iron-proxy -p centaur-sandbox-agent-k8s -p centaur-api-server --all-targets -- -D warnings`
- `cargo test -p centaur-iron-proxy`
- `cargo test -p centaur-sandbox-agent-k8s`
- `cargo test -p centaur-api-server company_context_manifest_configures_company_context_reader`
- `cargo test -p centaur-perms translates_pg_dsn_to_input_with_roundtrip_foreign_id`
- Console `principal_sync_config_snapshot_test.rb` (50 runs, 189 assertions)
- `uv run ... pytest tools/productivity/company_context/tests -q` (54 passed)
- pinned route-selector builder stage produced `0.49.0-centaur-route-selector.1`

The full final-image build was attempted locally but the mirror stalled while pulling the existing upstream base layer; the pinned builder stage and binary version were verified successfully.


PengDeng-Cyber:
This fix is now also validated in the pinned 0.1.115 downstream release path: cyberconnecthq/centaur-overlay#406. The downstream patch replay, real-manifest regression test, tool-discovery tests, and company_context tests pass. Upstream workflows still require maintainer approval for this first-time fork contribution.

PR #1369, opened 2026-08-13
feat(console): add okta as an idp

pyoxa:
Adds Okta as a console SSO provider alongside Google and Slack, using OIDC discovery, PKCE, signed state, nonce-bound ID-token verification, and UserInfo for verified identity claims. 

Okta issuer and client credentials are configured through `CENTAUR_CONSOLE_OKTA_*`. 
Includes Helm wiring, documentation, and focused login/provisioning coverage.


pyoxa:
@mslipper can you take a look at this one?

mslipper:
Yep will do!

mslipper:
Okta's default is `client_secret_basic` - was it intentional to only support `client_secret_post`? If so please document. Otherwise this looks good to me.

pyoxa:
oh thanks! i always used the creds in body method. adding this rn!

pyoxa:
@mslipper does this look correct?

mslipper:
looks good but i'll need to test on staging first

pyoxa:
sure! thanks for taking a look. fwiw we've been driving this patch for a bit (https://github.com/bitwise-invest/centaur) and it's been fine. i think at least

PR #1387, opened 2026-08-14
fix: fold duplicate credentials from repeat OAuth consent

cjustice:
## Summary

A repeat OAuth consent for an already-connected account currently mints a second `BrokerCredential` that can never be reconciled, and leaves two wrapper secrets competing to inject `Authorization` for the same hosts.

`Oauth::FlowsController#upsert_credential` de-duplicates on `find_or_initialize_by(oauth_app:, provider_subject: identity[:subject])`. Three providers — github, attio, linear — do not return account identity with the token, so `identity_from` returns `subject: "pending-<sha256(access_token)[0,32]>"` and a background job resolves the real identity afterwards. Because the provider issues a fresh access token on every consent, that pending subject differs each time, so the upsert never matches the existing row. The enrichment job then resolves the real subject and tries to write `foreign_id: "<provider>-<slug>-<subject>"`, which the original credential already owns; `validates :foreign_id, uniqueness:` raises `ActiveRecord::RecordInvalid`, and the job's existing `rescue` swallows it with a warn log. Nothing re-runs enrichment, so it never self-heals.

What an operator sees:

- The extra credential stays named "Pending … account" forever, with a live token.
- `after_commit :auto_grant_matching_principals` has already wrapped it in a `StaticSecret` and granted it, so multiple wrappers inject `Authorization` for the same hosts and the token that actually reaches an upstream is non-deterministic.
- A scope upgrade never applies: the new scopes land on the credential that gets discarded.
- Reconnecting cannot repair a dead credential — it produces another one instead of reviving it.

## Fix

`Oauth::DuplicateCredentialMerge` folds the duplicate into the credential that already represents the account, and the three enrichment jobs call it once the real subject is known, before their own `update!`. When a credential for that `(oauth_app, subject)` exists, the newest token, refresh token, scopes, expiry, and schedule move onto it (and it is revived if it had gone dead); the duplicate's wrapper secret is destroyed first — `BrokerCredential`'s `before_destroy` guard refuses to orphan the `token_broker` source, and destroying the wrapper drops the grants it auto-created — then the duplicate itself. The surviving wrapper's `update!` re-fires principal reconciliation, so matching principals keep access. When no other credential carries the subject, the service returns `nil` and the job enriches in place exactly as before.

Two consents can enrich out of order (a double-submitted first connect), so the token is only copied when the duplicate's `last_refresh` is at least as new; the duplicate is folded away either way.

This does not add a restriction — it removes a divergence. Providers that do return identity at the callback already upsert one credential per `(app, provider account)` and already keep the first linked user (`credential.created_by ||= current_user`). After this change the pending-identity providers behave the same way.

### DIFF
--- a/services/console/app/jobs/oauth/enrich_attio_credential_identity_job.rb
+++ b/services/console/app/jobs/oauth/enrich_attio_credential_identity_job.rb
@@ -39,15 +39,7 @@ module Oauth
       # derived from the fresh access token), so hand the token to the
       # credential that already represents this workspace instead of colliding
       # with its foreign_id below.
-      duplicate_oid = credential.oid
-      merged = Oauth::DuplicateCredentialMerge.new.call(duplicate: credential, subject: subject)
-      if merged
-        Rails.logger.info do
-          "attio oauth credential identity enrichment merged duplicate: " \
-            "duplicate=#{duplicate_oid} credential=#{merged.oid}"
-        end
-        return
-      end
+      return if DuplicateCredentialMerge.new.call(duplicate: credential, subject: subject)
 
       old_name = credential.name
       credential.update!(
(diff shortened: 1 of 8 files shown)


mslipper:
Superseded by #1398.

PR #1400, opened 2026-08-16
feat: send capability-gated Slack app DMs

vijith-reddy:
## Summary

- add `POST /api/slack/app-dms`, which uses api-rs's server-side `SLACK_BOT_TOKEN` to DM a Slack user as the Centaur app
- add a false-by-default `slack_app_dm` principal capability across Console settings, API payloads, and signed sandbox JWTs
- expose `slack.send_app_dm(user_id, text, idempotency_key)` plus the `slack app-dm` CLI command
- append server-controlled `Triggered by ... via Centaur` attribution and use a deterministic Slack `client_msg_id` for retry safety
- prevent `ctx.post_to_slack` from targeting user or DM conversation IDs, so workflows cannot bypass the dedicated permission
- make the injected Slack credential optional so principals with only the app-DM capability can load and use the Slack tool

## Security model

- the bot token never enters the sandbox or response
- exact-route authorization requires the signed `slack_app_dm` capability
- the recipient must be a Slack `U...` or `W...` user ID; arbitrary channels and DM channel IDs are rejected
- text and idempotency keys are bounded and validated; unknown request fields are rejected
- attribution is derived from a signed Console claim, normalized, escaped, and appended by api-rs
- blocks and link/media unfurls are not caller-controlled

## Dependency

This is intentionally stacked on #1394, which introduces the per-principal sandbox JWT capability model. Retarget this PR to `main` after #1394 merges.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`
- Slack client tests: 56 passed
- focused Console model/JWT/controller tests: 160 passed
- `bundle exec rubocop`: 368 files, no offenses
- `bundle exec brakeman --quiet --no-pager --exit-on-warn --exit-on-error`: no warnings
- full Console suite: 1,517 passed; 3 existing ParadeDB search tests could not run against local PostgreSQL because the `pdb` schema is unavailable

### DIFF
--- a/docs/pages/reference/configuration.mdx
+++ b/docs/pages/reference/configuration.mdx
@@ -117,7 +117,7 @@ created. Changing a default does not rewrite existing principals.
 | `apiRs.metrics.path` | Helm value, default `/metrics`. | Metrics scrape path for annotation-based discovery. |
 | `apiRs.metrics.annotations` | Helm value. | Additional scrape annotations for Prometheus-compatible collectors. |
 | `apiRs.activitySummary.*` | Helm values, default disabled. | Enables API-RS to summarize live session activity into durable `session.activity_summary` events. |
-| `SLACK_BOT_TOKEN` | Explicit `secretKeyRef` from `secretManager.existingSecretName`. | Slack Web API access for api-rs Slack proxy and workflow Slack helpers. |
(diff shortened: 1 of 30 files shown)


vijith-reddy:
Superseded by #1401, which is based directly on main and contains the simplified app-DM implementation without the new capability, Console migration, workflow registration, or idempotency machinery.

PR #1408, opened 2026-08-17
feat(console): add Slack channel autocomplete

gorried:
## Summary

- replace the full Slack channel selector on principal and role detail pages with a server-backed typeahead
- return at most 20 cached catalog matches while excluding channels already assigned to the owner
- support keyboard navigation, direct channel ID entry, accessible status text, and selection-aware save state

## Why

PR #1345 removed synchronous Slack API work from detail-page requests, but the page still rendered the complete cached channel catalog into its HTML. That makes the assignment control increasingly expensive and difficult to use as a workspace grows.

This follow-up keeps the cached catalog provider introduced by #1345, searches that snapshot through a bounded JSON endpoint, and only renders matching options when the operator opens or types into the control.

## User impact

Operators can search channels by name or ID on both principal and role detail pages. The first 20 channels are shown initially, typing searches the full cached catalog, and a valid direct channel ID remains available as a fallback. The save action stays disabled until the control contains a selected or valid channel ID.

## Validation

- `bundle exec rails test` — 1,524 runs, 6,065 assertions, 0 failures, 34 skips
- `bundle exec rubocop` — 365 files, no offenses
- `bundle exec brakeman --quiet --no-pager --exit-on-warn --exit-on-error` — 0 warnings
- `bundle exec rails tailwindcss:build` — successful
- browser verification — page rendered without an error overlay; disabled, valid-ID, and cleared save states behaved correctly
- local 1,000-channel fixture — bounded initial results, full-catalog search, keyboard selection, exclusion, and dropdown behavior verified

### DIFF
--- a/services/console/app/controllers/concerns/console/slack_channel_permission_management.rb
+++ b/services/console/app/controllers/concerns/console/slack_channel_permission_management.rb
@@ -8,10 +8,7 @@ module Console
       @slack_channel_catalog = SlackChannelCatalogProvider.fetch
       @slack_channel_names = @slack_channel_catalog.channels.to_h { |channel| [ channel.id, channel.name ] }
       @slack_channel_permissions = owner.slack_channel_permissions.ordered
-      @slack_channel_options = @slack_channel_catalog.channels.map do |channel|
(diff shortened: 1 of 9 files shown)


gorried:
Screenshot of typeahead picker
image

PR #1415, opened 2026-08-18
feat: instant "@bot status" reply from the control plane

oponder:
## Summary

- a bare `@bot status` (or `health`) mention in an allowlisted Discord channel gets an instant operational report — no sandbox turn, so it keeps working precisely when the agent pipeline is what's broken
- api-rs grows a read-only `GET /api/status` on the protected router: recent + in-flight executions (requester, one-line error), a 24h tally, active session sandboxes, warm-pool state, and a 7-UTC-day run histogram; api-rs owns the session schema, so the SQL lives there and the ingress bot only speaks HTTP
- discordbot renders monospace tables + a histogram block; anything longer than the bare keyword ("status of the deploy?") falls through to a normal agent turn

Example reply:

> **centaur status** · api-rs ✅ ready ✅ · data ✅
> ```
> 24h: 41 ok · 2 FAIL
>
>       THREAD                   WHO         AGE  TOOK
> run   fix the deploy pipeline  oliver       2m     -
> ok    GH PR webapp#1799        diid         5m    1m
> FAIL  flaky e2e tests          jaan        32m   12s
>       └ sandbox spawn timeout after 120s
> ok    Linear PE-8082 r…ew acks oliver       2h    4m
>
> sandboxes: 1 active · warm: 2 ready · warm 24h: 41 claimed, 1 spawn-failed
> ```
> ```
>      LAST 7 DAYS      RUNS FAIL
> Thu  ███████████        31    -
> Fri  ████████████████   44    1
> ...
> 7d: 160 runs · 4 failed (2.5%)
> ```

## Motivation

Operators currently learn about failed or stuck turns by noticing a ❌ reaction in some thread, or by shelling into the box. A status command any teammate can invoke from chat gives that visibility without standing up an authed dashboard — and because it bypasses the sandbox pipeline entirely, it still answers during the incidents it exists to surface.

## Security / hardening

- **Off by default**: the report exposes cross-platform activity (session titles, requester names, error snippets), so the fast-path only answers in channels listed in `DISCORDBOT_STATUS_CHANNEL_IDS` (plumbed via the chart's existing `extraEnv`, no chart change). Unset = feature disabled; other channels fall through to a normal turn.
- `/api/status` sits behind `authorize_api_request` like the rest of the api router; the bot authenticates with its existing `DISCORDBOT_API_KEY` bearer (health probes stay on the open `/healthz`/`readyz`). It's authorized as `SessionsRead` — the report only exposes thread keys, titles and execution state, all of which that capability already reaches, so it's not an escalation. A dedicated `StatusRead` would be reasonable too if you'd rather keep operational views separate; happy to switch.
- Every fetch has a timeout and is best-effort — one dead dependency degrades the report instead of blanking it or wedging the per-thread handler lock; there is no bot-side DB pool.
- Interpolated titles/errors are neutralized (backticks/newlines) so a hostile session title can't break out of the code block; failures get a generic reply with details kept in service logs.
- The report is cached in-process for 10s, which bounds the DB cost of scripted mentions. No index ships with this; if those scans ever get heavy, `session_executions(created_at)` is the one to add.
- Warm-pool `claimed`/`failed` rows are lifetime history (claiming flips status in place), so they're windowed to 24h churn — an unfiltered count reads like a leak.

## Validation

- discordbot: 153 bun tests pass (keyword matcher incl. mention markup; partial-outage collection — api down / report 500 / malformed body; bearer sent on the report fetch only; fence-injection neutralization; 2000-char budget), `tsgo` clean
- api-rs: `cargo check`, `cargo fmt --check`, `cargo test -p centaur-api-server` clean (rust:1 container)
- The pre-endpoint iteration of the formatter/matcher has been running live on our (0xSplits) deployment; the endpoint-shaped version is deploying there this week — draft until that live pass lands


oponder:
Closing this for now, might return with just the status API part. Thanks!

Issue #1359, opened 2026-08-12
Nanocodex pinned to outdated version forces it to be hardcoded to gpt-5.6-sol


marcoatpaladin:
Through sandbox vars, the model is adjustable:

```
sandbox:
  extraEnv:
    CODEX_MODEL: gpt-5.6-luna
```

However, on the current pinned nanocodex version, `bf022728c6d36501bdb1c380c7092743db3d5dae`, this is fully ignored as that version still hardcodes gpt-5.6-sol.

Nanocodex v0.4.0, released August 11, 2026, supports Sol, Terra and Luna as typed models. Its documented API is: `.model(Model::Luna)`.

## Fix

To integrate this, the nanocodex version must be incremented to v0.4.0 and the CODEX_MODEL env variable must be routed into a `.model` call inside the nanocodex builder. The fix only requires a few lines of code.

PR #1433, opened 2026-08-19
fix(chart): default codexNanocodexRolloutPercent to 0 (opt-in Nanocodex rollout)

tiagofneto:
## Problem

The chart defaults `apiRs.codexNanocodexRolloutPercent` to `50`, silently overriding the server's own default of `0` (`args.rs`, `default_value_t = 0`). A fresh production deployment therefore routes half of all Codex sessions to Nanocodex without the operator ever opting in.

As an operator I was genuinely surprised to find `session_harness_rollout_resolved` in the logs and realize our Codex threads had been landing on Nanocodex — nothing in the deploy flow hints an A/B experiment is on by default. The subtle `Codex*` footer in Slack is easy to miss. This is extra sharp while Nanocodex is pinned to an outdated version (#1359).

## Change

- `contrib/chart/values.yaml`: `codexNanocodexRolloutPercent: 50` → `0`, aligning the chart with the server default. The experiment becomes opt-in.
- `docs/pages/reference/configuration.mdx`: updated the default in the reference table.

No behavior change for deployments that set the value explicitly.


mslipper:
Superseded by #1437.

PR #1435, opened 2026-08-19
fix(api-rs): authorize githubbot for all its session key families

oponder:
We noticed our githubbot getting 403s from api-rs after syncing with upstream.

This seems to be since #1374. But I'm surprised there hasn't been an issue or other talk about so far. Has anyone else noticed it too?

## What's happening

#1374 scopes each ingress API key to a single session thread-key prefix — `github:` for githubbot. But githubbot mints four families:

| prefix | used for |
| --- | --- |
| `github:` | issue / PR comment threads |
| `github-issue:` | issue-work runs (issue assigned to the bot) |
| `github-manage:` | owned-PR management (CI fixes, merges, escalation) |
| `github-review:` | review runs |

Only the first is authorized, so anything touching the other three now comes back `caller is not authorized for this session`. All three have been minted that way since #956 (Jul 7), so this affects every githubbot version paired with api-rs at or past #1374.

We found it after picking up #1374 in a fork — a review request on one of our PRs silently went nowhere.

### DIFF
--- a/services/api-rs/crates/centaur-api-server/src/auth.rs
+++ b/services/api-rs/crates/centaur-api-server/src/auth.rs
@@ -65,7 +65,7 @@ pub(crate) struct AuthenticatedCaller {
     class: CallerClass,
     identity: String,
     capabilities: BTreeSet<Capability>,
-    platform_prefix: Option<&'static str>,
+    platform_prefixes: Option<&'static [&'static str]>,
     principal_subject: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


mslipper:
Happy to review this quickly once you mark it as ready for review. Which review request went nowhere?

mslipper:
Went ahead and merged this in. I'll set up the GitHub integration against this repo so we can catch things like this earlier, and dogfood the integration.

oponder:
Ah thanks! I had it as a draft because I was still verifying some final changes on our instance 😅 

I made some changes here after doing a code review. It should be fine though. Tmw morning I'll verify again.

oponder:
Clarifying the "review request went nowhere": that's about a review request on one of our own internal repos failing due to this bug

mslipper:
Sounds good, tThanks for fixing this! Please feel free to open another PR with any additional fixes you make.

PR #1442, opened 2026-08-20
fix(sandbox): make pi-subagents inherit parent pi config and proxy

dennisborshchevski:
Two changes ensure pi-subagents child processes use the same pi SDK and proxy configuration as the parent:

1. **Dockerfile**: symlink `@earendil-works/*` → `@mariozechner/*` so child pi processes resolve the same SDK version. Without this, npm-installed `@earendil-works/pi-coding-agent` bundles a newer `@anthropic-ai/sdk` that sends `eager_input_streaming`, which the LiteLLM proxy rejects with 400.

2. **entrypoint.sh**: write `~/.pi/agent/models.json` from `ANTHROPIC_BASE_URL` so pi routes through the HAIP proxy. pi's built-in model registry hardcodes `api.anthropic.com`, ignoring the env var. The `models.json` override is the supported way to redirect all models.

Prompted by: Niv

### DIFF
--- a/services/sandbox/Dockerfile
+++ b/services/sandbox/Dockerfile
@@ -90,7 +90,7 @@ RUN useradd -m -s /bin/bash -u 1001 agent \
 
 ARG CLAUDE_CODE_VERSION=2.1.154
 ARG CODEX_VERSION=0.144.0
-ARG PI_CODING_AGENT_VERSION=0.67.2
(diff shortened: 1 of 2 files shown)


nivye (approved):

dennisborshchevski:
Opened on wrong repository — closing.

PR #1460, opened 2026-08-22
fix(k8s): report terminated sandbox container reason

### LINKED ISSUE #1454: Report OOMKilled sandbox cause instead of transient status Created after stdout EOF
## Summary

With the `agent-k8s` sandbox backend, an agent container killed by its Kubernetes memory limit can be persisted and shown to the user as:

```text
sandbox stdout closed before terminal output; sandbox no longer accepts io (status Created)
```

The real terminal cause is available shortly afterward in the pod's container state as `OOMKilled`, exit code `137`. The current error obscures the actionable cause and makes this look like a sandbox lifecycle/attach failure.

We observed this three times across independent executions in one long-lived session.

## Environment

- Centaur `60bc16ff884e6becd31c727265dad3fe0f5b0a6c` (`centaur-0.1.117`)
- direct upstream `paradigmxyz/centaur` source and images; no fork or core patch
- `agent-k8s` backend with agent-sandbox controller `v0.4.6`
- Docker Desktop Kubernetes on arm64
- agent pod `restartPolicy: Never`
- sandbox request `1Gi`, limit `4Gi`
- node `MemoryPressure=False` throughout

## Evidence

| Attempt | Container finished | Kubernetes cause | Exit code |
|---|---:|---|---:|
| 1 | 2026-08-20 21:24:38 UTC | `OOMKilled` | 137 |
| 2 | 2026-08-20 22:31:00 UTC | `OOMKilled` | 137 |
| 3 | 2026-08-21 14:19:14 UTC | `OOMKilled` | 137 |

In every case, stdout EOF and the persisted generic failure occurred at the same boundary as the container termination. The workloads were memory-heavy JavaScript monorepo operations (dependency/native-module installation plus concurrent lint, typecheck, test, or Turbo tasks). Shorter turns in the same session completed normally.

Six recent failed agent pods on this deployment had the same combination: `OOMKilled`, exit `137`, and a `4Gi` container limit. This was pod-cgroup exhaustion, not node-wide memory pressure.

## Root cause

This appears to be a status-observation race followed by loss of backend detail:

1. The container hits its memory cgroup limit, stdout closes, and Centaur enters reattach/recovery.
2. At that instant Kubernetes can still report pod phase `Running` while the Ready condition has already fallen false. [`sandbox_status_from_pod`](https://github.com/paradigmxyz/centaur/blob/bd3e1855115ab8c8c1252e138af8b617555a9121/services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs#L618-L643) maps that combination to `SandboxStatus::Created`.
3. [`reattach_session_pipe`](https://github.com/paradigmxyz/centaur/blob/bd3e1855115ab8c8c1252e138af8b617555a9121/services/api-rs/crates/centaur-session-runtime/src/lib.rs#L4480-L4507) treats the high-level status as dead and formats only `status Created`.
4. [`fail_detached_execution`](https://github.com/paradigmxyz/centaur/blob/bd3e1855115ab8c8c1252e138af8b617555a9121/services/api-rs/crates/centaur-session-runtime/src/lib.rs#L4572-L4586) persists that generic detail. The later pod phase/container termination reason is never incorporated.

Current upstream `main` at `bd3e1855115ab8c8c1252e138af8b617555a9121` still contains this path.

## Reproduction

1. Deploy the upstream chart with `agent-k8s` and a bounded `sandbox.resources.limits.memory` value.
2. Start a session turn that exceeds the memory limit.
3. Observe the user/session execution error and the pod in parallel:

```bash
kubectl get pod <sandbox-id> -o json \
  | jq '.status.phase, .status.containerStatuses[0].state.terminated'
```

The session error can report `status Created`; the container state converges to `reason: OOMKilled`, `exitCode: 137`.

## Expected behavior

When stdout closes without terminal output, Centaur should retain the backend termination cause. For example:

```text
sandbox terminated: OOMKilled (exit 137)
```

A bounded re-read after EOF, or inspection of `containerStatuses[*].state.terminated` before collapsing to `SandboxStatus`, could capture `reason`, `exitCode`, and `finishedAt`. The durable failure event/error should include those fields while preserving `Created` as the transient attach-readiness state it represents.

## Downstream mitigation and result

We mitigated the workload failure using only documented upstream chart controls:

- dedicated 48 GiB / 16-core host with a 40 GiB Docker Desktop VM
- sandbox request `8Gi` / 2 CPUs, limit `20Gi` / 12 CPUs
- warm pool size `1`
- running-like sandbox limit `1`
- max lifetime `259200s`
- post-rollout eviction limited to unclaimed ready warm sandboxes so retiring API replicas cannot leave an old resource spec behind

The deployment remained pinned directly to upstream Centaur. The production rollout passed, a real Slack app mention received its reply in the same thread in about eight seconds, the durable execution completed without error, and the claimed agent pod had the `20Gi` / 12 CPU limit with zero restarts. After convergence exactly one agent pod was running, matching the configured host-wide capacity limit, and the Kubernetes node reported no memory pressure.

That removes the observed `4Gi` failure mode for this deployment, but the upstream diagnostic problem remains for any workload that reaches its configured limit.

omerbek:
Summary
- report a terminated sandbox container while the pod is still Running/Pending as an Unknown sandbox status
- include the container name, termination reason, exit code, signal, and message when Kubernetes provides them
- keep ready pods reporting Running and ordinary unready pods reporting Created

Fixes #1454.

Testing
- cargo fmt --all --check
- cargo test -p centaur-sandbox-agent-k8s maps_agent_sandbox_replicas_and_pod_readiness_to_status
- git diff --check

### DIFF
--- a/services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
+++ b/services/api-rs/crates/centaur-sandbox-agent-k8s/src/lib.rs
@@ -636,11 +636,43 @@ fn sandbox_status_from_pod(replicas: i32, pod: Option<&Pod>) -> SandboxStatus {
         .to_ascii_lowercase();
     match phase.as_str() {
         "running" if pod_ready(pod) => SandboxStatus::Running,
-        "running" | "pending" => SandboxStatus::Created,
+        "running" | "pending" => pod_terminated_container_reason(pod)
+            .map(SandboxStatus::Unknown)
+            .unwrap_or(SandboxStatus::Created),
         "succeeded" | "failed" => SandboxStatus::Stopped,
         "unknown" => SandboxStatus::Unknown("unknown".to_owned()),
         other => SandboxStatus::Unknown(other.to_owned()),
     }
+}
+
+fn pod_terminated_container_reason(pod: &Pod) -> Option<String> {
+    let container = pod
+        .status
+        .as_ref()?
+        .container_statuses
+        .as_ref()?
+        .iter()
+        .find(|status| status.name == DEFAULT_CONTAINER_NAME)
+        .or_else(|| {
+            pod.status
+                .as_ref()?
+                .container_statuses
+                .as_ref()?
+                .first()
+        })?;
+    let terminated = container.state.as_ref()?.terminated.as_ref()?;
+    let reason = terminated.reason.as_deref().unwrap_or("terminated");
+    let mut detail = format!(
+        "container {} terminated: {reason} (exit code {})",
+        container.name, terminated.exit_code
+    );
+    if let Some(signal) = terminated.signal {
+        detail.push_str(&format!(", signal {signal}"));
+    }
+    if let Some(message) = terminated.message.as_deref().filter(|message| !message.is_empty()) {
+        detail.push_str(&format!(": {message}"));
+    }
+    Some(detail)
 }
 
 fn pod_ready(pod: &Pod) -> bool {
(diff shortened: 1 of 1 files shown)


mslipper:
Implemented in #1532.

PR #1469, opened 2026-08-24
fix(session-runtime): recover orphaned turns on roll

urielxyz-gabriel:
Towards URI-587

## Context

Any change to centaur values rolls the api-rs deployment, and in-flight agent
turns die with it. The handoff machinery already exists (shutdown drain
releases stdout-owner leases, adoption scans claim them), but three holes
made the handoff fail exactly when the turn mattered:

1. `adopt_orphaned_execution` failed an execution the moment the sandbox no
   longer accepted io, *before* consulting the recorded output. A turn that
   completed during the reader gap (container terminalized while the old pod
   was rolling) was marked `failed` even though its final answer was in the
   pod logs.
2. A transient `read_output_since` error was swallowed into an empty
   transcript. Adoption and the stdout pump then fell straight to a live
   reattach, and a reattach against a terminalized container can only fail
   the execution.
3. The stdout pump gave the recorded-output recovery one attempt before the
   reattach fallback, so the same transient error killed the turn in the
   live-pump path as well.

The incident this fixes: 2026-08-24, a one-line values-only merge (#32) with
three Linear turns live at 36, 20 and 19 minutes; roughly 75 minutes of work
reported `failed: true`.

## Changes

- Adoption claims the lease first, reads the recorded output, and only then
  consults the sandbox status:
  - terminal in the recorded lines: adopt and complete from the logs
    (`session.execution_adopted` with mode `recorded_output`);
  - young sandbox not terminal yet (< 120s): skipped, retried by the
    periodic scan;
  - transient read or status error: claim released, error surfaced, retried
    on the next scan; the execution stays active.
- The stdout pump retries the recorded read (bounded: 3 attempts, 500ms
  delay) before the live-reattach fallback.
- `recover_detached_terminal_output` propagates transient read errors to the
  pump loop instead of mapping them to "no recorded output" (which selected
  the terminalized-container reattach).

## Testing

- `cargo +nightly fmt --all --check` and `cargo +nightly clippy
  --workspace --all-targets -- -D warnings`: clean.
- `cargo test -p centaur-session-runtime`: 119 pass. Five new tests, all
  Postgres-gated (skip without `SESSION_RUNTIME_TEST_DATABASE_URL`, run in
  CI):
  - stopped sandbox + recorded terminal output completes the execution
    instead of failing it;
  - transient read failure on the first adoption scan surfaces an error and
    leaves the execution running; the next scan completes it from the logs;
  - young non-terminal sandbox with no recorded output is skipped;
  - old non-openable sandbox with no recorded terminal output fails with
    `sandbox not openable (status Stopped) with no recorded terminal
    output`;
  - stdout EOF with one transient recorded-read failure completes the turn
    via recorded recovery with no reattach and no failure event.
- Live acceptance check (rolling api-rs during a multi-minute turn and
  confirming the turn finishes) is left to the deployment side: after this
  lands, merge a values-only change while a turn is running and watch the
  thread events for `session.stdout_owner_released` followed by
  `session.execution_adopted` (mode `recorded_output` or `live_attach`) with
  no `session.execution_failed`.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -31,8 +31,8 @@ permissions:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAMESPACE: paradigmxyz/centaur
-  IMAGE_SOURCE: https://github.com/paradigmxyz/centaur
+  IMAGE_NAMESPACE: uriel-xyz/centaur
+  IMAGE_SOURCE: https://github.com/uriel-xyz/centaur
   # Main/tags keep optimized release images. PRs and manual branch publishes
   # use debug builds so staging/dev iteration does not spend minutes optimizing
   # Rust binaries that are immediately replaced by the next test build.
--- a/services/api-rs/crates/centaur-session-runtime/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-runtime/src/lib.rs
@@ -3582,23 +3582,6 @@ impl SessionRuntime {
             return Ok(OrphanAdoption::Failed);
         };
         let id = SandboxId::new(sandbox_id);
-        let status = match self.sandbox_runtime.manager.status(&id).await {
-            Ok(status) => status,
-            Err(SandboxError::NotFound(_)) => SandboxStatus::Gone,
-            // Transient status failures must not fail a possibly live
-            // execution; surface the error and retry on the next startup.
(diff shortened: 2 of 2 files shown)


urielxyz-gabriel:
Closing per Uriel process: we don't open PRs against the upstream repo. The diff remains on this PR's ref until it is deleted.

PR #1470, opened 2026-08-24
fix(session-sqlx): let failed executions re-run

urielxyz-gabriel:
Towards URI-587

## Context

The unique idempotency index on `session_executions` covered every row, so a
failed execution permanently blocked a fresh execution carrying the same
`(thread_key, idempotency_key)`. For triggers with a stable key per commit
(GitHub reviews), a turn killed by a control-plane roll became un-rerunnable:
the insert hit the index and api-rs returned the dead row with
`created = false`. That is what happened to the `uriel-core#211` review on
2026-08-24: the work was lost and could not even be retried.

## Changes

- Migration 0053: drops `session_executions_thread_idempotency_idx` and
  creates `session_executions_active_idempotency_idx`, which covers only
  `queued`, `running` and `completed` rows.
- The `create_execution_with_request` upsert conflict clause is narrowed to
  match the index predicate exactly.
- Failed and cancelled rows fall out of the index, so a retry with the same
  key creates a fresh execution; in-flight and completed work still dedupes.

## Testing

- `cargo +nightly fmt --all --check` and `cargo +nightly clippy -p
  centaur-session-sqlx --all-targets`: clean.
- `cargo test -p centaur-session-sqlx`: 24 pass. New Postgres-gated test
  `failed_execution_does_not_block_rerun_on_same_idempotency_key`:
  - failed row with key K: a fresh `create_execution_with_request` with K
    returns `created = true` and a new execution id;
  - queued row with K: `created = false`, same id;
  - running row with K: `created = false`, same id;
  - completed row with K: `created = false`, same id.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -31,8 +31,8 @@ permissions:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAMESPACE: paradigmxyz/centaur
-  IMAGE_SOURCE: https://github.com/paradigmxyz/centaur
+  IMAGE_NAMESPACE: uriel-xyz/centaur
+  IMAGE_SOURCE: https://github.com/uriel-xyz/centaur
   # Main/tags keep optimized release images. PRs and manual branch publishes
   # use debug builds so staging/dev iteration does not spend minutes optimizing
   # Rust binaries that are immediately replaced by the next test build.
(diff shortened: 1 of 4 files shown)


urielxyz-gabriel:
Closing per Uriel process: we don't open PRs against the upstream repo. The diff remains on this PR's ref until it is deleted.

PR #1481, opened 2026-08-25
fix: allow rollback across newer migrations

PengDeng-Cyber:
Rolling back api-rs to an older image while the database already carries newer migrations fails at startup: the sqlx migrator reports `VersionMissing` for applied migrations the older binary does not embed, so the rollback target crashloops and the only way out is rolling forward.

Set `ignore_missing` on the migrator so an older binary boots cleanly against a newer schema. Checksum validation for the migrations the binary does know is unchanged, and pending migrations still apply in order; the regression test covers booting a binary whose embedded set is a strict prefix of the applied ledger.

We run Centaur at CyberConnect and have carried this as a local patch since 0.1.115.

### DIFF
--- a/services/api-rs/crates/centaur-session-sqlx/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-sqlx/src/lib.rs
@@ -18,11 +18,16 @@ use thiserror::Error;
 use time::{Duration as TimeDuration, OffsetDateTime};
 use uuid::Uuid;
 
-// The API binary embeds these migrations at compile time.
-static MIGRATOR: sqlx::migrate::Migrator = sqlx::migrate!("./migrations");
-
 pub const SESSION_EVENTS_CHANNEL: &str = "centaur_session_events";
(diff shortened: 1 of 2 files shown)


PengDeng-Cyber:
The workflow runs on these four PRs (#1481, #1482, #1483, #1484) have been sitting in `action_required` since they were opened on Aug 25 — the first-time-contributor gate for fork PRs — so CI has never actually reported on them. Could a maintainer approve the runs?

What I validated locally, in case it helps in the meantime:

- `cargo check --workspace --all-targets` is clean on each of the four branches
- #1484 was rebased onto main today (it conflicted with the reworked `stop()` cleanup); the merge keeps the new sandbox files ConfigMap deletion while preserving this PR's ordering, and `cargo test -p centaur-sandbox-agent-k8s` passes (65 tests) including the new egress test
- #1481, #1482 and #1483 still merge cleanly onto current main
- each PR is a single focused commit

Happy to rebase, split, or adjust any of them if that makes review easier.

PR #1484, opened 2026-08-25
fix: preserve sandbox egress during shutdown

PengDeng-Cyber:
Stopping a sandbox removes its per-sandbox proxy egress path while the pod is still terminating. Anything the sandbox does during TERM/preStop — flushing telemetry, closing upstream connections through iron-proxy, tool cleanup — races that removal and fails with a dead network, which shows up as noisy shutdown errors and half-finished cleanup.

Quiesce instead: keep the egress path until the pod is actually gone (`wait_until_pod_gone` with a bounded timeout), then tear it down. The agent-k8s test drives a stop against a fake kube API and asserts the proxy egress survives until the pod deletion is observed.

We run Centaur at CyberConnect and have carried this as a local patch since 0.1.115.

### DIFF
--- a/services/api-rs/Cargo.lock
+++ b/services/api-rs/Cargo.lock
@@ -973,6 +973,7 @@ dependencies = [
  "centaur-iron-proxy",
  "centaur-sandbox-core",
  "clap",
+ "http 1.4.2",
  "jiff",
  "k8s-openapi",
  "kube",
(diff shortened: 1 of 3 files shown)


PengDeng-Cyber:
Rebased onto main: `stop()` now also deletes the sandbox files ConfigMap, so the merge keeps that cleanup while preserving this PR's ordering — quiesce first, then the sandbox delete, and only afterwards the proxy egress teardown. `cargo test -p centaur-sandbox-agent-k8s` passes (65 tests), including `stop_keeps_proxy_egress_until_the_sandbox_pod_is_gone`.

Issue #1385, opened 2026-08-13
Egress allowlist is unreachable on Kubernetes: security docs point at a config file managed-mode proxies never read


tobernguyen:
`docs/pages/security.mdx` ("Egress policy") tells operators to lock down egress by editing `services/iron-proxy/iron-proxy.yaml`, replacing `domains: ["*"]` with an explicit hostname list, after which iron-proxy rejects everything else with a 403. On Kubernetes with iron-control that file is never read, and the console has no model that can deliver an `allowlist` transform. A deployment that follows the documented guidance still runs fully open, with no supported way to change it. Filing as a docs/security-model mismatch rather than a feature request, because the docs read as though egress lockdown is available when it is not.

Why it does not apply on Kubernetes (checked at `centaur-0.1.115`, unchanged at `0.1.117`):

1. **Managed mode renders no local config.** `centaur-sandbox-agent-k8s/src/iron_proxy.rs`: "Managed-mode proxies carry no rendered config", and `resolve_iron_proxy` hard-errors with "iron-proxy requires iron-control to be configured". The remaining settings arrive as `IRON_*` env vars. So `iron-proxy.yaml` is a compose/local-only surface.
2. **The console cannot emit an `allowlist` transform.** `PrincipalSyncConfigSnapshot.proxy_transforms_for` builds exactly five transform kinds (`gcp_auth`, `gcp_id_token`, `aws_auth`, `hmac`, `oauth_token`), with no generic passthrough. Confirmed live: a running proxy logs `"transform pipeline","transforms":"secrets"`.
3. **No model, endpoint, or UI.** `services/console/config/routes.rb` has no route matching allowlist, egress, rule, or domain.
4. **The sync controller says so.** `api/v1/proxy_sync_controller.rb` notes that the top-level `rules`, `mcp`, and `ingest_token` fields "are intentionally omitted: centaur-console has no models for them yet".

Per-secret `rules` are modeled, so credential injection is correctly scoped; an unrelated host gets no credentials. This issue is only about egress reachability.

Repro: Kubernetes, managed mode, sandbox NetworkPolicies in force. From a sandbox, `curl -o /dev/null -w '%{http_code}' https://example.com` returns 200 (measured at `0.1.106`; the relevant code paths are unchanged through `0.1.117`). Per the docs it should be 403 after lockdown. `ironProxy.upstreamDenyCidrs` exists but is CIDR-only and deny-only, so it cannot express a hostname allowlist.

[...]

Issue #1176, opened 2026-07-24
Support consuming external MCP servers from sandboxes (MCP client), with proxy-injected credentials


shrivats1:
## Context

We run Centaur as an overlay (self-hosted, env-secrets mode, iron-proxy credential injection). We're adding a capability where the agent opens GitHub pull requests, and evaluated pointing the sandbox at GitHub's official MCP server (`github/github-mcp-server`) instead of hand-rolling REST tools.

The existing MCP work we found (#883 scope tool visibility to a principal's granted secrets, #932 OAuth creds + `user-mcp` role for console-user principals, #841 / #718 MCP tool-host sandboxes, #1059 hosted MCP OAuth redirects) all concerns Centaur **as an MCP server** (exposing Centaur's own tools to external MCP clients). We're asking about the reverse: Centaur **as an MCP client**, where a sandbox consumes a third-party MCP server.

## The gap

As far as we can tell there's no supported path to:

1. **Deliver MCP client config** (`.mcp.json` / server list) into a sandbox through the control plane. The overlay contract mounts `tools/`, `.agents/skills/`, `workflows/` only; the console's `proxy_sync` payload omits a top-level `mcp` field ("no models for them yet"). So today you'd hand-bake config into the sandbox image, off the paved road.
2. **Scope credentials to an external MCP server** the same way `[tool.centaur].secrets` binds a tool's secret to a host. For a remote MCP server the sandbox needs (a) the server host added to the egress allowlist and (b) the token injected into the client's `Authorization` header. The placeholder-substitution model works for a static bearer, but there's no declarative way to register "this MCP server, this host, this injected credential, granted to this principal."

## Why it matters

The whole value of consuming a first-party MCP server (GitHub's, etc.) is not re-implementing its surface as bespoke REST tools. Without a client story, every integration collapses back to per-endpoint tools, which for narrow cases is fine but doesn't scale as the surface grows.

## Questions

1. Is sandbox-side MCP client consumption on the roadmap, or intended to stay overlay-baked?
2. If a team wants it today, is there a supported way to (a) deliver `.mcp.json` per sandbox and (b) register an external MCP server's host + injected credential through the console, analogous to the `http` / `gcp_auth` secret types, or is that exactly the missing model?
3. Would the per-tool `tool-<slug>` role grant (#883) extend to gating access to a consumed external MCP server per principal?

Happy to contribute if there's a shape you'd accept.

tobernguyen:
[...]

PR #1524, opened 2026-08-28
docs: propose overlay-native external MCP servers

gskril:
## Summary

- propose a typed `[tool.centaur.mcp]` declaration for remote, tool-only Streamable HTTP MCP servers in overlay tool packages
- keep CLI and external MCP capabilities independent: MCP-only packages do not synthesize executable tools, and external MCP uses a separate role by default
- compile one normalized catalog into native harness configuration and principal-scoped managed proxy routes, method/tool policy, route-only egress, and optional gateway-only credentials
- add source-attributed catalog generations and a strict proxy-plus-harness assignment barrier for external MCP
- preserve existing CLI, skill, persona, generic-secret, operator-overlay, public `/mcp`, and non-MCP sandbox behavior

### Tool-only version 1

Version 1 permits the MCP lifecycle, ping, cancellation and progress messages, `tools/list`, explicitly authorized `tools/call`, and the Streamable HTTP session operations required by the pinned protocol version.

Resources, prompts, completion, sampling, roots, elicitation, logging control, unknown methods, unrestricted server-to-client requests, general MCP OAuth, and local stdio are deferred until Centaur can enforce them explicitly.

### Gateway-only authority

A credentialed MCP declaration references an existing same-package secret declaration by name. That reference reuses Centaur’s existing secret-source and grant patterns, but the credential is bound only to the generated MCP gateway route rather than emitted as a generic host-scoped proxy secret.

The generated route owns both upstream egress and credential injection, so direct requests to the upstream URL do not inherit the MCP capability or credential.

### Independent authorization

External MCP defaults to a role such as `mcp-<package>-<server>` rather than silently expanding an existing `tool-<slug>` grant.

Interactive sessions require the conversation principal and, when present, the requester principal to hold the MCP role. Credentialless servers are still private by default; `public = true` is explicit and subject to deployment policy.

### Additive discovery and fail-closed shadowing

The discovery model represents CLI and external MCP capabilities independently, so an MCP-only package contributes no CLI entry and does not appear in Centaur’s existing public `/mcp` catalog.

External MCP declarations get their own fail-closed overlay semantics: an invalid later MCP override suppresses the earlier MCP capability without changing existing CLI, persona, skill, secret, or other overlay behavior.

### Managed generations and warm pools

api-rs remains the TOML parser and publishes a normalized, source-attributed catalog generation to Console and iron-control.

Catalog, grant, proxy, and final harness hashes are tracked separately. For assignments containing external MCP, Centaur verifies both proxy and harness generations before delivering agent input. Existing assignment behavior remains unchanged when no external MCP server is assigned.

## Scope

This is a draft design PR only. It does not make external MCP available or change runtime behavior.

The RFC recommends separate implementation PRs for:

1. typed discovery and MCP-specific overlay tombstones;
2. catalog publication, roles, and gateway-only credential bindings;
3. managed proxy routes, route-only egress, and the tool-only protocol profile;
4. harness rendering and the strict external-MCP assignment barrier; and
5. local stdio support in a later phase.

## Related work

- Related to #1176
- Related to #1385
- Builds on the Streamable HTTP header finding in #205
- Follows the role-oriented direction discussed in #883 without depending on it
- Complements the opposite MCP direction implemented in #841

## Validation described by the RFC

- compatibility tests for packages and secrets without the new MCP metadata
- parser, MCP-only package, shadowing, tombstone, catalog-generation, and principal/requester authorization tests
- managed-mode tests for bidirectional method policy, tool allowlists, sessions, SSE resumability, direct-upstream denial, redirects, DNS changes, and credential isolation
- cold-create, warm-pool, refresh, revocation, and harness-reload generation tests
- end-to-end tests using one anonymous and one controlled credentialed MCP server in the local Kubernetes stack

---

Written by GPT 5.6 Sol in Codex

### DIFF
--- a/services/api-rs/rfcs/0006-overlay-native-external-mcp.md
+++ b/services/api-rs/rfcs/0006-overlay-native-external-mcp.md
@@ -8,150 +8,94 @@ Target: `services/api-rs`, `services/sandbox`, `services/console`,
 ## Summary
 
 Let an overlay register a remote Model Context Protocol server by extending the
-existing per-tool `pyproject.toml` contract. Centaur should discover one typed,
-normalized declaration, publish it as versioned control-plane state, render the
-authorized server into each supported harness, and compile the same declaration
-into principal-scoped iron-proxy gateway, egress, credential, protocol, and tool
-policy.
+existing per-tool `pyproject.toml` contract. Centaur should discover one typed
+declaration, publish it as versioned control-plane state, render the authorized
+server into each supported harness, and compile the same declaration into
+principal-scoped iron-proxy route, egress, credential, method, and tool policy.
 
-Version 1 is deliberately a **remote, tool-only Streamable HTTP profile**. It
-allows the MCP lifecycle plus `tools/list` and explicitly authorized
-`tools/call` requests. Resources, prompts, completion, sampling, roots,
-elicitation, logging control, unknown methods, and local stdio servers remain
-out of scope until Centaur can enforce them explicitly in both protocol
-directions.
-
-An operator should not need to:
-
-- hand-write a different MCP config for every harness;
-- copy a token into a sandbox;
-- implement a bespoke Python JSON-RPC client for every MCP server; or
-- add a second root-level overlay manifest that duplicates tool, secret, role,
-  and source metadata Centaur already discovers.
+Version 1 is a remote, tool-only Streamable HTTP profile. It supports the MCP
+lifecycle, `tools/list`, and explicitly authorized `tools/call` requests.
+Resources, prompts, sampling, roots, elicitation, unrestricted extension
+methods, general MCP OAuth, and local stdio servers are deferred until Centaur
+can enforce them explicitly.
 
 The proposed unit remains a tool package directory, but its capabilities become
 explicit. A directory may provide a CLI, one remote MCP server, or both. An
 MCP-only directory does not synthesize a CLI or appear in Centaur's existing
-public `/mcp` tool catalog. Skills remain in `.agents/skills/`, where the overlay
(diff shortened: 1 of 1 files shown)


shrivats1:
+1 on this direction. We filed #1176 and this is the shape we were hoping for. The typed declaration plus a route-scoped credential is a lot cleaner than the CODEX_CONFIG_OVERLAY workaround.

One ask from a Claude Code deployment: could Claude Code rendering land alongside Codex instead of after it? The RFC puts it in phase 4 behind Codex, but Codex support doesn't help us since we run the claude-code harness in production, so we'd stay blocked through phases 1 to 3.

A few Claude Code specifics for the renderer, in case they're useful:

- settings.json can only enable or permission MCP servers (enableAllProjectMcpServers, enabledMcpjsonServers). It cannot define one. So CLAUDE_SETTINGS_OVERLAY on its own can't register a server, which matches the RFC's point that the overlays can't create proxy authority.
- Definitions have to live in .mcp.json (project root) or ~/.claude.json (user scope), key mcpServers. The sandbox entrypoint writes neither today, so a native Claude Code renderer would need to emit one of them.
- .mcp.json supports ${VAR} expansion in headers. For the gateway-only credential model the rendered file can carry a placeholder Authorization header and let the proxy inject, so no token sits in the sandbox.

The gateway-only credential binding is the part we care about most. We currently reach internal services with host-scoped proxy secrets, and the host-not-route gap is a real residual for us. Binding the credential to the generated MCP route instead closes it.

Happy to help validate the Claude Code harness path against a self-hosted, env-secrets stack. Our use case is the GitHub MCP server for opening PRs, which is tool-only and fits the v1 scope.

PR #1526, opened 2026-08-29
fix(console): accept current Granola MCP meeting tags

lucacseth:
## What changed

- accept the additional attributes now emitted on Granola MCP `<meeting>` tags
- treat an empty meeting window as a successful empty sync instead of calling `get_meetings` with an invalid empty ID array
- cover both current tag shape and empty-window behavior

## Why

Granola MCP currently emits attributes such as `captured_by_me`, `listed_as_participant`, and `is_workspace_visible` after `date`. The exact-tag regex in 0.1.115 consequently parses 374 returned meetings as zero, then calls `get_meetings` with `meeting_ids: []`, which Granola rejects.

## Verification

```text
4 runs, 30 assertions, 0 failures, 0 errors, 0 skips
```

Ran `test/services/granola/sync_credential_test.rb` in the pinned Console image against an isolated Postgres 16 database.

### DIFF
--- a/services/console/app/services/granola/sync_credential.rb
+++ b/services/console/app/services/granola/sync_credential.rb
@@ -17,7 +17,7 @@ module Granola
     DEFAULT_MAX_NOTES = 50
     WATERMARK_OVERLAP_SECONDS = 5 * 60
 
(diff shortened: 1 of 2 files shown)


lucacseth:
Closing this: current upstream main already replaced the regex with Nokogiri parsing, guards empty results, batches Granola detail calls, and includes a cursor-based Slack private sync with 120s ingestion timeout/rate-limit resumption. The deployed 0.1.115 pin predates those fixes; I will validate an upstream pin bump instead of backporting a narrower fix.

PR #1529, opened 2026-08-30
fix(k8s): own iron-proxy resources and reap orphans

### LINKED ISSUE #1466: iron-proxy pods can orphan with no ownerReferences and no reconcile can reach them
## Summary

An iron-proxy pod can be left running with **no `ownerReferences` and no corresponding `Sandbox` CR**. Once in that state nothing in api-rs can ever remove it: every deletion path is keyed by a `SandboxId` the caller already holds, and the reaper sweeps *observed sandboxes*, so an orphan whose Sandbox is gone is invisible to it. It runs until someone deletes it by hand.

This matters more than a stray pod because an iron-proxy holds the real credentials for its sandbox — it is the component that substitutes them into outbound requests. An orphan is a credential-bearing container with no owner, no reaper that can see it, and no expiry.

## Environment

- Centaur images `ghcr.io/paradigmxyz/centaur/centaur-api-rs:main-sha-46bebdf`, chart at `46bebdfebe94a0025b46c9faebb55c6c77174f85`
- direct upstream `paradigmxyz/centaur` images, no fork or patch
- `agent-sandbox-k8s` backend, agent-sandbox controller `v0.4.6`
- RKE2 `v1.35.7+rke2r1`, single node, arm64, Ubuntu 24.04.4
- `sandboxWarmPoolSize: 1`, `sandboxRunningLimit: 6`, `sandboxMaxLifetimeSecs: 259200`

## Observed

`asbx-1787506162504-8-proxy-1787506162526` ran for 32 minutes after its Sandbox was gone:

```
$ kubectl get pod -n centaur-system asbx-1787506162504-8-proxy-1787506162526 \
    -o jsonpath='{.metadata.ownerReferences}'
                       # empty

$ kubectl get sandboxes.agents.x-k8s.io -n centaur-system
asbx-1787506967159-5    ...   # -8 absent
```

Labels were intact (`centaur.ai/iron-proxy=true`, `centaur.ai/sandbox-id=asbx-1787506162504-8`, `centaur.ai/managed-by=api-rs`). The pod was created 17:29:22. Its Sandbox never reached Running — the surrounding window shows sibling creations failing the same way:

```
sandbox_create_failed: sandbox is not ready: sandbox asbx-... did not become
running before timeout; latest status: Created   startup_duration_ms: 92345
```

api-rs was then rolled at ~17:36, which rotated away the logs covering this pod's own failure, so I cannot say which unwind step failed for it specifically. Two sibling failures in the same window *were* cleaned up, so the unwind normally works.

The contrast is the useful part. Deleting six healthy sandboxes later the same day cascaded correctly — all six proxies went with them, zero orphans:

```
$ kubectl delete sandboxes.agents.x-k8s.io -n centaur-system --all
$ # orphan check: proxy sandbox-ids with no live Sandbox
$ comm -23 <(proxy sandbox-ids) <(live sandbox names)
                       # empty
```

So adoption and cascade are fine on the normal path. The gap is what happens when they are not.

## Mechanism

`AgentSandboxBackend::create` (`centaur-sandbox-agent-k8s/src/lib.rs:340-380`) creates proxy resources *before* the Sandbox CR, because the egress policies must exist before the pod starts, then binds them afterwards. It does unwind on each failure mode — but every unwind and the adoption itself are best-effort:

```rust
if let Err(err) = self.create_iron_proxy_resources(&id, ...).await {
    let _ = self.delete_iron_proxy_resources(&id).await;   // discarded
    return Err(err);
}
...
    Err(err) => {
        let _ = self.delete_iron_proxy_resources(&id).await; // discarded
        return Err(map_kube_error("create sandbox", err));
    }
...
// "Failure leaves them cleanable by stop() only."
if let Err(error) = self.adopt_iron_proxy_resources(&id, &created).await {
    tracing::warn!(...);                                    // warn only
}
if let Err(err) = self.wait_until_running(&id).await {
    let _ = self.stop(&id).await;                           // discarded
    return Err(err);
}
```

So there are two ways to reach the orphaned state: an unwind that fails (the error is dropped), or api-rs dying between the failure and the unwind completing. `adopt_iron_proxy_resources` also returns `Ok(())` early when `sandbox_owner_reference(sandbox)` is `None`, which leaves the pod unowned without an error at all.

What makes it permanent rather than transient is that there is no reconcile. `delete_iron_proxy_resources` is only ever called with a `SandboxId` the caller already knows, and per `centaur-sandbox-manager/src/reaper.rs`:

> The reaper sweeps the backend's **observed sandboxes** and stops any that exceed the configured max lifetime, releasing the sandbox, its proxy resources, and its node pod slots.

An orphaned proxy has no Sandbox to observe, so no sweep reaches it and `sandboxMaxLifetimeSecs` never applies.

## Ask

A reconcile keyed off the labels the pods already carry would close this regardless of which unwind failed: on startup and periodically, list pods with `centaur.ai/iron-proxy=true` and delete any whose `centaur.ai/sandbox-id` has no live Sandbox.

```
comm -23 \
  <(kubectl get pods -n centaur-system -l centaur.ai/iron-proxy=true \
      -o jsonpath='{range .items[*]}{.metadata.labels.centaur\.ai/sandbox-id}{"\n"}{end}' | sort -u) \
  <(kubectl get sandboxes.agents.x-k8s.io -n centaur-system -o name | sed 's|.*/||' | sort -u)
```

Two smaller things that would help on their own:

1. The `let _ = self.delete_iron_proxy_resources(...)` unwinds discard their errors. Even logging a failed unwind would make this diagnosable rather than requiring a pod-by-pod ownerReferences check.
2. `adopt_iron_proxy_resources` returning `Ok(())` on a `None` owner reference silently accepts an unowned pod. Warning there would surface the window.

Happy to open a PR for the reconcile if that shape is agreeable.

Related: #1432 (in-process state lost across an api-rs restart — different symptom, and it notes proxies are deliberately retained during idle-pause, which is not this case), #1454 (sandboxes stuck at status `Created`, the failure mode that preceded this orphan), #538.

### LINKED ISSUE #1528: iron-proxy Services orphan the same way pods do, and #1466's orphan check cannot see them
## Summary

#1466 describes an iron-proxy pod reaching an unowned state with no live `Sandbox` and no reconcile that can reach it. The proxy `Service` reaches that same state through the same gap, and it survives there far longer than the pod does. Six leaked Services on one cluster carried no `ownerReferences`, had no endpoints, and had no `Sandbox` CR. The oldest stood for 6d16h.

The orphan check I suggested in #1466 lists pods only, so it reported clean for the whole time those six stood.

## Environment

- `agent-sandbox-k8s` backend, namespace `centaur-system`
- single node, arm64, RKE2
- upstream `paradigmxyz/centaur` images and chart, no fork or patch

## Observed

Six proxy Services with no `Sandbox` CR, no endpoints, and no owner references:

| Service | Age |
| -- | -- |
| `asbx-1787457637809-1-proxy` | 6d16h |
| `asbx-1787506162504-8-proxy` | 6d3h |
| `asbx-1787664727363-1-proxy` | 4d7h |
| `asbx-1787817946457-9-proxy` | 2d12h |
| `asbx-1787849646952-1-proxy` | 2d4h |
| `asbx-1788028770377-15-proxy` | 140m |

`asbx-1787506162504-8-proxy` belongs to the same sandbox as the pod in #1466's finding, `asbx-1787506162504-8-proxy-1787506162526`. I deleted that pod by hand on 2026-08-23. Its Service outlived it by six days.

## Mechanism

A healthy proxy Service carries `ownerReferences: [Sandbox]`, so cascade deletion collects it along with its CR. None of the six carried one, which puts them past cascade and past any reconcile keyed on an observed sandbox. Nothing in the cluster would have collected them.

`create` builds the proxy resources before the Sandbox CR exists and binds them afterwards, so the Service passes through the same unowned window the pod does:

- `adopt_iron_proxy_resources` returns on the first failed patch. A pod patch error leaves the Service and both network policies unbound.
- It also returns `Ok(())` when `sandbox_owner_reference` is `None`, which accepts unowned resources without an error.
- `delete_iron_proxy_resources` discards its per-class delete errors, and the failure-path unwinds in `create` and `resume` discard theirs.

Adoption leaves more than the pod unowned, and the same is true of the network policies. The sandbox egress policy carries only the managed-by and sandbox-id labels, so a sweep selecting on `centaur.ai/iron-proxy` misses it.

## Impact

Bounded, which is why I am filing this rather than treating it as urgent. A Service consumes no pod budget, and the pod budget is what binds the node. The cost is a stale ClusterIP per leak, a name that cannot be reused, and an operator check that reports clean while the leak continues.

That last point is the reason to file. #1466 rests on a check that lists one resource class and cannot see this one.

## Detection

Widened to Services:

```
comm -13 \
  <(kubectl get sandboxes.agents.x-k8s.io -n centaur-system -o name \
      | sed 's|.*/||' | sort -u) \
  <(kubectl get svc -n centaur-system --no-headers \
      | awk '$1 ~ /^asbx-.*-proxy$/ {print $1}' | sed 's/-proxy$//' | sort -u)
```

Empty after I deleted the six above by hand.

## Ask

Bind every per-sandbox resource class during adoption instead of stopping at the first failure, surface the unwind errors, and widen the reconcile from pods to every labeled proxy class with a per-class count. PR follows.

0xAlcibiades:
Closes #1528
Closes #1466

## Context

#1466 reported the pod half of this: a failed create, resume, or unwind can leave an iron-proxy resource behind with no owner reference and no live `Sandbox` CR, and every cleanup path is keyed on an observed sandbox, so the orphan stands until someone deletes it by hand.

#1528 records the same state on the proxy `Service`. Six leaked Services stood for up to 6d16h with no endpoints and no owner references, while the pod-only orphan check reported clean the whole time. The Service is created before the Sandbox CR exists, so it reaches the unowned state through the same adoption and unwind gaps the pod does, and nothing observes it afterwards.

## Changes

- `adopt_iron_proxy_resources` attempts every class (pods, service, network policies) and returns the failures aggregated, so a pod patch error can no longer leave the service and policies unbound. A Sandbox CR with no name or uid now warns instead of accepting unowned resources in silence, and `resume` warns when the CR is missing and it cannot bind what it recreated.
- `delete_iron_proxy_resources` aggregates its per-class delete errors instead of discarding them, and the failure-path unwinds in `create` and `resume` log a failed unwind rather than dropping it.
- New label-keyed orphan sweep, run by the sandbox reaper at startup and on each sweep interval. It lists the proxy pods, services, and network policies and deletes any whose `centaur.ai/sandbox-id` has no live `Sandbox`, past a 10-minute grace that keeps an in-flight create or resume from racing the sweep. It reports the count deleted per class.

The reaper is now enabled whenever a sweep interval is configured rather than only when a max lifetime is set, since the orphan sweep is the backstop. The sweep is a no-op on backends that manage no proxy resources.

The sweep covers the sandbox egress policy as well. That policy carries only the managed-by and sandbox-id labels rather than the `centaur.ai/iron-proxy` label the pods and service carry, so a selector on the proxy label alone would miss it.

## Testing

- New unit tests for the orphan classification: live-sandbox skip, grace-window skip, missing-label skip, missing-timestamp skip, and the delete case. Reaper test for the enabled-without-max-lifetime config.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace`, all on stable.

I have not run the Kind sandbox suite (`centaur-sandbox-e2e`) against this.


services/api-rs/crates/centaur-sandbox-agent-k8s/src/iron_proxy.rs
```diff
@@ -30,6 +30,13 @@ use crate::{
 
 const IRON_PROXY_LABEL: &str = "centaur.ai/iron-proxy";
 const IRON_CONTROL_PROXY_ID_ANNOTATION: &str = "centaur.ai/iron-control-proxy-id";
+/// How long a labeled iron-proxy resource must survive with no live
+/// Sandbox before the orphan sweep may delete it. Creates and resumes build
+/// the resources before their Sandbox CR exists, so the grace keeps the
+/// sweep from racing an in-flight one.
+fn orphan_sweep_grace() -> jiff::Span {
+    jiff::Span::new().seconds(600)
```
0xAlcibiades: Arguably this would be exposed as config.

PR #1530, opened 2026-08-30
fix(session-runtime): recover orphaned turns on roll

### LINKED ISSUE #1513: api-rs cannot roll without killing every in-flight turn: stdout lease has no live handoff and the stream is not replayable
## What happens

Any redeploy of api-rs (a values change, an image bump) kills every running agent turn at once with `sandbox stdout closed before terminal output; stdout reattach attempts exhausted`. Observed on a single-node production deployment at `46bebdf`: a one-line config merge destroyed three live turns at 36, 20, and 19 minutes elapsed; two more earlier the same day at 73 and 34 minutes; and a review turn killed mid-flight could not be retried at all afterward, because the idempotency key is stable per commit and re-requests return the dead execution with `created=false`.

## Why it happens

The architecture already carries the right idea and the default deployment shape defeats it. `session_executions` has `stdout_owner_id` and `stdout_owner_lease_expires_at`, renewing on a ~60s TTL — so a replacement instance is clearly expected to claim the lease and reattach to a still-running sandbox. It cannot, because:

```
replicas: 1   strategy: RollingUpdate   maxSurge 25%   maxUnavailable 25%   grace 35s
```

With one replica there is no live instance to hand the lease to. The old pod's exec streams close on SIGTERM; the sandbox keeps producing output with no reader; a Kubernetes exec stream does not buffer for an absent reader, so those bytes are gone. The new pod's reattach finds a stream missing its terminal output, retries, and exhausts. The sandbox itself is healthy throughout — only the reader and the un-replayable bytes are lost.

## Open questions that decide the fix

1. Is api-rs safe multi-instance (workflow runtime, sandbox reaper, warm-pool management), or do those assume a singleton? The lease schema implies HA was designed for; the default chart shape never exercises it.
2. Can sandbox output survive a reader gap at all? `session_events` already persists a complete turn transcript, so the bytes exist somewhere durable — if that path is independent of the stdout pipe, reattach could reconcile from it instead of requiring an unbroken stream.

## Proposed direction

Any of: HA replicas with verified lease handoff; durable/replayable sandbox output (spool in harness-server, or reconcile reattach from the persisted event stream); or drain-then-handoff on SIGTERM. Failing all three, an interim mitigation: let a turn killed by a roll be retried (terminal failure should release the idempotency key) — losing the work is bad, losing it irrecoverably is worse.

Related: #1432 (api-rs restart also loses in-process idle timers — same restart, different loss).

0xAlcibiades:
Closes #1513

## Context

Any change to values rolls the api-rs deployment, and in-flight agent turns die
with it. The handoff machinery already exists — the shutdown drain releases
stdout-owner leases and adoption scans claim them — but three holes made the
handoff fail exactly when the turn mattered:

1. `adopt_orphaned_execution` failed an execution the moment the sandbox stopped
   accepting io, *before* consulting the recorded output. A turn that completed
   during the reader gap, where the container terminalized while the old pod was
   rolling, was marked `failed` with its final answer sitting in the pod logs.
2. A transient `read_output_since` error was swallowed into an empty transcript.
   Adoption and the stdout pump then fell straight through to a live reattach,
   and a reattach against a terminalized container can only fail the execution.
3. The stdout pump gave the recorded-output recovery one attempt before the
   reattach fallback, so the same transient error killed the turn in the live
   path too.

Observed as roughly 75 minutes of agent work reported `failed: true` after a
one-line values-only merge, with three turns live at 36, 20 and 19 minutes.

## Changes

Adoption claims the lease first, reads the recorded output, and only then
consults the sandbox status:

- a terminal in the recorded lines adopts and completes from the logs
  (`session.execution_adopted` with mode `recorded_output`);
- a young sandbox that has not terminalized yet (< 120s) is skipped and retried
  by the periodic scan;
- a transient read or status error releases the claim and surfaces the error, so
  the execution stays active and the next scan retries it.

The stdout pump retries the recorded read (3 attempts, 500ms apart) before the
live-reattach fallback, and `recover_detached_terminal_output` propagates
transient read errors to the pump loop instead of mapping them to "no recorded
output", which is what selected the terminalized-container reattach.

## Testing

`cargo test -p centaur-session-runtime`: 119 pass, including five new tests —

- a stopped sandbox with recorded terminal output completes the execution rather
  than failing it;
- a transient read failure on the first adoption scan surfaces an error and
  leaves the execution running; the next scan completes it from the logs;
- a young non-terminal sandbox with no recorded output is skipped;
- an old non-openable sandbox with no recorded terminal output fails with
  `sandbox not openable (status Stopped) with no recorded terminal output`;
- stdout EOF with one transient recorded-read failure completes the turn via
  recorded recovery, with no reattach and no failure event.

`cargo fmt --all --check` and `cargo clippy -p centaur-session-runtime
--all-targets` are clean.

### DIFF
--- a/services/api-rs/crates/centaur-session-runtime/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-runtime/src/lib.rs
@@ -3585,23 +3585,6 @@ impl SessionRuntime {
             return Ok(OrphanAdoption::Failed);
         };
         let id = SandboxId::new(sandbox_id);
-        let status = match self.sandbox_runtime.manager.status(&id).await {
-            Ok(status) => status,
-            Err(SandboxError::NotFound(_)) => SandboxStatus::Gone,
-            // Transient status failures must not fail a possibly live
-            // execution; surface the error and retry on the next startup.
-            Err(error) => return Err(SessionRuntimeError::Sandbox(error)),
-        };
-        if !status.can_open_io() {
-            self.fail_orphaned_execution(
-                thread_key,
-                execution_id,
-                sandbox_id,
-                &format!("sandbox no longer accepts io (status {status:?})"),
-            )
-            .await;
-            return Ok(OrphanAdoption::Failed);
-        }
         if !self.claim_expired_stdout_owner(execution_id).await? {
             // Deferrals repeat on every periodic scan while another control
             // plane pumps the execution; only the first observation is worth
@@ -3660,6 +3643,10 @@ impl SessionRuntime {
             Ok(lines) => lines,
             Err(SandboxError::Unsupported { .. }) => Vec::new(),
             Err(error) => {
+                // Transient read failure. The recorded output is the only
+                // source of truth once the container terminalized, so
+                // swallowing it here would fail live turns; release the
+                // claim and let the next scan retry the read.
                 warn!(
                     component = COMPONENT_SESSION_RUNTIME,
                     event = "execution_adoption_log_read_failed",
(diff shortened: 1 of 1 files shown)


0xAlcibiades:
Heads up on merge order: #1532 touches the same `adopt_orphaned_execution` path this PR restructures. Both branch from `main` independently, so whichever merges second needs a small rebase. I have the combined resolution working locally and will push whichever one ends up second.

PR #1531, opened 2026-08-30
fix(session-sqlx): let failed executions re-run

### LINKED ISSUE #1507: failed executions permanently hold their idempotency key; identical re-requests return the dead execution
## What happens

`session_execute` dedupes on `idempotency_key`, and the dedupe does not distinguish terminal failures from in-flight or completed work. A re-request whose key matches an execution that already ended in terminal `failed` returns the old failed execution (`created=false` in the `session_execute_enqueued` log) and schedules nothing. The key is never released, so for that key the failure is permanent.

## Why it hurts

Integration surfaces derive idempotency keys from the triggering event, which is exactly right while an execution is queued, running, or completed — but it also means a turn whose sandbox died mid-flight can never be retried by re-delivering the trigger. On a single-node deployment a code review whose sandbox was killed mid-turn was blocked this way: every re-request "succeeded" at the HTTP layer while returning the same dead execution, until the key itself was changed upstream, which the surfaces do not expose a way to do.

## Where in the code

- The dedupe is an upsert on `(thread_key, idempotency_key)`: [`create_execution_with_request`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/src/lib.rs#L389-L434). The `on conflict ... do update` arm returns the existing row with `created = execution_id = $1` and places no condition on the row's `status`.
- The unique index is partial only on `idempotency_key is not null` — terminal rows are not excluded: [0005_session_handoff_idempotency.sql#L11-L13](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/migrations/0005_session_handoff_idempotency.sql#L11-L13).
- The caller dispatches work only when the returned status is `queued`; a returned `failed` row is logged and handed back as-is: [`enqueue_session_execution`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-runtime/src/lib.rs#L2226-L2277).
- No code path clears or rewrites `idempotency_key` after failure; the upsert above is the only write to that column.

## Proposed direction

Terminal failure should release the key, so the next identical request runs a fresh execution. Options, roughly in order of invasiveness:

- Null out `idempotency_key` when an execution is moved to terminal `failed` (arguably `cancelled` too), keeping the upsert untouched.
- Extend the upsert's conflict arm to re-queue when the existing row is in a terminal failure state (reset `status`/`request`, report it as created), which preserves the execution row history less well but keeps one round trip.
- An explicit retry API for a failed execution id, if silent re-run on re-delivery is considered too surprising.

In-flight (`queued`/`running`) and `completed` executions should keep exactly the current dedupe semantics.

0xAlcibiades:
Closes #1507

## Context

The unique idempotency index on `session_executions` covered every row, so a
failed or cancelled execution permanently held its `(thread_key,
idempotency_key)` pair. For any trigger with a stable key the work then became
un-rerunnable: the insert hit the index and the caller got the dead row back
with `created = false`.

A GitHub review is keyed per commit, so this is reachable in normal operation.
A turn killed mid-flight — a control-plane roll, say — leaves a `failed` row
that no identical re-request can ever get past. The only recovery is editing the
row by hand or pushing a new commit to move the key.

## Change

Migration `0053` swaps the index in place so it covers only the statuses where
dedupe is still meaningful:

```sql
create unique index if not exists session_executions_active_idempotency_idx
    on session_executions (thread_key, idempotency_key)
    where idempotency_key is not null
      and status in ('queued', 'running', 'completed');
```

In-flight and completed work still dedupes, so a retried webhook cannot start a
second live turn or resurrect a finished answer. Failed and cancelled rows drop
out of the index, and the next insert with that key creates a fresh execution.
The upsert conflict clause is narrowed to match the new index.

## Testing

- `cargo fmt --all --check` and `cargo clippy -p centaur-session-sqlx
  --all-targets`: clean.
- `cargo test -p centaur-session-sqlx`: passes.

One caveat worth stating plainly: the new test,
`failed_execution_does_not_block_rerun_on_same_idempotency_key`, is
database-backed and returns early when `SESSION_RUNTIME_TEST_DATABASE_URL` is
unset. I had no Postgres available, so **it reported `ok` without executing, and
I have not exercised the migration or the new index locally.** CI has the
database and will actually run it. The test asserts that a failed execution is
followed by a fresh `created = true` execution on the same key.

`0053` is the next free number on top of `0052_session_execution_request.sql` at
the time of writing; it will need renumbering if another migration lands first.

### DIFF
--- a/services/api-rs/crates/centaur-session-sqlx/migrations/0053_session_execution_retry_idempotency.sql
+++ b/services/api-rs/crates/centaur-session-sqlx/migrations/0053_session_execution_retry_idempotency.sql
@@ -1,0 +1,18 @@
+-- A failed or cancelled execution must not block a fresh execution that
+-- carries the same idempotency key. The original index covered every row,
+-- so once a turn died -- for example on a control-plane roll before the
+-- adoption hardening landed -- the stable key of its trigger (a GitHub
+-- review is keyed per commit) made the work permanently un-rerunnable:
+-- the insert hit the index and the caller got the dead row back with
+-- created = false.
+--
+-- The index now covers only rows whose outcome still matters for dedupe:
+-- in-flight work (queued / running) and completed work. Failed and
+-- cancelled rows fall out of the index, so the next insert with the same
+-- key creates a fresh execution instead of resurrecting the dead one.
+drop index if exists session_executions_thread_idempotency_idx;
+
+create unique index if not exists session_executions_active_idempotency_idx
+    on session_executions (thread_key, idempotency_key)
+    where idempotency_key is not null
+      and status in ('queued', 'running', 'completed');
--- a/services/api-rs/crates/centaur-session-sqlx/src/lib.rs
+++ b/services/api-rs/crates/centaur-session-sqlx/src/lib.rs
@@ -401,6 +401,7 @@ impl PgSessionStore {
             values ($1, $2, $3, $4, $5, $6)
             on conflict (thread_key, idempotency_key)
                 where idempotency_key is not null
+                  and status in ('queued', 'running', 'completed')
             do update set
                 idempotency_key = excluded.idempotency_key,
                 request = case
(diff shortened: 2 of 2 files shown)


mslipper:
Thanks for looking into this. I don’t think we should merge this. It weakens the session API’s global idempotency contract: failed executions may already have produced external side effects, so allowing the same request key to create another execution can duplicate work.

I opened #1603 with a GitHub-specific fix instead. It keys explicit review requests by GitHub delivery ID, so a fresh request can run again while redeliveries remain idempotent. I’m going to close this PR in favor of that one.

Issue #1514, opened 2026-08-27
sandbox pods write raw App Server JSON-RPC frames, including reasoning deltas, to container stdout


0xAlcibiades:
## What happens

Tailing active `asbx-*` agent pods (excluding proxies) on a production deployment at `46bebdf` shows raw Codex App Server JSON-RPC frames on pod stdout/stderr:

- `item/reasoning/textDelta` frames carrying fragments of model reasoning text — model output landing in container logs, where kubectl and any log-shipping pipeline capture and retain it
- `account/rateLimits/updated` frames with every field null, repeated across many pods (pure noise)
- occasional `turn/completed` frames; several pods emit nothing at all in the same window

Example:

```
{"method":"item/reasoning/textDelta","params":{"contentIndex":0,"delta":"    );","itemId":"8ca83e337e146a7e","threadId":"...","turnId":"..."}}
```

## Why it hurts

These frames are the protocol `crates/harness-server` consumes, not operator-facing logging. Reasoning content can include anything the model is working with, so shipping it to the cluster log pipeline both leaks turn content into a second retention system and bloats log volume. The inconsistency (some pods emit frames, some nothing) also makes pod logs useless as an operational signal.

## Proposed direction

Identify which component tees the frames to stdout (harness-server, the harness CLI, or the sandbox entrypoint) and stop emitting protocol frames to the container's stdout/stderr — or gate it behind an explicit debug flag. The null `rateLimits` frames look like they should be dropped at source either way.

PR #1534, opened 2026-08-30
fix(harness-server): stop forwarding rate-limit notifications

### LINKED ISSUE #1556: harness-server forwards account/rateLimits/updated notifications that nothing consumes
Part of #1514 — the `rateLimits` half. That issue covers two separate things; this is the one with a clean fix, so splitting it out to keep the PR's close condition honest.

## What happens

`read_response_or_forward` in `crates/harness-server/src/codex.rs` forwards every codex notification it sees while waiting on a request, verbatim. `account/rateLimits/updated` rides that path with **every field null**, repeatedly, from every sandbox pod.

## Why it is dead weight

The method name appears nowhere else in `crates/harness-server`, and nowhere in `packages/harness-events` — which is where the wire contract clients parse is defined. Nothing on either side of the connection reads it.

Its only effect is volume on a stream that is also the sandbox's container log, so it is retained by whatever log pipeline the deployment runs, at a rate that scales with pod count.

## Expected

The notification is dropped at the forward boundary.

Worth being explicit about how: this should be a **deny list, not an allow list**. Forwarding is otherwise verbatim because the shape clients parse *is* the codex notification shape, so dropping unrecognised methods would silently break a client that does understand one harness-server has never heard of. A new codex method should keep flowing.

## Not in scope

The other half of #1514 — `item/reasoning/textDelta` frames on stdout — is not fixable the same way, and I would rather not conflate them:

1. Clients parse those frames. `packages/harness-events` types `{ method: "item/reasoning/textDelta"; params: ReasoningTextDeltaNotification }` directly, and `packages/rendering` has a test for suppressing reasoning blocks. Dropping them breaks rendering, not just logging.
2. That stdout is also the durable recovery channel. Adoption recovers a turn that completed during a reader gap by reading the sandbox's recorded output, which is the pod log.

So the reasoning leak is real, but it is a retention decision rather than an emission one — unless the recovery path moves off the container log. Left on #1514.

0xAlcibiades:
Closes #1556.

Part of #1514 — the `rateLimits` half. See below for why the reasoning half is not in here.

## Change

`read_response_or_forward` forwards every codex notification it sees while
waiting on a request, verbatim. `account/rateLimits/updated` is forwarded that
way with every field null, repeatedly, from every pod.

Nothing consumes it. The method appears nowhere else in `crates/harness-server`
and nowhere in `packages/harness-events`, which is where the wire contract
clients parse is defined. Its only effect is volume on a stream that is also the
sandbox's container log.

The filter is a **deny list, not an allow list**, on purpose. Forwarding is
otherwise verbatim because the shape clients parse *is* the codex notification
shape, so dropping an unrecognised method would silently break a client that
does understand it. The test asserts a made-up future method still forwards.

## Why the reasoning half is not here

#1514 proposes stopping protocol frames on stdout, or gating them behind a debug
flag. Two things found while implementing say that cannot be done as written,
and I would rather raise them than guess at your protocol:

1. **Clients parse the raw frames.** `packages/harness-events` types
   `{ method: "item/reasoning/textDelta"; params: ReasoningTextDeltaNotification }`
   directly, and `packages/rendering` has a test for suppressing reasoning
   blocks. Dropping the frames would break rendering, not just quieten logs.
2. **That stdout is also the durable recovery channel.** Adoption recovers a
   turn that completed during a reader gap by reading the sandbox's recorded
   output, which is the pod log. Suppressing frames there would remove the
   record recovery depends on.

So the leak is real but the fix is not "stop writing". The narrower framing is
that reasoning content is durable in the cluster log pipeline for as long as pod
logs are retained, which is a retention decision rather than an emission one —
unless the recovery path moves to a channel that is not the container log.

Happy to follow up with whichever direction you prefer.

## Testing

`cargo test -p harness-server`: 92 unit tests pass, including two new ones for
the filter. `cargo clippy --all-targets -- -D warnings` clean.

`tests/app_server_stdio.rs::turn_interrupt_rejects_wrong_thread_and_turn_without_killing_process`
fails in my environment with `read start log: NotFound`. It fails identically on
an unmodified checkout, so it is a local fixture issue and not this change.

### DIFF
--- a/crates/harness-server/src/codex.rs
+++ b/crates/harness-server/src/codex.rs
@@ -608,7 +608,7 @@ impl CodexJsonRpcChild {
                 }
                 return Ok(value.get("result").cloned().unwrap_or(Value::Null));
             }
-            if notification_method(&value).is_some() {
+            if notification_method(&value).is_some_and(should_forward_notification) {
                 write_value(stdout, &value)?;
             }
         }
@@ -910,6 +910,22 @@ fn notification_method(value: &Value) -> Option<&str> {
         return None;
     }
     value.get("method").and_then(Value::as_str)
+}
+
+/// Codex app-server notifications that reach no consumer and so should not be
+/// forwarded onto the blocks stream.
+///
+/// Forwarding is otherwise verbatim, because the wire contract clients parse
+/// (see `packages/harness-events`) is the codex notification shape. That makes
+/// this a deny list rather than an allow list: dropping an unrecognised method
+/// would silently break a client that does understand it.
+///
+/// `account/rateLimits/updated` is the case that earns the list. Nothing in
+/// this crate or in `packages/harness-events` names it, every field arrives
+/// null, and it repeats across every pod, so its only effect is volume on a
+/// stream that is also the sandbox's container log.
+fn should_forward_notification(method: &str) -> bool {
+    method != "account/rateLimits/updated"
 }
 
 fn error_notification_will_retry(value: &Value) -> bool {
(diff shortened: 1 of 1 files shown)


mslipper:
Thanks for looking into this. I'm going to close this PR because it puts deployment logging policy into the App Server protocol adapter. `harness-server` should operate independently of which downstream consumers currently use each notification.

If the goal is to reduce retained log volume, that should be configured in the deployment’s log collector by excluding or applying separate retention to session-sandbox pod output. If protocol output must not enter pod logs at all, we should address the underlying coupling between stdout transport, logging, and recovery rather than filtering individual protocol notifications.

PR #1537, opened 2026-08-30
fix(linearbot): stagger assignment turn starts

### LINKED ISSUE #1510: linearbot: assignment webhook bursts herd-start every delegated issue with no stagger or concurrency bound
### What happens

When a burst of qualifying `Issue` webhooks arrives, linearbot kicks an assignment turn for every delegated issue at effectively the same instant. Each webhook independently passes `parseIssueAssignmentWebhook`, commits the per-issue dedupe watermark, and fires `runThreadTurn` as a fire-and-forget background promise. There is no cross-issue coordination: no stagger, no jitter, no bound on how many turns start concurrently.

Bursts are easy to hit in practice:

- Linear redelivers webhooks that failed while the bot or its control plane was down; after recovery the backlog arrives nearly at once.
- `parseIssueAssignmentWebhook` falls back to a membership-only check when `updatedFrom` is absent, so redelivered or bulk `update` payloads for issues already delegated to the bot qualify as fresh handoffs.

Observed in our deployment (single node, shared local inference backend, `max_running=16`): linearbot fired assignment turns for 11 delegated issues within ~500ms; in a second incident following a control-plane restart it was 14 at once. All of them competed for sandbox admission and the same inference backend, and 8 executions were dropped with `sandbox running capacity exceeded during cold_create/resume`.

### Why it hurts

The capacity check is an admission limit, not a queue. A turn dropped there is terminal from linearbot's perspective:

- The dedupe watermark (`lastAssignmentTrigger`) is committed *before* the turn runs, so even a redelivered webhook for the same change is skipped as a duplicate afterwards.
- The turn itself runs via `backgroundWaitUntil` after the webhook has been acked, so the 503-and-redeliver path (`shouldAwaitLinearHandoff` plus the retryable-error 503 in the webhook handler) covers only the create/append handoff, never turn execution.
- In our incidents nothing landed on the affected issues. Each one sat delegated to the bot with no visible trace that its turn had been dropped; we found the 8 drops only in control-plane logs. Silent drops on delegated issues are the worst failure mode for an autonomous ingress — nobody is watching a thread that never started.

### Where in the code (main @ d0ccee44)

- `services/linearbot/src/index.ts:659` — `handleIssueAssignment`: one kick per webhook, no coordination across issues
- `services/linearbot/src/index.ts:679-688` — dedupe watermark committed before the turn runs
- `services/linearbot/src/index.ts:691-705` — `backgroundWaitUntil(runThreadTurn(...))`, fire-and-forget
- `services/linearbot/src/issue-comments.ts:90` — `parseIssueAssignmentWebhook`; membership-only fallback when `updatedFrom` is absent at `:113-124`
- `services/linearbot/src/index.ts:252-259` and `:1233` — the 503/redelivery contract, scoped to the handoff only
- `services/api-rs/crates/centaur-session-runtime/src/lib.rs:6995` — `SessionRuntimeError::CapacityExceeded`; admission at the `run_with_running_capacity` call for `cold_create` at `:3020`

### Proposed direction

Any of these would have absorbed our incidents; the third matters most:

1. A configurable stagger/jitter between assignment kicks (e.g. `LINEARBOT_ASSIGNMENT_KICK_INTERVAL_MS`), applied when multiple assignment webhooks land within a window.
2. A bound on concurrently *starting* assignment turns, with the excess queued in-process rather than raced into admission.
3. A visible trace when a turn is dropped at admission: post a comment (or reaction) on the issue, and/or roll back `lastAssignmentTrigger` so a webhook redelivery can retry instead of being deduped into the void.

0xAlcibiades:
Closes #1510

## Change

A qualifying `Issue` webhook kicks a turn immediately and independently, so a
burst of assignments starts every turn on the same instant. They reach sandbox
admission in the same tick, spawn their sandboxes together, and then issue their
first inference call together.

Admission is a limit rather than a queue, so a batch that would have fit had it
arrived spread out can still lose the check-then-act race. `runThreadTurn` does
retry, but on the cold-start backoff (250ms, doubling, three attempts) — far too
short to outlast a sandbox running someone else's work, so the losers just post
an error.

`LINEARBOT_ASSIGNMENT_STAGGER_MS` (default 250) spaces consecutive starts, plus a
random extra delay in `[0, staggerMs)` so they miss the grid. The spacing cursor
lives on the process, not the request, because the whole problem is that these
arrive as independent webhooks. Turns are delayed, never dropped.

## The concurrency knob is deliberately off

`LINEARBOT_ASSIGNMENT_CONCURRENCY` bounds turns in flight, and defaults to 0 —
unbounded.

A cap here only means something as this ingress's share of a sandbox fleet shared
with githubbot, slackbot and the console. Shares are meant to oversubscribe —
x/y/z across n pods, summing past n — so a quiet ingress leaves its pods usable
rather than idle. That makes the number a deployment decision, not one this file
can guess, and api-rs leaves its own capacity limit off by default for the same
reason.

A number guessed here is just an invisible work-in-progress cap. The earlier
revision of this PR defaulted it to 2 and held the slot for the whole turn: on a
fleet of 12 sandboxes that silently limited the bot to 2 issues at a time, which
is exactly the shape of bug the gate was supposed to prevent.

## Queued turns say so

When a share is configured the surplus queues rather than being dropped, and a
queued turn now posts a one-line notice on the issue. Waiting is otherwise
indistinguishable from being ignored: nothing posts, the status never moves, and
the only evidence the bot took the work is that it is not doing it.

## What the tests pin

Stagger: a burst arriving in one tick is spaced (`[250, 500, 750]`, the first
start due immediately); idle time is never banked, so a quiet bot starts the next
assignment at once; jitter lands inside the spacing; a zero stagger never sleeps.

Fleet share: the default bounds nothing — six concurrent turns all run; a
configured share admits its own and queues the rest, reporting the depth to the
caller; a throwing turn releases its slot instead of wedging the queue behind it;
and a freed slot goes to the longest-waiting turn rather than being raced for.

## What this does not change

The dedupe watermark still commits before the turn runs, so a queue lost to a
restart is lost. Moving the watermark after the turn would let a redelivery start
a second turn for an issue whose first is still running — a rare silent drop
traded for routine double-execution. With the bound off by default there is no
standing queue to lose; with one configured it is a real caveat, and I am happy
to follow up on the ordering if you would rather have it the other way.

## Testing

`bun test` in `services/linearbot`: 116 pass. `bun run check:types` clean.

### DIFF
--- a/services/linearbot/src/assignment-gate.ts
+++ b/services/linearbot/src/assignment-gate.ts
@@ -1,0 +1,164 @@
+/**
+ * Paces the assignment turns linearbot starts, and optionally bounds how many
+ * of them run at once.
+ *
+ * A qualifying `Issue` webhook kicks a turn immediately and independently, so a
+ * burst starts every turn on the same instant. Bursts are routine: someone
+ * hands the bot a batch of issues, Linear redelivers webhooks that failed while
+ * the bot or its control plane was down, and the membership-only fallback means
+ * redelivered `update` payloads for already-delegated issues qualify as fresh
+ * handoffs.
+ *
+ * Two separate problems come out of that, and this handles them separately.
+ *
+ * **Alignment.** Turns that start on the same instant hit sandbox admission in
+ * the same tick, spawn their sandboxes together, and then issue their first
+ * inference call together — the least useful shape for the model serving behind
+ * them. Admission is a limit rather than a queue, so a batch that would have
+ * fit had it arrived spread out can still lose the check-then-act race, and
+ * `runThreadTurn` only retries on the cold-start backoff (250ms, doubling,
+ * three attempts), far too short to outlast a sandbox running someone else's
+ * work. `staggerMs` spaces consecutive starts so they stop colliding. It delays
+ * turns; it never drops or bounds them, and it is on by default.
+ *
+ * **Fleet share.** `concurrency` is something else entirely: a cap on turns in
+ * flight, which only makes sense as this ingress's slice of a sandbox fleet
+ * shared with the other ingresses. Slices are meant to oversubscribe — x/y/z
+ * across n pods, summing to more than n — so that a quiet githubbot leaves its
+ * pods usable rather than idle. That makes it a deployment-wide decision, not
+ * something this file can guess, so it defaults to off (0). Left off, the only
+ * limit is the fleet's own admission limit, which is also off by default
+ * upstream.
+ *
+ * A bound that is set is a queue, not a rejection: surplus turns wait here for
+ * a slot instead of being dropped. Waiting is invisible on the issue, so `run`
+ * reports it — see `onQueued`.
+ */
+
+export type AssignmentGateOptions = {
+  /**
+   * Turns allowed in flight at once, as this ingress's share of the sandbox
+   * fleet. 0 disables the bound, which is the default: a cap set here that is
+   * unrelated to the fleet's real size is a work-in-progress limit nobody asked
+   * for. Values below 0 are treated as 0.
+   */
+  concurrency: number;
+  /**
+   * Minimum spacing between consecutive turn starts. Each start also waits a
+   * random extra delay in `[0, staggerMs)`, so starts scatter instead of
+   * landing on an exact grid; adjacent turns can swap order, which is the
+   * point. 0 disables the delay.
+   */
+  staggerMs: number;
+  /** Injectable for tests. */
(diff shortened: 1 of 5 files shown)


mslipper:
I think this fix is in the wrong place. `api-rs` should keep executions queued until capacity is available. A Linearbot-local queue can lose work on restart. Please open a new PR that uses `api-rs` to queue Linearbot executions.

PR #1540, opened 2026-08-30
feat(api-rs): retention for session_events

### LINKED ISSUE #1508: session_events grows without bound; no retention for per-line harness output
## What happens

Every event the control plane records for a session is persisted in `session_events`, including one row per harness stdout line, and there is no retention mechanism of any kind: no TTL, no partitioning, and no `delete from session_events` anywhere in `api-rs`. Rows only go away via the `on delete cascade` from `sessions`, and sessions are not deleted in normal operation, so the table grows monotonically for the life of the deployment.

## Why it hurts

On a modest single-node deployment we observed ~1.1M `session_events` rows per day, roughly 450MB+/day of Postgres growth — against the chart's default 20Gi Postgres PVC that is weeks to a full disk on the control-plane database. The bulk of the volume is `session.output.line` rows (per-delta harness output), which are read back for live streaming and recovery of recent executions; multi-week-old output lines for long-terminal executions are dead weight.

## Where in the code

- Table DDL — plain identity-keyed table, no TTL/partitioning: [0001_session_control_plane.sql#L47-L57](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/migrations/0001_session_control_plane.sql#L47-L57).
- Write path: [`append_event`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/src/lib.rs#L1027-L1049) and [`append_event_if_stdout_owner`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/src/lib.rs#L1051-L1100).
- Every harness stdout line becomes a `session.output.line` row: [`append_output_line`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-runtime/src/lib.rs#L6502-L6521) via [`SESSION_OUTPUT_LINE_EVENT`](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-runtime/src/lib.rs#L60); the session runtime has ~37 further `append_event` call sites for lifecycle events.

Worth noting for implementation: the bundled database is `paradedb/paradedb` with `shared_preload_libraries=pg_search,pg_cron` ([contrib/chart/values.yaml#L730-L739](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/contrib/chart/values.yaml#L730-L739)), and migration [0007_absurd_workflows.sql](https://github.com/paradigmxyz/centaur/blob/d0ccee44f1e8bfe69f80985049e83a7296a73338/services/api-rs/crates/centaur-session-sqlx/migrations/0007_absurd_workflows.sql#L2854-L2860) already builds pg_cron-based partition maintenance for the workflow tables — so there is in-tree precedent for scheduled in-database cleanup.

## Proposed direction

A configurable retention window for `session_events`, default off for compatibility: delete events older than N days that belong to terminal executions (or simply all events past the window). Either an in-process sweep on the existing cleanup interval or an opt-in pg_cron job following the migration-0007 pattern would fit; whichever lands should be documented in the configuration reference so operators size the Postgres PVC deliberately instead of discovering the growth rate from a full disk.

0xAlcibiades:
Closes #1508

Adds opt-in, bounded retention for `session.output.line` events after both the event and execution completion time pass the cutoff, preserving other event types. Retention defaults to off, accepts up to 3650 days, and uses an independent sweep interval.

Operators install the partial index concurrently using a documented manual SQL script, outside API startup migrations. Each sweep skips deletion and logs an actionable warning if the index is missing or invalid, then resumes after repair without another restart. Includes installation and recovery instructions and bumps the chart to 0.1.137.

Validated the manual install and a real interrupted index build, skipped sweeps, concurrent repair, and resumed deletion against PostgreSQL. The 25 SQLx tests pass serially; a fresh parallel run encountered the existing migration 0054 deadlock. Formatting, Clippy, Helm lint, and migration-order checks pass.


mslipper:
Thanks for this. I made a few changes prior to merging:

1. Deletion is limited to `session.output.line`. We need to preserve lifecycle and terminal events.
2. The index is now a partial index over just `session.output.line` events. It's also a manual migration now. It's not safe to build this index in the hot migration path because we have no way of knowing how many rows the index will be built over, and if the migration fails it will leave behind a useless invalid index and cause the retention worker to do a full table scan.
3. The worker checks for the new index before running.

PR #1542, opened 2026-08-30
fix(sandbox): bound paused proxies to node capacity

### LINKED ISSUE #1541: Paused sandboxes retain their iron-proxy pod with no bound against node pod capacity
## Problem

Paused sandboxes keep their per-sandbox egress proxy pod for fast resume, with no bound relative to node capacity. When enough accumulate, the node goes past its allocatable pod count and the scheduler refuses new sandbox pods with `FailedScheduling: Too many pods`, so turns fail before they run.

Observed on a single-node deployment: 100 proxy pods against 12 agent pods, 88 of them belonging to paused sessions; the node at 256 scheduled pods against 250 allocatable, and 636 turn failures in 8 hours. Draining the paused proxies by hand took the node back to 160 pods and failures went to zero immediately.

## Why the bound still matters after #1502

#1502 made `pause()` tear the proxy down, so steady-state retention is now zero. That is worth saying plainly, because it means this is no longer the fix for the observed incident on an up-to-date deployment. Three cases still reproduce the same node starvation:

1. **Upgrade state.** Sandboxes paused before #1502 keep their proxy pod indefinitely — nothing else deletes it. Any deployment carrying paused sessions across that upgrade inherits exactly the pod-budget exhaustion above, and has no path out but deleting them by hand.
2. **Crash window.** `pause()` patches `replicas: 0` and then deletes the proxy. A process death between the two leaves a paused CR with a live proxy pod, which nothing reclaims.
3. **Regression.** Any future change that re-introduces retention.

The bound is the invariant that keeps the node schedulable whatever the retention policy is. #1502 removes the accumulation; it does not remove the need for the bound.

## Design

A periodic retention sweep in the k8s sandbox backend:

- Each pass censuses the steering nodes' `status.allocatable.pods` and their scheduled (non-terminating) pod load, then identifies *retained paused proxies*: iron-proxy pods whose Sandbox CR is at `replicas: 0`.
- Per node, retention may hold at most `allocatable − load − margin` pod slots (margin defaults to 2 — the agent pod plus proxy pod a new sandbox schedules). Beyond that, the **longest-paused** proxies are evicted first (LRU by the existing `centaur.ai/paused-at` annotation, sandbox-id tiebreak for determinism).
- An optional absolute cap bounds retention globally, and never expands a node's headroom.
- Eviction deregisters the proxy from iron-control by its durable OID (a pod annotation, so it survives api-rs restarts) and deletes the proxy pod, service and network policies. The Sandbox CR is **untouched**, so resume takes the existing rebuild path and only the evicted session pays a slower resume.
- The sweep never touches running or warm sandboxes, sandboxes with an in-flight resume, sandboxes without an orderable paused-at, or nodes whose pod budget is unreadable (fail-open).
- Each pass logs the per-node budget and each eviction logs its sandbox, so the pod budget is observable from one place instead of being reconstructed from `FailedScheduling` events.

## Config

| Value | Env | Default |
| --- | --- | --- |
| `apiRs.sandboxPausedProxySweepIntervalSecs` | `SESSION_SANDBOX_PAUSED_PROXY_SWEEP_INTERVAL_SECS` | `300` (0 disables) |
| `apiRs.sandboxPausedProxyMarginPods` | `SESSION_SANDBOX_PAUSED_PROXY_MARGIN_PODS` | `2` |
| `apiRs.sandboxPausedProxyCap` | `SESSION_SANDBOX_PAUSED_PROXY_CAP` | `0` (none) |

A read-only ClusterRole for `pods` and `nodes` is required: the namespace-scoped sandbox-manager Role cannot read node allocatable. Without it the sweep fails open and evicts nothing.

## Acceptance criteria

- Unit tests for per-node and cap-based eviction selection: headroom, saturation, cap interaction, unobservable nodes, tie-breaking.
- Sweep logs the per-node budget and each eviction.
- Chart renders the knobs and the read-only cluster RBAC; `helm lint` passes.

0xAlcibiades:
Closes #1541

Bounded retention of paused sandboxes' iron-proxy pods, sized from node capacity. Two commits, the Rust half independently cherry-pickable.

## Read this first: what #1502 changed

#1502 made `pause()` tear the proxy down, so **steady-state retention is now zero** and this is not the fix for the incident in #1541 on an up-to-date deployment. I would rather say that up front than have you find it in review.

What it still covers:

1. **Upgrade state.** Sandboxes paused before #1502 keep their proxy indefinitely; nothing else deletes them. A deployment carrying paused sessions across that upgrade inherits the pod-budget exhaustion with no path out but hand-deletion.
2. **Crash window.** `pause()` patches `replicas: 0` then deletes the proxy. A death between the two leaves a paused CR with a live proxy that nothing reclaims.
3. **Regression cover**, and the per-node pod-budget logging, which is useful on its own — today the budget is only visible by reconstructing it from `FailedScheduling` events.

If you read #1502 as making this redundant, say so and I will close it. I think 1 and 2 are real and unaddressed, and 3 is worth having regardless.

## Commit 1 — `fix(sandbox): bound paused proxies to node capacity`

New `paused_proxy_retention` module in `centaur-sandbox-agent-k8s`:

- Censuses steering nodes' `status.allocatable.pods` and scheduled (non-terminating) pod load, and identifies retained paused proxies: iron-proxy pods whose Sandbox CR is at `replicas: 0`.
- Per node, retention holds at most `allocatable − load − margin` slots; beyond that the longest-paused go first, ordered by the existing `centaur.ai/paused-at` annotation with a sandbox-id tiebreak so selection is deterministic.
- Optional absolute cap, which never expands a node's headroom.
- Eviction deregisters the proxy in iron-control by its durable OID (a pod annotation, so it works across api-rs restarts) and deletes the proxy pod, service and network policies. **The Sandbox CR is untouched** — resume takes the existing rebuild path, so only the evicted session pays a cold create.
- Never touches running or warm sandboxes, in-flight resumes (in-process guard set at `resume()` entry and cleared on every exit path), unorderable pauses, or nodes with an unreadable budget. That last one is fail-open on purpose: evicting against an unknown budget is worse than not evicting.

## Commit 2 — `feat(chart): paused-proxy retention knobs and node-budget RBAC`

Renders the three env vars, adds the read-only ClusterRole/Binding (`pods`, `nodes` get/list) that the namespace-scoped sandbox-manager Role cannot provide, documents the knobs, and adds them to `values.schema.json`.

The split keeps commit 1 cherry-pickable on its own: without the chart half the sweep cannot read nodes, fails open, and evicts nothing.

## Testing

10 new tests — 8 on eviction selection (headroom, saturation to zero, cap interaction, cap-never-expands-headroom, unobservable node, tie-breaking, config toggle) and 2 on arg parsing.

`cargo test -p centaur-sandbox-agent-k8s -p centaur-api-server`: 122 and 56 pass. `cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` clean. `helm lint` passes and `helm template` renders the env block and both RBAC objects.

### DIFF
--- a/contrib/chart/templates/apirs.yaml
+++ b/contrib/chart/templates/apirs.yaml
@@ -176,6 +176,34 @@ roleRef:
   kind: Role
   name: {{ include "centaur.componentName" (dict "root" . "component" "api-rs-sandbox-manager") }}
 ---
+# Read-only cluster census for the paused-proxy retention sweep: it sizes
+# proxy eviction against each steering node's allocatable pod count, which
+# the namespace-scoped Role above cannot see.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "centaur.componentName" (dict "root" . "component" "api-rs-sandbox-node-budget") }}
+  labels:
+{{ include "centaur.componentLabels" (dict "root" . "component" "api-rs") | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "nodes"]
+    verbs: ["get", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "centaur.componentName" (dict "root" . "component" "api-rs-sandbox-node-budget") }}
+  labels:
+{{ include "centaur.componentLabels" (dict "root" . "component" "api-rs") | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ $apiRsName }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "centaur.componentName" (dict "root" . "component" "api-rs-sandbox-node-budget") }}
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -330,6 +358,15 @@ spec:
               value: {{ .Values.apiRs.sandboxCleanupIntervalSecs | quote }}
             - name: SESSION_SANDBOX_IDLE_CLEANUP_BACKSTOP_SECS
               value: {{ .Values.apiRs.sandboxIdleCleanupBackstopSecs | quote }}
+            # Paused-proxy retention sweep. api-rs needs read-only cluster
+            # pods+nodes to size proxy eviction against each node's
+            # allocatable pod count (see the ClusterRole above).
+            - name: SESSION_SANDBOX_PAUSED_PROXY_SWEEP_INTERVAL_SECS
+              value: {{ .Values.apiRs.sandboxPausedProxySweepIntervalSecs | quote }}
+            - name: SESSION_SANDBOX_PAUSED_PROXY_MARGIN_PODS
+              value: {{ .Values.apiRs.sandboxPausedProxyMarginPods | quote }}
+            - name: SESSION_SANDBOX_PAUSED_PROXY_CAP
+              value: {{ .Values.apiRs.sandboxPausedProxyCap | quote }}
             - name: SESSION_ACTIVITY_SUMMARY_ENABLED
               value: {{ .Values.apiRs.activitySummary.enabled | quote }}
 {{- if .Values.apiRs.activitySummary.enabled }}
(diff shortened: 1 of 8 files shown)


mslipper:
Thanks for writing this up. I'm going to close this PR. #1502 already made steady-state proxy retention zero, and the remaining upgrade/crash issues are bounded by the sandbox reaper. The new capacity controller, configuration, and cluster-wide pod/node RBAC feel disproportionate compared to that, especially now that #1529 will reap orphaned proxies.

PR #1544, opened 2026-08-30
feat(api-rs): wait for a sandbox slot instead of dropping the turn

### LINKED ISSUE #1543: Sandbox admission drops turns at the running limit instead of waiting for a slot
## What happens

Sandbox admission is a limit, not a queue. When the running cap is reached and
`ensure_running_slot` cannot free one — warm sandboxes stopped, idle sessions
paused, stale assignments reconciled, and still over cap — `cold_create` and
`resume` fail immediately with `CapacityExceeded`.

That refusal is terminal. Nothing retries it, and nothing surfaces on the
originating PR, issue or thread; the only trace is an api-rs log line. A burst
of concurrent turns therefore loses every turn past the cap, while the slots
they needed free seconds later as idle sandboxes pause.

## Observed

On a single-node deployment with `max_running=12`:

- An assignment burst of ~10 turns within one second; the turns past the cap
  failed with `capacity exceeded during resume: running=12, max_running=12`.
- A separate window with 9 × `capacity exceeded during cold_create` plus 3
  `failed to ensure session sandbox` in under ten minutes.
- In an earlier incident at `max_running=6`, three consecutive executions on one
  thread were dropped inside 17 minutes, none retried, nothing rendered to the
  user.

The pool drains normally within seconds to minutes in each case. The turns are
lost to the timing of the check, not to genuine sustained saturation.

## Why waiting is the right shape

The eviction machinery in `ensure_running_slot` is already correct and already
idempotent — it simply runs once. Everything needed to admit these turns exists;
what is missing is a second look.

Raising the cap is not equivalent. It trades one silent failure for another
(memory pressure on the node) and does nothing for the burst that exceeds the
new number.

## Proposed direction

Re-check admission on a short tick until a slot opens or a bounded budget
expires, configurable and defaulting to something well under a human's patience.
Three properties matter:

1. **Release the admission lock while waiting.** Holding it head-of-lines every
   other turn behind the blocked one, including the turns whose completion frees
   the slot it is waiting for.
2. **Re-take it per attempt**, so check-then-act stays atomic and the cap is
   still never exceeded.
3. **Record the wait once, durably.** A queued turn that looks identical to a
   hung one is only a partial improvement; the waiting state should be visible
   on the session.

On timeout the turn should fail with exactly the current error, so a genuinely
saturated pool behaves as it does today, and `0` should restore fail-at-once as
a rollout switch.

## Acceptance

- A blocked admission that sees a slot free mid-wait runs its action.
- A blocked admission that never sees one fails with the current
  `CapacityExceeded` message.
- The wait is recorded exactly once per blocked admission.
- Existing capacity-controller tests pass unchanged.

0xAlcibiades:
Closes #1543

## Change

`SESSION_SANDBOX_ADMISSION_WAIT_SECS` (chart: `apiRs.sandboxAdmissionWaitSecs`,
default `600`). A blocked admission re-checks on a 2s tick until a slot opens or
the budget expires, instead of failing the turn at once.

The eviction machinery in `ensure_running_slot` is unchanged and already
idempotent, so re-running it per attempt is safe. What this adds is the second
look.

## Three properties worth reviewing

**The admission lock is released while waiting.** Holding it would head-of-line
every other turn behind the blocked one — including the turns whose completion
frees the slot it is waiting for, which would deadlock the pool under exactly
the burst this fixes.

**It is re-taken per attempt**, so check-then-act stays atomic and the cap is
never exceeded. That is why this is a re-check loop rather than a condvar on the
held lock.

**The wait is recorded once, durably.** The first block appends one
`session.sandbox_capacity_waiting` event with `running`, `max_running` and the
remaining budget. A queued turn that looks identical to a hung turn is only half
a fix. A failed append warns and does not fail the wait: losing the event costs
visibility, losing the turn costs the work.

## On the default

`600` changes behaviour for every deployment, so it is worth being explicit: the
current behaviour is that a turn past the cap is silently destroyed, and the
timeout path still fails with exactly the same `CapacityExceeded` error, so a
genuinely saturated pool is unaffected. `0` restores fail-at-once as a rollout
switch. I am happy to default it to `0` instead if you would rather this be
opt-in.

## Testing

New test: two running sandboxes fill a pool of 2 with nothing evictable, a slot
frees after a second, and the admission is granted rather than refused —
asserting both that it waited and that exactly one waiting event was recorded
with the right payload. The mock's default status is `Suspended` so stale rows
from other tests skip rather than freeing a slot early and hiding the wait.

The existing eviction test is pinned to `admission_wait: ZERO`, so it still
asserts the fail-fast path.

`cargo test -p centaur-session-runtime -p centaur-api-server` passes (122 / 115 /
54). `cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings`
clean. `helm lint` passes and `helm template` renders the value.

One caveat: the new test is database-backed and returns early without
`SESSION_RUNTIME_TEST_DATABASE_URL`. I had no Postgres available, so **it
reported `ok` without executing** and the wait path rests on review and CI here.


mslipper:
Closing this because the current approach is not durable. Waiting happens in process after the execution is marked running, so deploys or crashes can still lose the turn. It also bypasses caller timeouts.

If we revisit this, the design should keep the execution durably queued or introduce a persisted waiting-for-sandbox state with recoverable ownership.

PR #1550, opened 2026-08-30
feat(harness-server): let a deployment name retriable engine errors

### LINKED ISSUE #1549: Retriable engine errors are hardcoded, so a deployment cannot cover its own endpoint's transient failures
## What happens

`is_retriable_engine_error` decides whether a failed turn is transparently
re-submitted. It matches two hardcoded substrings — `Engine not found`, and
`Job registration failed` with `404` — which cover codex's own
engine-registration failures.

Those are the only transient failures it can cover. `OPENAI_BASE_URL` points the
harness at any OpenAI-compatible endpoint, and a deployment whose endpoint has a
different transient failure has no lever at all: the classifier is compiled in.

## The case that surfaced it

An endpoint rejecting a request body it could not parse:

```json
{"error":{"message":"Unterminated string starting at: line 1 column 9 (char 8)",
          "type":"BadRequestError","param":null,"code":400}}
```

Twice in one window, byte-identical including the offset. The proxy audit log
shows the same signature both times — a long successful POST, then a brand new
connection ~230ms later carrying a request the upstream rejects in 15–25ms:

```
22:06:01.171  POST /v1/responses   200   duration_ms 12470
22:06:01.403  CONNECT              200   (new connection)
22:06:01.421  POST /v1/responses   400   duration_ms 14.8
```

The 400 returns before any work, so nothing was streamed. `char 8` in a body
opening `{"model":"..."` is exactly where the model-name string begins, so the
server read 8 bytes and hit a string that never terminates — the shape of a body
truncated very early, or a framing mismatch between what was sent and read.

The turn dies, and because a failed execution holds its idempotency key it
cannot be re-run on the same trigger, so the cost is the whole turn.

## What I am not claiming

Which side truncates the request is **unidentified**. The proxy's transform on
these requests is header-only (an `Authorization` swap, reported `allow` on the
failing request too), so it relayed what it was given — but that leaves both the
client's encoding on a fresh connection after a long stream, and the server's
body read under concurrency, as live candidates. Settling it needs a captured
body.

So this is not a request to hardcode a new pattern upstream. Asserting a
classification you cannot verify would be worse than leaving it configurable.

## Proposed direction

Let a deployment extend the retriable set — an env naming additional error
message substrings, empty by default so nothing changes unless opted into.

The existing guards make this safe to widen: retries are bounded by
`CODEX_ENGINE_RETRY_MAX`, and the retry path is only reached before the turn has
streamed any output. A pattern that turns out to match a deterministic failure
costs a few resubmissions and then fails exactly as it does today.

## Acceptance

- Additional retriable patterns are configurable and empty by default.
- The built-in engine-registration patterns still match with extras configured.
- An unrelated error is not matched.

0xAlcibiades:
Closes #1549

## Change

`CODEX_ENGINE_RETRY_PATTERNS` — comma-separated substrings added to the built-in
retriable set, matched case-insensitively. Empty by default, so nothing changes
unless a deployment opts in.

## Why a lever and not a new hardcoded pattern

The case that surfaced this is an OpenAI-compatible endpoint rejecting a request
body it could not parse (`Unterminated string starting at: line 1 column 9`),
returning 400 in 15–25ms with no work done. I could have added that string to
the allowlist. I did not, because **which side truncates the request is still
unidentified** — both the client's encoding on a fresh connection after a long
stream and the server's body read under concurrency remain live candidates, and
settling it needs a captured body.

Hardcoding a pattern for a failure whose cause you cannot verify asserts a
classification into everyone's build. Letting the deployment that can observe it
decide seemed the better trade, and it also fixes the general problem:
`is_retriable_engine_error` covers codex's own registration failures and cannot
cover an arbitrary endpoint's, so any deployment hitting a different transient
failure currently has to patch this binary.

## Why widening is safe

Both existing guards still apply, and they are what make this bounded:

- retries are capped by `CODEX_ENGINE_RETRY_MAX`;
- the retry path is only reached **before the turn has streamed output**
  (`!self.streamed`), so nothing partial is ever re-submitted.

A pattern that turns out to match a deterministic failure therefore costs a few
resubmissions and then fails exactly as it does today.

## Testing

Five new tests over the classifier, split from the env read so they need no
process-global state: off by default (the reported error is not retriable with
no patterns configured, and the built-ins still are), extras extending rather
than replacing the built-in set, case-insensitive matching, an unrelated error
not matching, and the parser ignoring blanks and whitespace.

`cargo test -p harness-server`: 95 unit tests pass. `cargo clippy --all-targets
-- -D warnings` clean.

`tests/app_server_stdio.rs::turn_interrupt_rejects_wrong_thread_and_turn_without_killing_process`
fails in my environment with `read start log: NotFound`, identically on an
unmodified checkout, so it is a local fixture issue rather than this change.

### DIFF
--- a/crates/harness-server/src/codex.rs
+++ b/crates/harness-server/src/codex.rs
@@ -869,19 +869,62 @@ fn retry_backoff(retry: u32) -> Duration {
     Duration::from_millis((500u64 << shift).min(5_000))
 }
 
+/// Extra error-message substrings a deployment considers retriable, from
+/// `CODEX_ENGINE_RETRY_PATTERNS` (comma-separated). Empty by default, so
+/// behaviour is unchanged unless a deployment opts in.
+///
+/// The built-in set below covers codex's own engine-registration failures. It
+/// cannot cover the transient failures of an arbitrary inference endpoint, and
+/// a deployment that hits one today has no lever short of patching this binary
+/// -- for example an OpenAI-compatible server rejecting a request body it could
+/// not parse, which is a framing failure that succeeds on resubmission.
+fn configured_retry_patterns() -> Vec<String> {
+    parse_retry_patterns(env::var("CODEX_ENGINE_RETRY_PATTERNS").ok().as_deref())
+}
+
+fn parse_retry_patterns(raw: Option<&str>) -> Vec<String> {
+    raw.map(|raw| {
+        raw.split(',')
+            .map(str::trim)
+            .filter(|pattern| !pattern.is_empty())
+            .map(|pattern| pattern.to_lowercase())
+            .collect()
+    })
+    .unwrap_or_default()
+}
+
 /// True for codex's transient "engine warming up" failure, which surfaces as a
 /// -32602 error notification (`willRetry:false`) whose message ends in
 /// "...status 404 Not Found: Engine not found". These resolve on resubmission;
 /// other -32602s (genuine invalid params) are left untouched.
+///
+/// Retries are bounded by `CODEX_ENGINE_RETRY_MAX` and only reached before the
+/// turn has streamed any output, so a pattern that turns out to match a
+/// deterministic failure costs a few resubmissions and then fails exactly as it
+/// does today.
 fn is_retriable_engine_error(value: &Value) -> bool {
+    is_retriable_error_message(value, &configured_retry_patterns())
+}
+
+fn is_retriable_error_message(value: &Value, extra_patterns: &[String]) -> bool {
     let Some(message) = value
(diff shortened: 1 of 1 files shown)


mslipper:
Thanks for investigating this. I'm closing this PR because configuring retries through error-message matching is not safe.

`Unterminated string` is a generic parse error, not necessarily evidence of a transient failure. It could come from anywhere. A deployment-wide regex would silently classify all instances as retryable, which could lead to undefined behavior. I'd rather investigate the root cause of the unterminated string errors you're seeing and fix that instead.

PR #1560, opened 2026-08-30
fix(linearbot): stop working an issue that is taken back

### LINKED ISSUE #1559: linearbot keeps working an issue after it is un-delegated
Taking an issue back from linearbot has no effect. The bot finishes the turn
regardless, holding a sandbox for work nobody wants any more.

## Why nothing happens

`parseIssueAssignmentWebhook` returns `null` unless the issue is *currently*
assigned or delegated to the bot:

```ts
const assignedToBot = stringValue(data.assigneeId) === botUserId;
const delegatedToBot = stringValue(data.delegateId) === botUserId;
if (!assignedToBot && !delegatedToBot) return null;
```

An un-delegation webhook is precisely the payload where both are false, so the
event that means "stop" is the one the parser drops. Nothing else in linearbot
watches `Issue` updates, so it is discarded outright.

## What that costs

Two shapes, depending on how far along the turn is.

**Not started yet.** A handoff turn does not begin the instant the webhook
lands — it waits its place among concurrent handoffs, then spends seconds
fetching issue context and spawning a sandbox. Un-delegate inside that window
and the turn still starts afterwards: the bot posts on an issue it no longer
holds, and moves it to In Progress.

**Already running.** The execution runs to completion. Its sandbox counts
against `--session-sandbox-running-limit` the whole time, so on a busy fleet an
abandoned turn is refusing admission to a real one, and the answer lands on an
issue whose owner changed while it was being written.

## Suggested behaviour

Un-assigned *and* un-delegated is a release. Drop a turn that has not started
(nothing posted, no sandbox, so never starting is the whole cleanup) and
interrupt one that has via `POST /api/session/{thread_key}/interrupt`, which
already exists and ends the execution so the sandbox falls to the normal idle
reclaim.

Two cases that should NOT count as a release: losing one field while still
holding the other (a delegate that is still the assignee is still meant to be
working), and the bot releasing the issue itself, which is how an agent
normally signs off at the end of a turn.

0xAlcibiades:
Closes #1559

## Change

`parseIssueReleaseWebhook` is the mirror of `parseIssueAssignmentWebhook`, and
has to be a separate parser because that one returns `null` the moment the bot
is no longer on the issue — the payload that says "stop" is exactly the one it
discards.

A release is the bot being dropped as assignee **and** as delegate, with
`updatedFrom` naming the bot as the previous value. That last part is load
bearing: "not assigned to the bot" describes almost every issue in a workspace,
so without it every unrelated edit anywhere would qualify.

Two cases deliberately do not count:

- **Losing one field while still holding the other.** A delegate that is still
  the assignee is still meant to be working.
- **The bot releasing the issue itself.** Handing the issue back at the end of a
  turn is how an agent signs off; acting on it would interrupt the turn that
  just did the work.

## Handling splits on whether the turn started

A handoff turn does not begin the instant its webhook lands — it waits its place
among concurrent handoffs, then spends seconds fetching issue context and
spawning a sandbox. So `pendingAssignments` tracks, per thread, whether the turn
has started, because the two cases need different things:

- **Not started.** Dropped where it stands. Nothing has been posted and no
  sandbox exists, so never starting is the whole of the cleanup.
- **Started.** Interrupted through `POST /api/session/{thread_key}/interrupt`,
  which ends the execution and lets the sandbox fall to the normal idle reclaim
  instead of holding a slot against `--session-sandbox-running-limit` for the
  rest of a turn nobody wants.

A thread this process has no record of is interrupted too. A turn that outlived
a restart is exactly the case where the issue looks busy and no local state
explains it, and `interrupt` reports `interrupted: false` harmlessly when there
is nothing running.

The map is per process, like the work it tracks. A restart loses the entries,
which costs nothing: it also loses the turns.

## What this does not do

It does not touch the issue's status. Whoever took the issue back decides where
it belongs, and moving it here would fight them.

## Testing

Nine parser tests: dropped as delegate; dropped as assignee; handed to someone
else outright; still held by the other field (no fire); an unrelated edit on
someone else's issue (no fire); no `updatedFrom` at all (no fire); the bot
releasing it itself (no fire); a `create` (no fire); and `updatedFrom` nested
under `data`.

`bun test` in `services/linearbot`: 117 pass. `bun run check:types` clean.

### DIFF
--- a/services/linearbot/src/index.ts
+++ b/services/linearbot/src/index.ts
@@ -19,6 +19,7 @@ import pg from "pg";
 import {
   parseIssueAssignmentWebhook,
   parseIssueCommentWebhook,
+  parseIssueReleaseWebhook,
   type IssueAssignmentEvent,
   type IssueCommentEvent,
 } from "./issue-comments";
@@ -57,6 +58,7 @@ import { extractMessageOverrides, resolveStickyProvider } from "./overrides";
 import {
   executeSessionTurn,
   forwardToSessionApi,
+  interruptSession,
   isRetryableSessionApiError,
   openSessionEventStream,
   serializeMessage,
@@ -241,6 +243,9 @@ export function createLinearbot(options: LinearbotOptions): Linearbot {
         ) ??
         requestContext.run(context, () =>
           handleIssueAssignment(rawBody, handlerInput),
+        ) ??
+        requestContext.run(context, () =>
+          handleIssueRelease(rawBody, handlerInput),
         );
       if (handled) handoffTasks.push(handled);
       try {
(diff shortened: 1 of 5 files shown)


0xAlcibiades:
Folded into #1537 — the release path and the start stagger are the same handoff path, and reviewing them apart meant reading the same call site twice.

0xAlcibiades:
Reopened — this and #1537 touch the same call site but are separate fixes, so they are easier to review apart. #1537 lands first; this one rebases onto it.

PR #1564, opened 2026-08-30
feat(api-rs): enforce a node memory budget on sandbox admission

### LINKED ISSUE #1563: Sandbox admission is blind to node memory; OOM kills lose in-flight turns
## What happens

Sandbox admission counts slots (`sandboxRunningLimit`) and is blind to the memory those slots commit. On a single-node deployment, every sandbox pod commits a large agent-container memory limit plus its iron-proxy limit, and the sum of committed limits across the running set can reach multiples of the node's allocatable. When a co-located working set peaks, the kernel OOM-killer takes sandbox containers. The in-flight agent work dies with the container and the turn is lost (no retry under the same idempotency key).

## Observed

On a single-node deployment with a 24 GiB container memory limit:

- Node-level kernel OOM kills repeatedly (two per day) across two days, at 60s-scrape working sets of 12.4 GiB and 4.6 GiB.
- `container_oom_events_total = 0` for both affected pods — this is node-level kernel OOM, not limit-driven cgroup OOM.
- The admission counter at the time was within `max_running`; the node was simply out of memory for the sum of committed limits.
- Each kill destroyed an in-flight turn with no retry, no user-visible error, and no trace beyond the kubelet event.

## Why it hurts

The running limit is a count, not a memory budget. An operator who sets `sandboxRunningLimit=12` is implicitly committing 12 × (agent limit + proxy limit) of memory to the node, with no feedback if that exceeds what the node can sustain. The failure mode is silent: the turn is destroyed, nothing surfaces on the originating thread, and recovery requires a human to notice the OOM event and restart the sandbox.

## Proposed direction

An operator can size a node memory budget for running session sandboxes. Admission then enforces the smaller of the running limit and `budget / per-sandbox committed size`, where the per-sandbox size is derived from the same resource settings api-rs already uses to build the pods (agent limit plus iron-proxy limit). Every admission decision — admit, wait, refuse — records the computation, so a refusal names which budget turned the turn away.

Config: `SESSION_SANDBOX_MEMORY_BUDGET_GIB` / chart `apiRs.sandboxMemoryBudgetGib`, default 0 (disabled). Unset budget changes nothing: count-based admission exactly as today.

## Acceptance

- Unset budget (0) changes nothing: count-based admission exactly as today.
- Set budget below the count limit: the effective cap is `budget / per-sandbox`; with nothing evictable, the turn is refused with `CapacityExceeded` carrying `admission_limit`, `memory_budget_bytes`, `per_sandbox_memory_bytes`.
- `sandbox_capacity_admitted` event and the refusal error carry the computation.
- Startup fails, with an actionable message, when the budget cannot be sized: budget set with no running limit; no `SESSION_SANDBOX_RESOURCES`; agent memory limit absent or not a whole-byte quantity; budget below one sandbox's share.
- Tests: quantity parsing; `admission_limit` arithmetic (count-only, memory-only, intersect, undersized budget); args wiring and each startup-failure case.

Related: #1509 (running count drift), #1543 (admission drops turns at the running limit), #1454 (OOMKilled status detail).

0xAlcibiades:
Closes #1563

## Change

`SESSION_SANDBOX_MEMORY_BUDGET_GIB` (chart: `apiRs.sandboxMemoryBudgetGib`, default `0` = disabled). Admission enforces the smaller of the running limit and the budget divided by one sandbox's committed share (agent memory limit plus iron-proxy memory limit), and records the computation in the `sandbox_capacity_admitted` event and the capacity refusal error.

The per-sandbox size is derived, not configured: it reads `SESSION_SANDBOX_RESOURCES` `limits.memory` plus `KUBERNETES_IRON_PROXY_RESOURCES` `limits.memory` — the same values api-rs uses to build the pods.

## Startup validation

A budget that cannot be sized fails startup with an actionable message:

- Budget set with no running limit → the budget caps the limit, it does not replace it.
- No `SESSION_SANDBOX_RESOURCES` → admission cannot size one sandbox against the budget.
- Agent memory limit absent or fractional → cannot compute a whole-byte share.
- Budget below one sandbox's share → mis-sized; would refuse everything.

## Error message

The refusal now names the computation:

```
sandbox running capacity exceeded during cold_create: running=2, admission_limit=2 (max_running=12, memory_budget_bytes=Some(107374182400), per_sandbox_memory_bytes=Some(42949672960))
```

## Testing

- 3 new tests in `centaur-sandbox-core`: quantity parsing (fleet shapes, unsizable forms, limit map read).
- 5 new tests in `centaur-api-server`: admission arithmetic (count-only, memory-only, intersect, undersized budget), args wiring, each startup-failure case.
- `cargo test -p centaur-sandbox-core -p centaur-session-runtime -p centaur-api-server`: 122 + 114 + 59 pass.
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `helm lint` passes and `helm template` renders the value.

## Dependency note

The error-message and event fields added here are additive to the existing
`CapacityExceeded` variant and `sandbox_capacity_admitted` event. If #1544
(admission wait) lands first, its `sandbox_capacity_waiting` event should pick
up the same fields — the `admission_limit()` method makes the computation
available to any caller of the config.


mslipper:
I don't think we should add this. Kubernetes should be the source of truth for node capacity and memory allocation.

We already expose `sandbox.resources` and `ironProxy.resources`, so operators can set memory requests and limits for both pods. The scheduler will use those requests when deciding whether another sandbox fits on an eligible node. If memory must not be overcommitted, set each memory request equal to its limit.

We also apply `sandbox.priorityClassName` to both the sandbox and its paired proxy. Operators can use that PriorityClass to scope a ResourceQuota, limiting the aggregate memory allocated to sandbox workloads, while allowing lower-priority sandboxes to be preempted or evicted before control-plane workloads.

PR #1589, opened 2026-09-01
feat(githubbot): skip centaur-skip checks, and read check detail without the Checks API

oponder:
Two changes to githubbot's CI evaluation; the second is what makes the first work on a fine-grained PAT.

**Skip a check.** A check whose job id contains `centaur-skip` is dropped from the CI evaluation and from the escalation comment's failing list. Some required checks can never be turned green by an agent — an N-human-approvals gate, a manual sign-off — and today githubbot burns all three fix attempts on them and then escalates. The marker rides the job id, which GitHub uses as the check-run name and recomputes every run, rather than the workflow name, which GitHub caches per consuming repo for ruleset-injected required workflows and never refreshes.

**Read check detail without the Checks API.** A fine-grained PAT — the token type GitHub recommends — has no Checks permission, so `statusCheckRollup` returns null check nodes and the evaluation collapses to the bare aggregate: `failingNames` is always empty, fix turns get "Failing checks: unknown", escalations say "Still failing: the CI checks". When the nodes are unreadable, githubbot now rebuilds the checks from the Actions API (a job is the same object as its check run) and trusts that only when it accounts for every context GitHub counted — commit statuses and other Apps' check runs aren't Actions jobs — otherwise it keeps the existing aggregate behaviour.

No new configuration; auto-merge is untouched. Verified end-to-end on a deployment using a fine-grained PAT: gate-only red evaluates green, and gate plus a real failure fires exactly one fix turn that names only the real failure.

https://claude.ai/code/session_01NY7uaJuKZDbHdD9UaBQZVc

### DIFF
--- a/services/githubbot/AGENTS.md
+++ b/services/githubbot/AGENTS.md
@@ -26,6 +26,9 @@ behavioral contract and supported webhook events are documented in `README.md`.
 - Automated merge behavior applies only to explicitly owned PRs and must honor
   draft, hold, mergeability, settled-check, attempt-limit, and human-handoff
   gates. Keep deterministic merge decisions outside the agent prompt.
+- The Actions fallback may stand in for unreadable check detail only when it
+  accounts for every context GitHub counted. An unseen check must never read as
+  green.
 - Bundled prompts remain generic and fully overrideable. Do not embed private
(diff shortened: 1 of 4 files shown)


mslipper:
Superseded by #1683.

PR #1592, opened 2026-09-03
fix(workflows): route reaction invites to Slack live queue

decofe:
## Summary

- route the Slack reaction calendar poller to the isolated Slack-live workflow queue
- route its DM sender to the same queue
- cover both workflow names in the queue-classification test

## Validation

- `cargo +nightly fmt --package centaur-workflows -- --check`
- `cargo test -p centaur-workflows scheduled_etls_use_isolated_etl_queues`
- `cargo +nightly clippy -p centaur-workflows --all-targets -- -D warnings`

Prompted by: @letstokenize


mslipper:
Is this a tempo-specific workflow? If so we should find a better way of configuring this than hardcoding.

PR #228, opened 2026-05-27
feat(websearch): Parallel-backed search + deep research

NormallyGaussian:
## Summary
Reimplements the `websearch` tool on [Parallel Web Systems](https://docs.parallel.ai) — same interface, strictly better behavior. Search now works with **zero credentials** via the free hosted Parallel Search MCP, and `deep_research` is rebuilt on the Parallel Task API so it returns **correctly-cited reports** instead of failing citation validation every run.

## What changed
- **`search`** — Parallel retrieval, layered by credential:
  - no key → free hosted Search MCP (anonymous; attribution surfaced in `meta.attribution`)
  - `PARALLEL_API_KEY` → Search REST (domain/recency filters, `num_results`, `mode`)
  - `ANTHROPIC_API_KEY` → optional Claude reviewer→writer→citation-repair synthesis (pipeline unchanged from prod)
- **`deep_research`** — Parallel Task API with auto schema; streams progress, renders the model's cited report with a matching **1-based** `## Sources` block. Gated to the `pro`/`ultra` family.
- Uses the official `parallel-web` SDK (`>=0.6.0`) for REST + Task; the Search MCP is a hand-rolled Streamable-HTTP JSON-RPC client.
- Removed Exa.

## Deep research: quality + cost

**Citations — before vs. after.** Prod's citation validator (`_extract_sources_section_ids`) only matches source lines like `[N] url`, but the writer emits **bulleted** lines `- [N] url`. The regex therefore matches nothing, so every report is flagged `synthesis failed: Sources section missing cited IDs: [...]` even when the report is perfectly cited (reproduced live on the retained synthesis path — flagged on *every* call, plus 2 wasted Claude repair attempts per query). This PR fixes that regex **and** moves `deep_research` to the Task API, whose auto-schema output is model-grounded:

- inline `[N]` markers are 1-based into the `basis` citation list; we render the `## Sources` block 1-based to match. Verified on live output:
  - prose "parallel.ai **[1]**" → Sources `[1]` = `https://parallel.ai/`
  - prose "LinkedIn **[2]**" → Sources `[2]` = `linkedin.com/company/parallel-web`
  - prose "Fortune **[11]**" → Sources `[11]` = `fortune.com/2026/05/19/...`
  - dangling citations: **none**
[...]

PR #1600, opened 2026-09-04
Add swappable websearch backend, make Tako search default

jed326:
Supersedes https://github.com/paradigmxyz/centaur/pull/1175
Addresses https://github.com/paradigmxyz/centaur/pull/228#pullrequestreview-4376811566

As requested in the review comment above, this PR introduces a generic, swappable backend to the websearch tool and uses [Tako Search](https://tako.com/search) as the default backend.

All existing functionality is preserved and keyless search still works through Tako MCP. The Tako backend uses the async lane of [`tako-sdk`](https://pypi.org/project/tako-sdk/) for its request and response models.

## Testing

- 84 unit tests across 5 files, run the way `run-tool-tests.sh` runs them; `validate_cli_packaging.py` and `uv build --wheel` pass. The unit tests cover the stream paths a live run can't force: a dropped stream resuming with `starting_after`, a non-SSE body on resume, an `agent_result` frame the SDK can't parse, and a stream that ends without a result, each falling through to polling the run; plus the 401 and 403 fallback from keyed to anonymous search.
- Against production from a workstation: anonymous search, keyed search (`tako:api`, plain, `--effort deep`, and `--include-domain` with `--max-age-hours`), keyed deep-research (`tako:agent`, 99s, $0.15), and the fallback for a bad Tako key. Parallel keyed and anonymous, the bad Anthropic key, and a blank or bogus `WEBSEARCH_BACKEND` were exercised before the Tako backend moved onto the SDK; that code did not change.
- From a sandbox on a local kind cluster with `jed326/centaur@websearch-tako-backend` as `toolServer.ref` (its tool tree is identical to this PR's): the shim reinstalled with the SDK dependency, `websearch --help` works, an ungranted principal gets `tako:anonymous`, and the principal with `centaur-perms principals grant --tool websearch` gets `tako:api` plus a deep-research report (`tako:agent`). The SDK's httpx client trusted the sandbox proxy's CA and the proxy rewrote `X-API-Key`. Sandbox pods never hold `TAKO_API_KEY`; only the per-sandbox proxy does.


jed326:
Hey @gakonst or @mslipper , could I get a review on this one? Thanks!

mslipper (changes requested):
Thanks for putting this together. I like making the websearch tool backend-agnostic, but I don't want Tako to become the default. Existing deployments should continue using Parallel unless they explicitly set WEBSEARCH_BACKEND=tako.

This PR is also several changes bundled together: the provider, the Tako implementation itself, various changes to the shared CLI/API, and refactors to the existing Parallel implementation. Can we split this into:

1. A provider-neutral refactor that preserves the current Parallel behavior and public contract.
2. A follow-up that adds Tako as an explicitly configured backend.

`websearch` should be vendor-neutral and normalize results, provenance, cost, and errors. The backend choice should be explicit. We should not switch between vendors based on authentication.

jed326:
Thanks @mslipper for the feedback. 

To be clear, this change does not make it so web search backends are swapped based on authentication, `WEBSEARCH_BACKEND` is the single source of truth for that. Both Tako and Parallel backends provide anonymous, unauthenticated free search via MCP, and then deep research is unlocked with authN. I was thinking since the out of the box behavior is the same then the provider flip would be fine, but I do realize that this is breaking specifically for users who have currently set a parallel key, so **I've undone that flip**.

Part of making the web backend provider-neutral is introducing generic parameters names instead of using the parallel specific verbiage (ie `--processor`). This PR keeps that public contract by maintaining the parallel specific parameter names as aliases, which is why there is some churn in those files. Rather than go straight to the PR split you proposed, I re-staged this PR with those as 2 separate commits so they can be viewed independently as there is not much file overlap between the 2. Let me know if that helps, or if you would still prefer the 2 PR strategy.

jed326:
Hey @mslipper , just checking in to see if you think the current approach is reasonable or if you would still prefer 2 separate PRs? Thanks!

jed326:
Bumping this @gakonst @mslipper - I've tested that these changes are non-breaking for existing users, happy to do any additional testing too!

Issue #1517, opened 2026-08-27
Tool dedupe precedence disagrees between api-rs metadata discovery and the sandbox shim installer


oddharsh:
## Summary

When two tool sources publish a tool with the same name, the sandbox installs one copy and api-rs registers the other copy's metadata.

`api-rs` resolves the collision **last wins**, treating the last source as an overlay that overrides the base. `install_tool_shims.py` resolves it **first wins**. Both walk the same ordered source list, so they always pick opposite copies.

No source ordering satisfies both. Reversing the order just swaps which pass gets the copy you wanted.

Because the `[tool.centaur] secrets` declaration comes from the metadata pass and the executable comes from the installer pass, a collision produces a sandbox where the installed client's requests don't match the proxy's injection rule. Nothing is injected and calls fail as 401s that look like a bad key.

## The two paths

`services/api-rs/crates/centaur-api-server/src/tool_discovery.rs`, `collect_plugin_metadata`, **last wins** by overwrite:

```rust
if let Some(prev_dir_idx) = seen_tools.insert(meta.name.clone(), dir_idx) {   // L373
    if let Some(prev_pos) = tools.iter().position(|tool| tool.name == meta.name) {
        warn!(tool = %meta.name, shadowed_dir = ?tool_dirs.get(prev_dir_idx), by_dir = ?base_dir,
              "api-rs tool metadata shadowed");
        tools[prev_pos] = meta;                                              // L382
    }
} else { tools.push(meta); }
```

This looks deliberate. The existing test `discovers_http_oauth_and_overlay_shadowing` passes `&[base, overlay]` and asserts the overlay wins:

```rust
let discovered = discover_tool_proxy_fragment(&[base.clone(), overlay.clone()]).unwrap();
assert_eq!(secrets[0].id.as_deref(), Some("OVERLAY_TOKEN"));
```

`services/sandbox/install_tool_shims.py`, `_copy_published_tools`, **first wins** by skip:

```python
existing = {package_dir.name: package_dir for package_dir in _tool_package_dirs(tool_dir)}   # L149
for package_dir in _tool_package_dirs(published):                                            # L150
    ...
    if tool_name in existing:                                                                # L158
        print(f"skipping duplicate tool {tool_name}: {package_dir} conflicts with {existing[tool_name]}")
        continue
```

`_refresh_tool_dir` copies `sources` in list order and `_copy_published_tools` skips names already present, so the earliest source in the list is the one that survives.

## Repro

[...]

PR #1604, opened 2026-09-04
fix(sandbox): let later overlay tools replace earlier packages

josescuderoh:
## Summary
- Addresses #1517: the sandbox shim installer skipped duplicate tool names (first wins) while api-rs metadata discovery already last-wins for the same ordered source list.
- Later overlay sources now replace an earlier package of the same name, including when the overlay lives in a different category path, matching documented overlay shadowing.

## Test plan
- [x] `python -m unittest discover -s services/sandbox -p 'test_*.py'` (Python 3.11)
- [x] Reproduced first-wins skip with the current agent image, then confirmed last-wins with the patched installer bind-mounted
- [x] CI sandbox Python tests

### DIFF
--- a/services/sandbox/install_tool_shims.py
+++ b/services/sandbox/install_tool_shims.py
@@ -156,11 +156,13 @@ def _copy_published_tools(tool_dir: Path, published: Path) -> None:
         if tool_name in blocklist:
             continue
         if tool_name in existing:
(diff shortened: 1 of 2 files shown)


josescuderoh:
@mslipper small fix for open issue I came across. Thanks!

PR #962, opened 2026-07-08
feat(harness): first-class omp (oh-my-pi) harness

czxtm:
Adds `HarnessType::Omp` as a first-class harness: registration across the enum/DB constraint (migration 0036), container-arg mapping, and an `OmpHarness` server implementation (`crates/harness-server/src/omp.rs`) driving the omp CLI (`@oh-my-pi/pi-coding-agent`, a pi-mono fork) per turn in JSON mode — argv prompt with `--` guard, `-r <id>` session resume, model fall-through to the agent config dir, amp-style terminal settle on `agent_end`, and a normalizer mapping omp's event stream (message_update deltas, toolcall start/delta/end, usage) onto the blocks protocol.

Also included:
- sandbox image: omp + bun (>= 1.3.14) installed in the agent image; `harness/omp/` config fragments (LiteLLM-first static model registry; see the models.yml comment on discovery races)
- `workflows/restart_survival_demo.py`: checkpointed steps + durable sleeps proving restart resume (validated live: api-rs killed mid-sleep resumes at scheduled wake, pre-kill checkpoints byte-identical, attempts=1)
- fix: engines with no auth-mode env (the amp pattern — omp too) no longer force a harness auth fragment, so iron-proxy deployments boot without a placeholder fragment for engines whose credentials arrive as plain env or proxy replace rules
- chart: `sandbox.harnessEngine` docs mention omp; rustfmt pass over harness-server

Running in production on our k3s cluster (sessions, GitHub-bot jobs, PR auto-reviews all drive this harness through the session API). Disclosure: developed with AI assistance; validated by the crate's test suite (84 passing in harness-server incl. 26 omp normalizer/adapter tests) and live E2E turns.

Happy to split the workflow demo or the fragment fix into separate PRs if preferred.

PR #1609, opened 2026-09-06
oh-my-pi harness support

czxtm:
here's the version with omp support we're using in production, as discussed in #962 - I cleaned it up a bit and just tested it, seems okay. most of the slash commands work as well, the advisor in omp is one of the main reasons I was keen on using it.

Note: I'm using omp's native auth-gateway and at the same time using iron-proxy for the other harnesses. So far, works just fine. Omp does it's own routing/load-balancing which is tightly integrated so I avoided getting in its way. 

@mslipper

### DIFF
--- a/harness/omp/config.yml
+++ b/harness/omp/config.yml
@@ -1,0 +1,26 @@
+# Baked omp (oh-my-pi) agent config. Copied to $PI_CODING_AGENT_DIR/config.yml
+# at container start by entrypoint.sh, which substitutes __OMP_LITELLM_BASE_URL__
+# (OMP_LITELLM_BASE_URL) so each deployment points omp at its own LiteLLM gateway.
+#
+# models.yml defines the gateway model STATICALLY; the model id goes on the wire
+# unchanged (discovery-based registries fuzzy-merge ids against models.dev).
+providers:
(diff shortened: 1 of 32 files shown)


harness/omp/config.yml
czxtm, line 1: Pointing this out bc we use LiteLLM internally which is not the ideal default for your users most likely - replace as you see fit.

We also use hindsight as the memory provider which is also deployed via ArgoCD and self-hosted. let me know if you want these manifests included

services/api-rs/crates/centaur-api-server/src/args.rs
```diff
@@ -2003,11 +2003,19 @@ impl IronProxyHarnessArgs {
     /// so sessions restarted onto another harness still get working
     /// credentials through the proxy.
     fn fragments(&self) -> Result<Vec<ProxyFragment>, ServerError> {
-        let mut fragments = vec![self.fragment()?];
+        let mut fragments = Vec::new();
```
czxtm, line 2006: This is the main notable change that's worth review - understandable to remove this part but I have not tested the iron-proxy path. Would need to test that first if you decide to remove this

Issue #262, opened 2026-05-28
Slack Socket Mode support for slackbot


k7k0:
## Summary

Add support for running the slackbot service in [Slack Socket Mode](https://api.slack.com/apis/socket-mode) as an alternative to the current HTTP Events API path. This is purely additive — operators that prefer the current public-webhook architecture can keep using it.

## Motivation

The current architecture requires Slack to POST events to `/api/slack/events` over the public internet, which means the slackbot needs a publicly-resolvable hostname, a public TLS endpoint, and HMAC verification (`services/slackbot/src/slack/signature.ts`). For self-hosted operators that's a non-trivial set of moving parts.

Socket Mode flips the direction: the slackbot opens an outbound WebSocket to Slack and pulls events down. That has some concrete benefits:

- **No public ingress required.** The bot runs entirely on private networks (VPN/zero-trust/on-prem clusters with no public DNS). In our case we run on a WARP-gated `*.sand.<org>.co` zone where the cluster's hostname isn't publicly resolvable, and we had to stand up a dedicated Cloudflare Tunnel + path-scoped Access policies just to expose `/api/slack/events`. Many self-hosters will hit the same wall.
- **Smaller attack surface.** No public HTTP endpoint, no need to harden signature verification against replay/timing edge cases, no inbound webhook receiver to keep patched.
- **Recommended by Slack** for self-hosted apps where a public callback URL is awkward — see [Socket Mode docs](https://api.slack.com/apis/socket-mode).

## Suggested shape

Roughly:

1. Add `@slack/socket-mode` as a dependency (or migrate the service to `@slack/bolt`, which transparently supports both transports).
2. Read a new `SLACK_APP_TOKEN` (xapp-…) from the existing `centaur-infra-env` Secret when socket mode is enabled. Keep `SLACK_BOT_TOKEN` as-is.
3. Chart flag, e.g. `slackbot.socketMode.enabled: true`, that:
   - Skips creating the public Service/route for `/api/slack/events` (the chart currently expects an Ingress at that path)
   - Drops `SLACK_SIGNING_SECRET` from required env (still allowed if both modes need to coexist)
4. App manifest example with `settings.socket_mode_enabled: true` and the `connections:write` scope.

## Trade-offs worth calling out in docs

- Socket Mode is a single-consumer model (one connection per app token), so horizontal scaling of the slackbot needs to be thought through. The Slack SDK reconnects on drop, but there's no built-in HA across replicas.
[...]

PR #1610, opened 2026-09-07
feat: add Slack Socket Mode support to slackbotv2

yanpla:
Receive Slack events over an outbound Socket Mode connection instead of inbound webhooks, so operators without a public endpoint can run the Slackbot. Opt in with slackbotv2.socketMode (SLACKBOTV2_SOCKET_MODE); the bot reads an app-level token (SLACK_APP_TOKEN, xapp- with connections:write) and no longer requires SLACK_SIGNING_SECRET.

In socket mode the Chat SDK would otherwise never initialize, since its lazy init is triggered by the first inbound webhook; initialize eagerly at startup so the adapter's socket connection actually opens.

#262 Was incorrectly marked as completed, merging this would actually complete it. :)

### DIFF
--- a/contrib/chart/templates/slackbotv2.yaml
+++ b/contrib/chart/templates/slackbotv2.yaml
@@ -58,6 +58,20 @@ spec:
                 secretKeyRef:
                   name: {{ include "centaur.secretEnvName" . }}
                   key: {{ printf "%sSLACK_SIGNING_SECRET" .Values.secretManager.envPrefix }}
+{{- if .Values.slackbotv2.socketMode }}
+                  # Optional in Socket Mode.
+                  optional: true
+{{- end }}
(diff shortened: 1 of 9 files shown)


mslipper (changes requested):
Thanks for adding this. I found a few blockers before it’s safe to merge:

- Socket events bypass the external-org checks, including for block actions.
- Socket startup races the initial Postgres connection. A temporary DB failure can leave the pod healthy but disconnected until restart.
- A missing or invalid `SLACK_APP_TOKEN` only logs an error; startup or readiness should fail.
- Fresh secret bootstrap does not include `SLACK_APP_TOKEN`.
- Socket events skip auto-join and late-file repair.

Also, disabling the shared ingress would disable GitHub and Linear webhooks.

PR #1614, opened 2026-09-07
feat: support signed sandbox commits

brendanjryan:
## Motivation

Some repositories require cryptographically signed commits. Sandboxes need a supported signing path to satisfy those repository rules.

## Summary

- Add default-off OpenPGP commit signing backed by an existing Kubernetes Secret
- Configure and preflight signing during sandbox startup, failing closed when unusable
- Reconcile retained sandboxes when signing configuration changes
- Document overlay policy and secret provisioning

## Key design considerations

- Keep private key material out of chart values and overlay repositories
- Pin protected signing environment variables after operator overrides
- Scope the opt-in to deployments where every sandbox principal may use the signing identity

### DIFF
--- a/contrib/chart/Chart.yaml
+++ b/contrib/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: centaur
 description: Helm chart for the trusted Centaur control plane
 type: application
-version: 0.1.136
+version: 0.1.137
 appVersion: "0.1.0"
 dependencies:
   - name: connect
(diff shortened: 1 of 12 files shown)


mslipper (changes requested):
Thanks for putting this together. My main concern is that the control plane and chart now have a large amount of GPG-specific configuration and reconciliation logic. I'd rather build something more general, and have the GPG script and tests be an example or documented recipe than a built-in control plane behavior. Something like this:

1. Mount an operator-provided Secret into sandbox pods.
2. Run an operator-provided startup hook.
3. Pass configuration through the existing sandbox environment.
4. Recycle retained sandboxes when the configuration changes.

This way we can support additional features like SSH signing and the like without requiring new parameters for each use case. If this is something you'd like to implement I'd be happy to leave it to you, otherwise I will take the implementation here.

PR #1654, opened 2026-09-14
fix: ignore collab subagent turn completions

### LINKED ISSUE #1696: Subagent turn/completed closes the parent execution's event stream
## Summary

When a Codex turn spawns collab subagents, the first subagent's `turn/completed` notification is treated as end-of-execution: `api-rs` emits `session.execution_completed`, the execution-scoped event stream closes, and everything the parent (and later subagents) produce afterward is persisted but never delivered to connected clients. The parent's `final_answer` is lost to the stream.

## Expected behavior

The execution stays open until the parent/root turn completes; subagent completions stream through as progress.

## Actual behavior

The stream closes within milliseconds of the first subagent turn completing. Streaming clients stop receiving events while the parent turn is still `inProgress`, and the parent's final answer lands with no open stream.

## Evidence

Parent turn `turn-parent` on `thread-parent` runs long, spawning subagents sequentially. A subagent on `thread-child-1` emits (IDs illustrative):

```json
{"method": "turn/completed", "params": {"threadId": "thread-child-1", "turn": {"id": "turn-child-1", "status": "completed"}}}
```

while the parent turn is still `inProgress`. Within milliseconds, `session_events_stream_completed` fires for the execution; the remaining parent output (further subagents plus `final_answer`) never reaches the stream. The parent's own `collabAgentToolCall` (`wait`, `status: completed`) in the same window confirms the orchestrator treated the child as one step, not the end.

## Impact

- Any multi-subagent run delivers only up to the first child completion over a live stream; final answers are silently dropped.
- Follow-up messages open new streams positioned after the missed events, so clients see a short "already complete" summary instead of the real handoff.
- The same any-`turn/completed`-is-terminal assumption exists in the ingress stream parsers and the rendering mapper, which also stop at the first child completion.

## Suspected root cause

Completion detection keys on any `turn/completed` output line without checking `params.threadId` / `params.turn.id` against the execution's root turn.

josescuderoh:
## Summary

- Codex multiplexes collab subagent turns on the same stdout stream, each emitting `turn/completed` with its own thread/turn ids. Every layer treated any `turn/completed` as end-of-execution, so the first child completion closed the event stream while the parent turn was still running and the parent final answer was never delivered.
- `api-rs`: stdout pump and recovery replay track the root turn per execution and only treat its completion as terminal; child agent text stays out of the final-answer buffer and first-token detection.
- New shared `TurnCompletionFilter` in `@centaur/harness-events`, used by all five ingress stream parsers (slackbotv2, discordbot, githubbot, linearbot, teamsbot) and the rendering mapper, which also routes child agent text to commentary so the final answer stays parent-only.
- Conservative by design: completions for turns never seen started stay terminal, so pump restarts and stream reconnects degrade to historical behavior instead of hanging.

## Test plan

- `cargo test -p centaur-session-runtime --lib` (134 pass, incl. new collab cases), `cargo fmt --all --check`, `cargo clippy -p centaur-session-runtime --all-targets -- -D warnings`
- harness-events filter tests 7/7, rendering `codex-app-server.test.ts` 29/29
- discordbot/githubbot/linearbot/teamsbot suites fully green; slackbotv2 `session-api` 49/49 (remaining slackbotv2 failures and type errors verified pre-existing on upstream `main`)
- Bugbot review clean

Fixes #1696


josescuderoh:
@mslipper this PR fixes silent drops on every multi-worker collab run: the first child's turn/completed was closing the event stream while the parent was still running, so the parent's final answer never delivered. Root-turn tracking in `api-rs` + all 5 bots + renderer, fail-safe fallbacks, CI green — one commit, ready to merge.

PR #1677, opened 2026-09-16
fix: skip stale scheduled tasks before execution and Slack delivery

Zygimantass:
## Problem
Disabling or deleting a Console scheduled task stops new scheduling but leaves already-enqueued `console_workflow` runs alive. Those runs use an input snapshot and can continue sending reminders after the task is gone.

Incident: https://tempoxyz.slack.com/archives/C0AJ07U8Z1N/p1789556437526329

## Proposed fix
- Add an admin-authenticated, non-cacheable Console endpoint that checks live task existence/enabled state, active author, execution-principal ownership, destination, and current Slack delivery policy.
- Reuse the workflow runtime's Console client to check immediately before `ctx.agent_turn` and every `ctx.post_to_slack`. Do not checkpoint the eligibility read. Authentication, transport, and malformed-response errors fail closed.
- Return a silent skipped result for ineligible runs. Handle it before the empty-response fallback, at the root post, and between chunk replies; preserve delivery checkpoints on replay.

## Scope / rollout
- Deploy the Console endpoint before the API runtime; deploy the Python workflow skip handling before or with the runtime guard. An old Console makes checks fail closed, and an old Python workflow does not understand skipped results.
- This drains stale queued runs without executing agents or delivering results; it does not proactively cancel the queue or interrupt an agent already running. It cannot retract a Slack request already issued between a state check and disable/delete.
- Manual runs of disabled tasks are also skipped. Re-enabling before a pending run is checked can make it eligible again; there is no cancellation generation/tombstone in this patch.
- The conflicting “NEVER create or update a scheduled task” wrapper and one-shot completion semantics remain separate work.

## Validation
Passed:
- `cargo test -p centaur-workflows --lib` — 54 tests
- `cargo test -p centaur-iron-control --lib` — 95 tests
- `cargo fmt --all --check`
- `cargo clippy -p centaur-workflows -p centaur-iron-control --all-targets -- -D warnings`
- `pytest workflows/tests/test_console_workflow.py -q` — 11 tests
- Targeted RuboCop for all four changed Ruby files; Brakeman — no warnings
- `git diff --check`

Not fully verified, hence draft:
- Added Rails integration tests could not start locally: native `libvips` is missing. Available Ruby was 3.3.11 rather than the repository's 3.4.8.
- Docker is unresponsive, the local database connection times out, and ParadeDB is unavailable on its expected port. Runtime image build/local-stack end-to-end verification remains outstanding. No production deployment was used for testing.
- Ruff reports existing `TRY004` and `B023` findings in `console_workflow.py`; confirmed the same findings on the base revision and left them unchanged.


mslipper:
Superseded by #1678.

PR #1681, opened 2026-09-16
fix: report sandbox replacement clearly

0x666c6f:
## Problem

When a sandbox pod disappears mid-turn for an infrastructure reason — eviction,
node replacement, `Gone`, or `NotFound` — the stdout pump treats it like any
other dead sandbox. Requesters then see a raw internal detail such as `sandbox no
longer accepts io` or `sandbox stdout closed before terminal output`.

Automatic replay is unsafe: a tool may complete an external side effect before
its stdout is persisted.

## Change

- Detect infrastructure replacement from a changed pod instance, Kubernetes
  `Stopped` + `Evicted`, `Gone`, or `NotFound`.
- Fail the execution without replaying it and return a clear requester message:
  `The sandbox running this response was replaced before it finished; please re-ask.`
- Preserve existing failure details for non-eviction terminal states.
- Classify replacement failures as `evicted` in session failure metrics.

## Testing

- `cargo fmt --all --check`
- `cargo check -p centaur-session-runtime -p centaur-session-sqlx --tests`
- `cargo clippy -p centaur-session-runtime -p centaur-session-sqlx --tests -- -D warnings`
- Focused tests for pre-deletion eviction, replaced pod instances, non-eviction
  terminal states, and failure classification.


mslipper (changes requested):
After looking at this more closely, I don't think we can safely replay a turn based on the absence of persisted stdout. The sandbox could have completed an external side effect before its output was recorded.

Could we reduce this PR to detecting sandbox replacement and failing the execution with a clearer requester-facing message?

If we want automatic replay later, I think we'll to modify the harness/tools to be idempotent rather than looking at output events.

Would the smaller PR still resolve the immediate user problem you're trying to address?

0x666c6f:
Yes, this is definitely not an urgent nor critical matter.

Addressed in `4b8c179f`: removed automatic replay and all replay-only store/event machinery. Infrastructure replacement now always fails with the requester-facing message, including Kubernetes `Stopped` + `Evicted`, `Gone`, `NotFound`, and pod-instance replacement shapes.

PR #1701, opened 2026-09-20
fix: search shared drives in gsuite Drive list and Drive ETL

parithosh:
`drive_list` in the gsuite tool and `GoogleDriveReadonlyClient.list_docs` in the Drive ETL both pass `includeItemsFromAllDrives` without `corpora`, so Drive searches the default `user` corpus and omits files reachable only through shared drive membership. An account whose access is entirely via shared drives gets `No files found` from every unqualified list or search and an empty `google_drive_sync`. Set `corpora=allDrives` in both so My Drive, shared-with-me, and every shared drive the account belongs to are searched. Also add `gsuite drive drives` (`drive_list_drives`) to enumerate the shared drives the account can see, so an agent can pick a drive id for `drive list --folder`. Related: #1702 (creating files inside a shared drive).


parithosh:
Companion: #1702 lets `docs create` place a document inside a folder or shared drive, so an agent can both find and write docs in a shared-drive-only account. Independent of this change.

